PR #12, opened 2023-10-17
Add GHA to publish to galaxy

tonyskapunk:
This GitHub Action publishes the collections to ansible-galaxy when creating a release

This is a proposal to publish to galaxy whenever we create a release, we could do it on merge in main or with a particular tag, let me know what do you think

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,7 +13,7 @@ jobs:
 
       - name: Add release, date and commit hash to version in galaxy.yml
         run: >
-          ts=$(date +%Y%m%d%H%M);
+          ts=$(TZ=UTC0 git show ${{ github.sha }} --no-patch --format=%cd --date=format-local:'%Y%m%d%H%M')
           sha=$(echo "${{ github.sha }}" | cut -c1-7);
           rel=$(grep ^Release ansible-collection-redhatci-ocp.spec | grep -Po '\d+\.');
           sed -ir "s/^(version: .*)$/&-${rel}${ts}git${sha}/" galaxy.yml;
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,17 @@
+name: Ansible Galaxy Publish
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Ansible Publish to Galaxy
+        uses: ansible/ansible-publish-action@v1.0.0
+        with:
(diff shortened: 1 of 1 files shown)


tonyskapunk:
Last change attempts to bring a closer parity to the versioning used in the RPM, a difference likely occur in the date timestamp as it may be a difference of minutes between RPM build and galaxy publication

nsilla:
It looks like a must-have feature to me, but I'd like to hear other opinions.

Thanks.

thekad:
> Last change attempts to bring a closer parity to the versioning used in the RPM, a difference likely occur in the date timestamp as it may be a difference of minutes between RPM build and galaxy publication

The version time stamp should be obtained from the commit time stamp, that way it will be the same for everything. Also the semver specification says this sort of info should be in the form ${major}.${minor}.${patch}+${metadata}` galaxy will reject the package if it doesn't comply with semver.

thekad:
here's the spec https://semver.org/#spec-item-10

fredericlepied:
> > Last change attempts to bring a closer parity to the versioning used in the RPM, a difference likely occur in the date timestamp as it may be a difference of minutes between RPM build and galaxy publication
> 
> The version time stamp should be obtained from the commit time stamp, that way it will be the same for everything. 

That's a good idea and we need to do the same thing in dci-packaging to have exact same timestamp.

fredericlepied:
> > > Last change attempts to bring a closer parity to the versioning used in the RPM, a difference likely occur in the date timestamp as it may be a difference of minutes between RPM build and galaxy publication
> > 
> > 
> > The version time stamp should be obtained from the commit time stamp, that way it will be the same for everything.
> 
> That's a good idea and we need to do the same thing in dci-packaging to have the exact same timestamp.

that could be done like this: `git show -s --format=%cd --date=format:'%Y%m%d%H%M%S'`

tonyskapunk:
> The version time stamp should be obtained from the commit time stamp, that way it will be the same for everything. 

+1, my last patch now uses the commit date.

> That's a good idea and we need to do the same thing in dci-packaging to have the exact same timestamp.

+1, I'll take a look to the packaging unless @thekad beat me to do it ;)



> Also the semver specification says this sort of info should be in the form `${major}.${minor}.${patch}+${metadata}` galaxy will reject the package if it doesn't comply with semver.

Yup, it is covered, here is an example on how the latest commit would show in galaxy, [it adheres to semver](https://jubianchi.github.io/semver-check/#/version/0.2.0-1.202310311522git63d9b6a)

```
version: 0.2.0-1.202310311522git63d9b6a
```

tonyskapunk:
`redhatci` ns in galaxy has been created in ansible/galaxy#3284 dcibot has been added to the group and ready to publish

PR #13, opened 2023-10-17
Recreate Storage class

betoredhat:
- Storage class parameters are inmutable, removing the SC to be re-created with the new or the same parameters.
- That is the recommended practice, it does not affect the already provisioned volumes


betoredhat:
The SC creation is not failing if already exists. The issue with SRIOV is not related to this change but happened because there cluster already had a CSV for SRIOV installed.

tonyskapunk (approved):
LGTM

PR #17, opened 2023-10-20
Remove depencencies on DCI var

betoredhat:
- kubeconfig path is a variable from DOA, we apply the pass the KUBECONFIG using apply. This will allow to run the role independently
- Minor updates to the docs

### DIFF
--- a/roles/check_resource/README.md
+++ b/roles/check_resource/README.md
@@ -15,3 +15,33 @@ resource\_to\_check         | Yes       | MachineConfigPool      | Name of the r
 check\_wait\_retries        | Yes       | Undefined              | Number of times in which the wait task is performed.
 check\_wait\_delay          | Yes       | Undefined              | Time spent between wait tasks' iterations.
 check\_reason               | No        | Undefined              | Reason for the check to be done.
+
+## Requirements
+
+A running OpenShift cluster with the proper credentials is required, credentials must be passed as by setting the KUBECONFIG environment.
+
+## Example of usage
+
+Confirm that Machine Config Pools are not updating
+```
+- name: "Wait for updated MCP after applying ICSP"
+  include_role:
+    name: redhatci.ocp.check_resource
+  vars:
+    resource_to_check: "MachineConfigPool"
+    check_wait_retries: 120
+    check_wait_delay: 10
+    check_reason: "Apply ICSPs for mirrored catalogs"
+```
+
+Confirm that OVN pods are running
+```
+- name: "Wait for updated MCP after applying ICSP"
+  include_role:
+    name: redhatci.ocp.check_resource
+  vars:
+    resource_to_check: "SriovNetworkNodeState"
+    check_wait_retries: 120
+    check_wait_delay: 10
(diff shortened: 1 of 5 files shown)


betoredhat:
check dallas ocp-4.13-vanilla

roles/check_resource/README.md
```diff
@@ -15,3 +15,33 @@ resource\_to\_check         | Yes       | MachineConfigPool      | Name of the r
 check\_wait\_retries        | Yes       | Undefined              | Number of times in which the wait task is performed.
 check\_wait\_delay          | Yes       | Undefined              | Time spent between wait tasks' iterations.
 check\_reason               | No        | Undefined              | Reason for the check to be done.
+
+## Requirements
+
+A running OpenShift cluster with the proper credentials is required, credentials must be passed as by setting the KUBECONFIG environment.
+
+## Example of usage
```
ramperher, line 23: Hi Beto, I'm a bit confused here. Doesn't it mean we need to pass the following when calling check_resource?

```
    apply:
      environment:
        KUBECONFIG: "{{ kubeconfig_path }}"
```

Because I can see this in the storage_tester example, but not here.

If this is true, this would require to update all the calls to check_resource to pass the KUBECONFIG (in the agent, in partners' hooks like SPK, etc.)

roles/check_resource/README.md
```diff
@@ -15,3 +15,33 @@ resource\_to\_check         | Yes       | MachineConfigPool      | Name of the r
...
+  include_role:
+    name: redhatci.ocp.check_resource
+  vars:
+    resource_to_check: "MachineConfigPool"
+    check_wait_retries: 120
+    check_wait_delay: 10
+    check_reason: "Apply ICSPs for mirrored catalogs"
+```
+
+Confirm that OVN pods are running
```
manurodriguez, line 37: Can we update the wording of this step? just to make it match what it does, it reads `OVN pods` but includes a check_resource for SriovNetworkNodeState, maybe something about "Confirming SRIOV node state.."

betoredhat: Done

roles/check_resource/README.md
betoredhat: Hello Ramon, sorry about the confusion. I removed that var from the storage_tester role. 

We do not need to update the check_resources calls as we are indirectly passing the KUBECONFIG level at each DOA stage: https://github.com/redhat-cip/dci-openshift-agent/blob/84a991f1b927a8358655c1d63066ab37bc708d69/dci-openshift-agent.yml#L419-L423 

That is inherited in roles, plays hooks, etc. But let me do testing with VCP jobs. 

The requirement advice in the READMEs to define that variable should be enough if someone wants to use the role without DOA.

betoredhat:
check dallas ocp-4.14-vcp example-cnf

manurodriguez (approved):
LGTM

manurodriguez:
recheck

ramperher (approved):
Thanks for the clarification Beto!

PR #18, opened 2023-10-20
Preflight 1.7.1 improving config handling and tests for helm-operators

tkrishtop:


manurodriguez (approved):
LGTM, 
I see the new version in the task ` [Run preflight runtime-assets]` but wouldn't be cool to have a preflight component? something like we do in tnf, we retrieve it with a task and include it in the job during execution : )

ramperher:
> LGTM, I see the new version in the task ` [Run preflight runtime-assets]` but wouldn't be cool to have a preflight component? something like we do in tnf, we retrieve it with a task and include it in the job during execution : )

That could be a good feature, yes. In tnf, we just take the commitId from the repo and create a component based on that commit, maybe preflight can do the same. 

Apart from this, the change LGTM

tkrishtop:
> LGTM, I see the new version in the task ` [Run preflight runtime-assets]` but wouldn't be cool to have a preflight component? something like we do in tnf, we retrieve it with a task and include it in the job during execution : )

Hi @manurodriguez and @ramperher , thanks for pointing that out, I'll add it in a separate PR.

PR #20, opened 2023-10-20
Add wait conditions during cleanup

betoredhat:
BGPPeers are still alive by the time their dependencies are been deleted too 
Waiting for proper resources termination

### DIFF
--- a/roles/metallb_setup/tasks/clean-resources.yml
+++ b/roles/metallb_setup/tasks/clean-resources.yml
@@ -6,6 +6,11 @@
     kind: BGPPeer
     name: "{{ peer.name }}"
     namespace: "{{ mlb_namespace }}"
+    wait: true
+    wait_condition:
+      status: False
+    wait_sleep: 2
+    wait_timeout: 300
   loop: "{{ mlb_bgp_peers }}"
   loop_control:
     loop_var: peer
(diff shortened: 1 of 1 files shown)


betoredhat:
check dallas ocp-4.13-vanilla

betoredhat:
Trying to fix this issue: https://www.distributed-ci.io/jobs/b4fb63e3-1291-44b2-975f-f9d26141590b/jobStates?sort=date&task=4d806f4b-15f4-4997-b667-e4ebc4e2ca7e

manurodriguez:
> Trying to fix this issue: https://www.distributed-ci.io/jobs/b4fb63e3-1291-44b2-975f-f9d26141590b/jobStates?sort=date&task=4d806f4b-15f4-4997-b667-e4ebc4e2ca7e

I remember this, and the problem is that there are two main groups of resources to delete during the teardown, one for each BGP IPAddress pool, so when the first group tries to delete BFDProfile it fails with a 403 code because is still in used by the other IPAddress pool (from a diff vlan) so that's why I added the ignore_errors, of course if was a simple hack, when the resources of the second IPAddress pool get deleted now works because there are no more resources using the BFDProfile.
Maybe we can create two BGPProfiles (it does not matter if they have the same settings) one per IPAddress pool, then we can delete the respective BFDProfile of each in the teardown to avoid this issue.

thekad (approved):

PR #21, opened 2023-10-23
do not set provisioningNetwork in acm_setup role

manurodriguez:
currently in deployments with IPI provisioningNetwork is set to Managed and that requires extra provisioning* parameters to define the CIDR, IP, and Range, but in Assisted mode none of that information is set or required.

This change removes the provisioningNetwork definition when adding watchAllNamespaces:true to allow compatibility with Assisted Installer


ramperher (approved):
LGTM

thekad (changes requested):
Now that this role lives in a different place from DCI OCP agent we gotta stop relying on facts set in the agent. A better approach would be to use a variable (if there isn't one already) to set the provisioning network, then pass that appropriately from the OCP agent.

thekad (approved):

thekad:
@manurodriguez let me know when you feel comfortable for this PR to be merged, I saw the successful runs but not sure if you have some more tests in mind.

tonyskapunk (approved):
:+1:

manurodriguez:
> @manurodriguez let me know when you feel comfortable for this PR to be merged, I saw the successful runs but not sure if you have some more tests in mind.

Thanks @thekad I'm going to merge now, apologies for the delay and lots of tests without more details, till yesterday we were able to sync about the final changes, but we tested all the scenarios now,   I avoid the need to use facts from the agent, but I agree that if needed we need to find a way to pass them to the collections and document them as requirements.

More details in https://issues.redhat.com/browse/CILAB-1283

Thanks

PR #22, opened 2023-10-24
Remove dependency on mirror_images role

betoredhat:
- Remove dependency on mirror_images role and DCI variable


manurodriguez (approved):
LGTM

PR #25, opened 2023-10-30
Repeat MCP check if hitting OCPBUGS-22677

ramperher:


nsilla (approved):
Looks OK. Thanks for the work.

thekad (approved):
yup

ramperher:
Indentation was not correct, retesting it

ramperher:
I'll finally close this one, the workaround will not help in this issue. It has already commented in the OCPBUGS that the real cause of the issue is [OCPBUGS-22409](https://issues.redhat.com/browse/OCPBUGS-22409), related to reserved CPU selection in the performance profile, same case than ours.

PR #26, opened 2023-10-30
fbc_catalog: Add creds option during the bundle inspection

pierreblanc:


roles/fbc_catalog/tasks/add-bundle.yml
```diff
@@ -27,6 +27,9 @@
   shell:
     cmd: >
       skopeo inspect
+      {% if partner_creds | length %}
+      --authfile {{ partner_creds }}
```
tonyskapunk, line 31: Did you mean fbc_opm_auths?

tonyskapunk (changes requested):

roles/fbc_catalog/tasks/add-bundle.yml
tkrishtop: For me, the `partner_creds` seems like the correct choice here because we have to access the operator, potentially hosted in the private registry.

tkrishtop (approved):

pierreblanc:
check dallas ocp-4.13-vanilla preflight

roles/fbc_catalog/tasks/add-bundle.yml
tonyskapunk: If `fbc_opm_auths` was not meant in here and `partner_creds` was, then this is adding a new set of credentials that need to be documented and follow the variable role naming.

tkrishtop:
cc: @betoredhat We checked together during the CI sync and it probably makes sense to call the role fbc_catalog with the variable pointing to the creds file instead of using DOCKER_CONFIG. We'll need to modify the whole role fbc_catalog for this, doing the similar changes that @pierreblanc does. wdyt? 

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/preflight/tasks/mirroring.yml#L29-L32

betoredhat:
> cc: @betoredhat We checked together during the CI sync and it probably makes sense to call the role fbc_catalog with the variable pointing to the creds file instead of using DOCKER_CONFIG. We'll need to modify the whole role fbc_catalog for this, doing the similar changes that @pierreblanc does. wdyt?
> 
> https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/preflight/tasks/mirroring.yml#L29-L32

@tkrishtop @pierreblanc I think we can replace skopeo with oc-mirror to get the default channel. As oc-miror and opm use the same auth mechanism we can keep the same auths behavior. 

Let me work on that change

roles/fbc_catalog/tasks/add-bundle.yml
betoredhat: the DOCKER_CONFIG  + '/config.json' contains the credentials file required for skopeo. I suggest just passing to the auth-file option

```
      {% if lookup('env', 'DOCKER_CONFIG') | length %}
      --authfile {{ lookup('env', 'DOCKER_CONFIG')+'/config.json' }}
      {% endif %}
```

betoredhat:
> > cc: @betoredhat We checked together during the CI sync and it probably makes sense to call the role fbc_catalog with the variable pointing to the creds file instead of using DOCKER_CONFIG. We'll need to modify the whole role fbc_catalog for this, doing the similar changes that @pierreblanc does. wdyt?
> > https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/preflight/tasks/mirroring.yml#L29-L32
> 
> @tkrishtop @pierreblanc I think we can replace skopeo with oc-mirror to get the default channel. As oc-miror and opm use the same auth mechanism we can keep the same auths behavior.
> 
> Let me work on that change

Hello, 

Sorry, what I was thinking didn't work, oc-mirror can extract the default channels for indexes not for bundles, I tried with crane and had no luck. I added a suggestion to use the already defined env variable (DOCKER_CONFIG) to be used by skopeo. 

Thanks

tkrishtop:
check workload preflight-green

pierreblanc (commented):
LGTM, feel free to merge it :)

betoredhat:
check workload preflight-green

betoredhat:
check workload preflight-green

tonyskapunk (approved):
LGTM

roles/fbc_catalog/tasks/add-bundle.yml
nsilla: just for clarification, we checked the status of the docker config file in the affected lab. The file existed but it was empty.

PR #27, opened 2023-10-31
Change the way we are retrieving the commitId for repos that come from PRs

ramperher:
Please check https://issues.redhat.com/browse/CILAB-1288 to understand this story.

This change does the following:

- Create a new play, reusing the logic already included in tnf, and use it for dev_gits_to_components cases, so that the component created has the correct commitId
- Remove this code from tnf to make it generally available
- In tnf, call to include_components when we're testing a stable version, following gits_to_components method.

### DIFF
--- a/roles/cnf_cert/tasks/pre-run.yml
+++ b/roles/cnf_cert/tasks/pre-run.yml
@@ -83,11 +83,12 @@
       mode: u+wx,g+wx
       recurse: true
 
-  # We need to check the SHA in this way, because extract-dependencies changes the
-  # commit id when creating the new branch based on the PR, but only if new PRs are merged
-  # in the repo and we have not included them in our change. In these cases, the original
-  # id is saved in .git/ORIG_HEAD file. Else, we can find it in the location called
-  # `.git/refs/heads/prXXX`
+  # Retrieve the SHA to use it when building the tnf image locally.
+  # Following the same method it is defined in include_components role,
+  # track_dev_git_repo.yml playbook.
+  # However, in this case, we don't need to create the component in this
+  # role, since it is already done in dci-openshift-app-agent through
+  # dev_gits_to_components variable.
   - name: "Check if .git/ORIG_HEAD file exists"
     stat:
       path: "{{ tnf_dir }}/.git/ORIG_HEAD"
@@ -134,35 +135,6 @@
       --build-arg OPENSHIFT_VERSION={{ ocp_version_full }} .
     args:
       chdir: "{{ tnf_dir }}"
-
-  # This is needed because the DCI component is not automatically created, as
-  # it is done in case the git repo is directly used, in whose case it is created.
-  # Doing this in the same way that is done in track_git_repo play but with
-  # the arguments obtained in previous tasks (e.g. commit SHA), as we cannot
-  # rely on the last commit id. For that reason, it is better to just do it
-  # here rather than changing pre-run and track_git_repo to be adapted to this
-  # casuistic
-  - name: Manually create the corresponding tnf component for DCI
-    ansible.legacy.dci_component:
-      display_name: "{{ test_network_function_project_name }} {{ tnf_version_image[:7] }}"
-      version: "{{ tnf_version_image }}"
-      team_id: "{{ job_info['job']['team_id'] }}"
-      topic_id: "{{ job_info['job']['topic_id'] }}"
-      type: "{{ test_network_function_project_name }}"
-      url: "https://github.com/test-network-function/cnf-certification-test/commit/{{ tnf_version_image }}"  # noqa 204
-      state: present
-    register: tnf_git_component
-    tags: [dci]
-
-  - name: Attach tnf component to the job
-    ansible.legacy.dci_job_component:
-      component_id: "{{ tnf_git_component.component.id }}"
-      job_id: "{{ job_id }}"
-    register: job_component_result
-    until: job_component_result is not failed
-    retries: 5
-    delay: 20
-    tags: [dci]
 
   when:
     - test_network_function_src_dir is defined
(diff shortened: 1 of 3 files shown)


ramperher:
> * SUCCESS https://www.distributed-ci.io/jobs/a0d5daee-10ad-4957-a04c-12539b10634a/jobStates
>   from [Change the way we are retrieving the commitId for repos that come from PRs #27](https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/27)

This is working fine. Now, cnf-certification-test component is created again, pointing to the last commit right now in the repo, which is https://github.com/test-network-function/cnf-certification-test/commit/d6177a4ada53101b0b02b43745b9b9bc2c80832f. And the component for this repo, it's taking the last commit by the time I ran the test.

Last job failed because of a variable being modified in two different tasks, I'm fixing it with my last commit.

ramperher:
Change is now working in the second case tested (it is about to finish). Here we have the components created, and they're pointing to the latest commitId in the PR created:

[ansible-collection-redhatci-ocp da8621a](https://www.distributed-ci.io/topics/516ea031-83a0-46b7-add2-1862f82dd12a/components/fec7ad08-b21c-49db-830b-682e4aa1a654) -> https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/27/commits/da8621a25c5939890cf76a35fcdc022355c0bddc
[cnf-certification-test 3aedee7](https://www.distributed-ci.io/topics/516ea031-83a0-46b7-add2-1862f82dd12a/components/16170331-b555-4356-9896-206b38423468) -> https://github.com/test-network-function/cnf-certification-test/pull/1567/commits/3aedee7ed52c2e73454f5007bb5216526e75454a

Change is ready for review

ramperher:
Next execution will have the correct component creation, without embedding the Github token, see this one for ansible-collection-redhatci-ocp: https://www.distributed-ci.io/topics/516ea031-83a0-46b7-add2-1862f82dd12a/components/9a091050-c5a4-4179-86fa-e18e03c4148a (the one for tnf is reusing the one from the previous job since it's the same, but for next jobs, when creating a new component, the URL should be now correct).

roles/cnf_cert/tasks/pre-run.yml
```diff
@@ -237,6 +209,17 @@
...
     set_fact:
       tnf_commit_sha: "{{ tnf_commit_sha_cmd.stdout }}"
+
+  - name: Create cnf-certification-test component calling to include_components role
+    vars:
+      gits_list:
+        - "{{ tnf_dir }}"
+      ic_gits: "{{ gits_list|flatten }}"
+      ic_rpms: []
+      ic_dev_gits: []
```
manurodriguez, line 219: Just curious, I assume the play `track_dev_git_repo.yml` will never be called from CNF role tasks if we pass this `ic_dev_gits: []` right? it will always ended up using `track_git_repo.yml` because we pass the `ic_gits` list with the `tnf_dir`. So this is expected right?

Also, in all the jobs you referenced, I see that the play `track_dev_git_repo.yml` is the agent pre-run, creates the git components, I just couldn't find where we pass `ic_dev_gits`, by default is empty, so please if you happen to know where is that overridden, I'll appreciate it : )

manurodriguez (approved):
LGTM, just related questions

roles/cnf_cert/tasks/pre-run.yml
ramperher: Hi @manurodriguez ! thanks for your comments, I'll answer you below:

> Just curious, I assume the play track_dev_git_repo.yml will never be called from CNF role tasks if we pass this ic_dev_gits: [] right? it will always ended up using track_git_repo.yml because we pass the ic_gits list with the tnf_dir. So this is expected right?

Exactly, that's expected. The thing is, before, the cnf-certification-test component when testing stable versions (so, when the code doesn't come from a PR; for example, `v4.5.2`, `HEAD`, etc.) was created automatically, but now it isn't. With this code, we use `include_components` to create that component. As the code doesn't come from a change, we need to use `ic_gits`, and the rest of arguments are passed as empty list to avoid calling them.

> Also, in all the jobs you referenced, I see that the play track_dev_git_repo.yml is the agent pre-run, creates the git components, I just couldn't find where we pass ic_dev_gits, by default is empty, so please if you happen to know where is that overridden, I'll appreciate it : )

It's updated in the test-runner: https://github.com/redhat-cip/dci-pipeline/blob/master/tools/test-runner#L286-L289; so, for each folder extracted by extract-dependencies (so, they come from a change), they're added to `dev_gits_to_components` variable, which is eventually passed to `include_components role through `ic_dev_gits`. That's the way it's working for ansible-collection-redhatci-ocp and cnf-certification-test (where we are testing a PR).

ramperher:
Last two jobs failed because of an issue not related to this change. In fact, you can check that the components are now created correctly and with the correct URL:

ansible-collection-redhatci-ocp 081b068
 -> https://www.distributed-ci.io/topics/afbb1c48-277e-4ddb-aa14-d24fe27434a7/components/b1b4a86c-03dc-44a0-b763-47bceb125a35
cnf-certification-test 7bfc20a
 -> https://www.distributed-ci.io/topics/afbb1c48-277e-4ddb-aa14-d24fe27434a7/components/fc242583-ea70-4986-bca6-de5f74376928

URL now points to the correct commit and can be opened correctly.

I've addressed @manurodriguez's comments, just confirming what he was asking. If you don't mind, I'll close this change since it has passed two days and there are no more feedback.

roles/cnf_cert/tasks/pre-run.yml
manurodriguez: Thanks for the clarification, now I see where the magic happens : )

PR #29, opened 2023-11-01
use opm client from stable 4.13 release for fbc_catalog

manurodriguez:
build-depends: 30006

the opm client in 4.14 is not returning olm.bundle.object property from operator images, required by fbc_catalog role to generate the catalog and be able to install operators


betoredhat (approved):
Thanks, LGTM

manurodriguez:
Thanks

PR #31, opened 2023-11-02
Use opm new stable release in fbc and preflight

tonyskapunk:
Remove the unused deprecated olm.bundle.object, is not used nor needed in FBC creation.

New render format deprecates olm.bundle.object but some another field could provide the list of images used (relatedImages), using that instead.

build-depends: 30017


tonyskapunk:
check dallas ocp-4.13-vanilla

betoredhat (approved):
LGTM!, thanks

Good catch

betoredhat:
check dallas ocp-4.7-vanilla tnf-test-cnf

PR #33, opened 2023-11-06
Do not use cache during image build

betoredhat:
- No cached images and forcing to always pull to avoid working with outdated version of opm image

### DIFF
--- a/roles/cnf_cert/tasks/pre-run.yml
+++ b/roles/cnf_cert/tasks/pre-run.yml
@@ -130,6 +130,7 @@
     shell: |
       set -x
       podman build -t {{ tnf_image }} \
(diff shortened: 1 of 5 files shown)


tonyskapunk:
Thanks @betoredhat do you mind doing the same on the other roles? I found at least these:

roles/fbc_catalog/tasks/main.yml
roles/preflight/tasks/prepare_preflight_image.yml
roles/operator_sdk/tasks/build_scorecard_image.yml
roles/cnf_cert/tasks/pre-run.yml
roles/opcap_tool/tasks/build.yml

tonyskapunk (approved):
LGTM, thanks!

PR #34, opened 2023-11-06
Adding a retry to account for timeout errors when pulling container images

AbhishekBandarupalle:
While doing container certification tests for container images stored in an artifactory repository - I've seen timeout errors when podman tries to pull the images. 

The same container check executes successfully on a retry. Hence, I think we should add a retry option to account for these errors.

### DIFF
--- a/roles/preflight/tasks/test_preflight_check_container_one_image.yml
+++ b/roles/preflight/tasks/test_preflight_check_container_one_image.yml
@@ -80,7 +80,7 @@
           register: podman_rc
           retries: 2
           delay: 120
-          until: podman_rc.rc == 0
+          until: podman_rc is succeeded
       rescue:
         - name: Do not fail when preflight check container throws an error
           debug:
--- a/roles/preflight/tasks/test_preflight_check_container_one_image.yml
+++ b/roles/preflight/tasks/test_preflight_check_container_one_image.yml
@@ -77,8 +77,10 @@
             {% if cert_project_id | default('') | length %}
             --submit
             {% endif %}
+          register: podman_rc
           retries: 2
           delay: 120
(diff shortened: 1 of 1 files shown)


tonyskapunk (changes requested):
Thanks @AbhishekBandarupalle , left a comment inline that's required for the retries to work.

roles/preflight/tasks/test_preflight_check_container_one_image.yml
```diff
@@ -77,6 +77,8 @@
             {% if cert_project_id | default('') | length %}
             --submit
             {% endif %}
+          retries: 2
+          delay: 120
```
tonyskapunk, line 81: Needs an `until` condition in here otherwise the retry won't work [1]

[1] https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met

AbhishekBandarupalle: Added

tonyskapunk (commented):
Looks good, but I think you will add another change based on our conversation, I'll approve after that. Thanks!

tkrishtop (changes requested):
Hi @AbhishekBandarupalle, thank you for the contribution!

> While doing container certification tests for container images stored in an artifactory repository - I've seen timeout errors when podman tries to pull the images.

Could you please provide a link to the failing DCI job? I'm not sure that this retry should be added on the DCI side and not on the Preflight side. 
We usually avoid patching the external tool using DCI wrappers.

AbhishekBandarupalle:
Closing this PR as per my discussion with @tkrishtop.
Will open the issue with preflight.

PR #36, opened 2023-11-07
Allow multiple values_files

betoredhat:
Some charts requires mutiple custom files with configuration values

### DIFF
--- a/roles/chart_verifier/tasks/tests.yml
+++ b/roles/chart_verifier/tasks/tests.yml
@@ -71,7 +71,7 @@
       get_url:
         url: "{{ item }}"
         dest: "{{ work_chart_dir }}/"
-        mode: "0750"
+        mode: "0640"
       loop: "{{ chart.values_file }}"
       when:
         - chart.values_file is defined
@@ -80,7 +80,6 @@
 
 - name: "Process local chart artifacts for {{ chart.chart_file }}"
   block:
-
     - name: "Check chart_file stat"
       stat:
         path: "{{ chart.chart_file }}"
--- a/roles/chart_verifier/tasks/tests.yml
+++ b/roles/chart_verifier/tasks/tests.yml
@@ -90,29 +90,25 @@
         - chart_stat is defined
         - chart_stat.stat.exists
 
-    - name: "Check values_file stat"
+    - name: "Check if chart.values_file exist"
       stat:
-        path: "{{ chart.values_file }}"
-      register: values_stat
-      when:
-        - chart.values_file is defined
+        path: "{{ item }}"
+      loop: "{{ chart.values_file }}"
+      register: file_stats
 
-    - name: "Chart values file path not found"
+    - name: "Fail if any values file is missing"
       fail:
-        msg: "values_file cannot be found - {{ chart.values_file }}"
-      when:
-        - values_stat is defined
-        - chart.values_file is defined
-        - values_stat.stat.exists | bool == false
+        msg: "File {{ item.item }} does not exist."
+      loop: "{{ file_stats.results }}"
+      when: item.stat.exists == false
 
     - name: "Copy values_file to working directory"
       copy:
-        src: "{{ chart.values_file }}"
+        src: "{{ item }}"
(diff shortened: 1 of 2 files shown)


roles/chart_verifier/tasks/tests.yml
```diff
@@ -90,29 +90,25 @@
         - chart_stat is defined
         - chart_stat.stat.exists
 
-    - name: "Check values_file stat"
+    - name: "Check if chart.values_file exist"
       stat:
```
nsilla, line 94: Hi Beto,
What if the values_file list include a web URL? This task will return false even if the URL is valid.

roles/chart_verifier/tasks/tests.yml
```diff
@@ -90,29 +90,25 @@
...
-        path: "{{ chart.values_file }}"
-      register: values_stat
-      when:
```
nsilla, line 97: Also, I guess we should keep the "when" condition here as long as the values_file variable is optional.

roles/chart_verifier/tasks/tests.yml
```diff
@@ -140,7 +136,11 @@
       -v {{ work_chart_dir }}:/app/chartverifier:Z
       {{ chart_verifier_image }}
       verify --set profile.vendorType=partner
-      {% if chart.values_file is defined %} -F /app/chartverifier/{{ chart.values_file | basename }} {% endif %}
+      {%- if chart.values_file is defined %}
```
ramperher, line 139: Should we check here that chart.values_file is iterable?

ramperher (commented):
One comment from my side

roles/chart_verifier/tasks/tests.yml
betoredhat: This block is only is applied only for local files. See line 113.

roles/chart_verifier/tasks/tests.yml
betoredhat: Thanks, good idea. Added the validation at the beginning

roles/chart_verifier/tasks/tests.yml
betoredhat: This block is to check local files

roles/chart_verifier/tasks/tests.yml
```diff
@@ -34,35 +43,36 @@
       when:
         - file_response.status | int != 200
 
-    - name: "Check if the chart values file exists"
+    - name: "Download chart file"
+      get_url:
+        url: "{{ chart.chart_file }}"
+        dest: "{{ work_chart_dir }}/"
+        mode: "0750"
```
tonyskapunk, line 50: Don't think we need it +w, can we do 0640?

roles/chart_verifier/tasks/tests.yml
```diff
@@ -90,29 +100,29 @@
...
         - chart.values_file is defined
-        - values_stat.stat.exists | bool == false
+        - item.stat.exists == false
 
     - name: "Copy values_file to working directory"
       copy:
-        src: "{{ chart.values_file }}"
+        src: "{{ item }}"
         dest: "{{ work_chart_dir }}/"
+      loop: "{{ chart.values_file }}"
```
tonyskapunk, line 123: Is this one looping through a list that could have URLs? If so, would need to take the basename of the URL

roles/chart_verifier/tasks/tests.yml
betoredhat: done

roles/chart_verifier/tasks/tests.yml
betoredhat: This block is to process local files. The chart_file location determines if the files are local or remote. If the chart_file is local, the values files must be local too. If the chart file is remote, the files must be remote. That has been clarified in the readme and is the use case exposed by the requester.

Maybe in the next iteration, we can add support for files in remote (uri) and local (copy/stat).

Thanks

tonyskapunk (approved):
LGTM, thanks!

PR #40, opened 2023-11-09
Separate args for sed replacement in galaxy

tonyskapunk:
sed `-i ` option could have an argument, need to pass as two diff options to sed.


betoredhat (approved):
LGTM

manurodriguez (approved):
:+1:

PR #41, opened 2023-11-09
Add required README files for publishing in galaxy

tonyskapunk:
Seems like we are hitting this issue: https://github.com/ansible/galaxy/issues/2438

Some people have reported being able to publish collections by adding the missing READMEs in their roles, but some others have reported that this is not working any more, starting with this.


betoredhat (approved):
LGTM, very verbose

betoredhat (approved):
:+1:

PR #14, opened 2023-10-17
Remove custom virt module

tonyskapunk:
Use community.libvirt instead of custom module
Baremetal deploy role used a custom libvirt module, which was a copy of
an old libvirt module. The current functionality of the community module
is enough to stop using the custom module.

Test-Hints: sno

PR #44, opened 2023-11-12
Fix Ansible test sanity issues for collection

sshnaidm:
Add CI job which will check PRs for Ansible sanity tests.
CNF-10325


tkrishtop:
check dallas ocp-4.12-vanilla example-cnf

plugins/filter/ocp_compatibility.py
```diff
@@ -12,22 +17,28 @@ def ocp_compatibility(self, after_api, curr_version, junit_ocp_file):
...
+                api['removedInRelease']) == version}
             for version in failed_versions
         }
 
         # convert major.minor version to semvers' major.minor.patch
-        ocp2semver = lambda x : VersionInfo.parse(x + ".0")
+        def ocp2semver(x):
+            return VersionInfo.parse(x + ".0")
         # convert semver's major.minor.patch to OCP major.minor
-        semver2ocp = lambda x : '.'.join(str(x).split('.')[:-1])
```
tkrishtop, line 26: Hi, @sshnaidm is lambda-function always a wrong choice in ansible filters?

roles/odf_setup/tests/inventory
tkrishtop, line 1: having empty inventories could be risky, wdyt?

plugins/filter/ocp_compatibility.py
sshnaidm: **Assigning** lambda function is wrong according to [PEP-8 ](https://peps.python.org/pep-0008/#programming-recommendations), it triggers [lambda-assignment (E731)](https://docs.astral.sh/ruff/rules/lambda-assignment/) error. Using it _inside_ is not a problem of course.

roles/odf_setup/tests/inventory
sshnaidm: I'm not familiar with this role and what it does here, probably worth to remove? I just removed execution bits from it and other YAML files, it triggered an error.

tonyskapunk (changes requested):
Hi @sshnaidm thanks for the contribution, some of the changes in here duplicates DCI-655 there's a change already opened: 
https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/28

Some of the pieces that are not considered in #28 are plugins and scripts, I see you have included them in here, would you mind adjusting this change to target **only** plugins and scripts instead?

I'll also request to avoid IPI plugins for now, as you could see #14 is to remove one of them, but we would like to tackle them until we move to collections in baremetal-deploy repo, I have a branch where I'm working this: https://github.com/openshift-kni/baremetal-deploy/compare/master...tonyskapunk:baremetal-deploy:use_collection

Thanks again!

sshnaidm:
> Hi @sshnaidm thanks for the contribution, some of the changes in here duplicates DCI-655 there's a change already opened: #28
> 
> Some of the pieces that are not considered in #28 are plugins and scripts, I see you have included them in here, would you mind adjusting this change to target **only** plugins and scripts instead?

It's a little different approach in two patches, I ran ansible-test sanity checks, which are different from ansible-lint checks. They are not overlapping usually, just a common points maybe when there is an obvious issue. So these are different checks for different purposes, for publishing to galaxy and certification you need ansible-test. also it keeps all - roles, playbooks, modules etc with a requirements of Ansible. Ansilbe linter is supposed to keep good practices and recommended way for roles and playbooks.
So it should complete each other, it doesn't replace.

> I'll also request to avoid IPI plugins for now, as you could see #14 is to remove one of them, but we would like to tackle them until we move to collections in baremetal-deploy repo, I have a branch where I'm working this: [openshift-kni/baremetal-deploy@master...tonyskapunk:baremetal-deploy:use_collection](https://github.com/openshift-kni/baremetal-deploy/compare/master...tonyskapunk:baremetal-deploy:use_collection)

Sorry, not sure I understand. Should I remove something from this patch..?

tonyskapunk:
> > Hi @sshnaidm thanks for the contribution, some of the changes in here duplicates DCI-655 there's a change already opened: #28
> > Some of the pieces that are not considered in #28 are plugins and scripts, I see you have included them in here, would you mind adjusting this change to target **only** plugins and scripts instead?
> 
> It's a little different approach in two patches, I ran ansible-test sanity checks, which are different from ansible-lint checks. They are not overlapping usually, just a common points maybe when there is an obvious issue. 

You're right, I thought sanity checks included ansible-lint. In that case please take a look to this branch: https://github.com/redhatci/ansible-collection-redhatci-ocp/compare/main...sanity_checks.  It's a simple version of sanity checks and will not hold anyone for changes just yet, but we can enable whenever the repo is ready.

> So these are different checks for different purposes, for publishing to galaxy and certification you need ansible-test. 

Not sure it it's needed for publishing to galaxy, we are already doing it. And certification is not a goal for the time being.

> > I'll also request to avoid IPI plugins for now, as you could see #14 is to remove one of them, but we would like to tackle them until we move to collections in baremetal-deploy repo, I have a branch where I'm working this: [openshift-kni/baremetal-deploy@master...tonyskapunk:baremetal-deploy:use_collection](https://github.com/openshift-kni/baremetal-deploy/compare/master...tonyskapunk:baremetal-deploy:use_collection)
> 
> Sorry, not sure I understand. Should I remove something from this patch..?

Yes, changes to beremetal deploy, that includes plugins: The only 2 modules available (virt, nmcli). And roles: installer and node_prep

Thanks!

sshnaidm:
> > > Hi @sshnaidm thanks for the contribution, some of the changes in here duplicates DCI-655 there's a change already opened: #28
> > > Some of the pieces that are not considered in #28 are plugins and scripts, I see you have included them in here, would you mind adjusting this change to target **only** plugins and scripts instead?
> > 
> > 
> > It's a little different approach in two patches, I ran ansible-test sanity checks, which are different from ansible-lint checks. They are not overlapping usually, just a common points maybe when there is an obvious issue.
> 
> You're right, I thought sanity checks included ansible-lint. In that case please take a look to this branch: [main...sanity_checks](https://github.com/redhatci/ansible-collection-redhatci-ocp/compare/main...sanity_checks). It's a simple version of sanity checks and will not hold anyone for changes just yet, but we can enable whenever the repo is ready.

Sorry, what do you mean? Isn't repo ready now? I've already fixed all issues for you in this pr :smile:  And it's better to have a job to make sure all next merges will go through this check.

> > So these are different checks for different purposes, for publishing to galaxy and certification you need ansible-test.
> 
> Not sure it it's needed for publishing to galaxy, we are already doing it. And certification is not a goal for the time being.

It's better to have sanity clear than not to have sanity clear.

tonyskapunk:
> Sorry, what do you mean? Isn't repo ready now?

I requested to not include changes to baremetal deploy roles and plugins.  By not including them the sanity checks will block PRs, that's why I mentioned that the branch https://github.com/redhatci/ansible-collection-redhatci-ocp/compare/main...sanity_checks has a way to not hold changes and that we can enable when is time.

> I've already fixed all issues for you in this pr 😄 And it's better to have a job to make sure all next merges will go through this check.

This change is still changing baremetal-deploy plugins and roles, I've requested them to not be modified for the time being.

> It's better to have sanity clear than not to have sanity clear.

I'm not disagreeing, I'm saying that sanity checks are not a blocker for publishing in galaxy.

I'm in favor of sanity checks :wink:  that's why I had them already in a branch, but I'm asking to simplify them https://github.com/redhatci/ansible-collection-redhatci-ocp/compare/main...sanity_checks

PR #45, opened 2023-11-13
setup_minio: add variable to specify service type

manurodriguez:


roles/setup_minio/README.md
```diff
@@ -55,6 +57,7 @@ See below how to consume the services provided by Minio.
 
 ## From outside the cluster:
 
+> NOTE: The variable `sm_service_type: LoadBalacer` is required to enable external access to the service.
```
tonyskapunk, line 60: Sorry I didn't see there was another line below this that I assume it shouldn't be part of the NOTE, in that case it needs a new line:
```suggestion
> NOTE: The variable `sm_service_type: LoadBalacer` is required to enable external access to the service.

```

If the line below is also part of the NOTE then disregard this suggestion

manurodriguez: np, and correct the lines below should not be part of the NOTE, they are related but it's a separate sentence, so I'm adding the new line.

tonyskapunk (approved):
LGTM

manurodriguez:
Last job failed because when using a service type ClusterIP there is no NodePort
```
[kni@provisioner.cluster4.dfwt5g.lab ~]$ oc -n minio get svc
NAME            TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
minio-service   ClusterIP   172.30.107.230   <none>        9000/TCP   7m22s

[kni@provisioner.cluster4.dfwt5g.lab ~]$ oc -n minio get svc  minio-service -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2023-11-13T21:16:24Z"
  name: minio-service
  namespace: minio
  resourceVersion: "62003"
  uid: d7ec5c43-7b6e-46e9-9220-122c2b536a33
spec:
  clusterIP: 172.30.107.230
  clusterIPs:
  - 172.30.107.230
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: http
    port: 9000
    protocol: TCP
    targetPort: 9000
  selector:
    app: minio
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```
NodePort is only added when service type is NodePort or LoadBalancer
```
[kni@provisioner.cluster4.dfwt5g.lab ~]$ oc -n minio get svc
NAME            TYPE       CLUSTER-IP       EXTERNAL-IP   PORT(S)          AGE
minio-service   NodePort   172.30.251.192   <none>        9000:30521/TCP   8s

[kni@provisioner.cluster4.dfwt5g.lab ~]$ oc -n minio get svc minio-service -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2023-11-13T22:04:42Z"
  name: minio-service
  namespace: minio
  resourceVersion: "79039"
  uid: eb1f7a21-67f4-4aef-b961-eaf7d404fd2e
spec:
  clusterIP: 172.30.251.192
  clusterIPs:
  - 172.30.251.192
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: http
    nodePort: 30521
    port: 9000
    protocol: TCP
    targetPort: 9000
  selector:
    app: minio
  sessionAffinity: None
  type: NodePort
status:
  loadBalancer: {}
```
so now I'm using service type NodePort as default and I updated the readme and plays accordingly:

roles/setup_minio/tasks/install.yml
```diff
@@ -124,51 +124,54 @@
...
+        http://{{ random_worker_ip }}:{{ node_port }} \
+        {{ sm_access_key_id }} {{ sm_access_key_secret }} --api S3v4;
+        {{ minio_dir }}/mc mb --ignore-existing {{ instance }}/{{ sm_bucket_name }}
+      register: minio_cmd
+      retries: 18
+      delay: 5
+      until: minio_cmd is not failed
+      notify:
+        - "Delete temp directory"
+  when: sm_service_type == "NodePort"
```
betoredhat, line 176: I have concerns about this only working with nodePort now, as you mentioned with LoadBalancer we also have IPs Nodes IP addresses. We may need to check for a way to always be able to reach the Minio endpoint from the Jump host. Or maybe use the API to create the bucket from inside the cluster using the API.

manurodriguez: Yes, we can add LoadBalancer type in the condition too because it provides a node-port, ideally we should use the external IP provided by MetalLB, but in our case is not reachable from the jumphost when it's configured in BGP mode, and might be the same for other users, but then we'll need a way to reach the endpoint when using ClusterIP option, and as you said maybe using the API internally. Is this bucket creation only a external test? If workloads want to consume minio, they somehow use the internal endpoint right? If so we can do something : )

manurodriguez: I synced with Beto, we'll merge this and improve the part that creates the bucket to consider all service type option in another change, we were thinking to create the bucket using the internal service endpoint.

PR #46, opened 2023-11-14
Add prune catalog role

betoredhat:


roles/prune_catalog/tasks/main.yml
```diff
@@ -0,0 +1,126 @@
...
+
+    - name: "Copy configs.Dockerfile to temp directory"
+      copy:
+        content: |
+          FROM quay.io/operator-framework/opm:latest
+          ENTRYPOINT ["/bin/opm"]
+          CMD ["serve", "/configs"]
+          ADD configs /configs
+          LABEL operators.operatorframework.io.index.configs.v1=/configs
+          LABEL maintainer="Telco DCI team"{% if pc_expire %} quay.expires-after={{ pc_expire_time }}{% endif %}
```
tonyskapunk, line 99: Do you know if we need the maintainer label? if not can we remove it or make it more generic like "redhatci.ocp"

betoredhat: Found that the maintainer property is required to see the expired, I allowed the maintainer to be defined in a variable with a default. 

Tested the change from 4.13-4.15.

Thanks

tonyskapunk (approved):
So far LGTM :)

roles/prune_catalog/tasks/main.yml
```diff
@@ -0,0 +1,126 @@
...
+      --authfile $DOCKER_CONFIG/config.json \
+      --tls-verify=false docker://{{ pc_source_catalog }};
+    else
+      skopeo inspect --no-tags \
+      --tls-verify=false docker://{{ pc_source_catalog }};
+    fi
+  register: skopeo_result
+  retries: 2
+  delay: 10
+  ignore_errors: true
```
tonyskapunk, line 26: I think we can stop ignoring errors, and fail if the catalog is not pullable, in the end all the rest of the tasks won't run if this does not succeed.

betoredhat: In this case it is important to simple do nothing and not fail, that will allow the agent to detect that that image was not created on that iteration and fallback to the previous minor version. I do not think doing nothing on this role if the image is not pullable.

tonyskapunk: I see, this is more a requirement on the agent, rather than the role itself :thinking: 

Would it be worth doing a conditional for the failure instead?

betoredhat: @tonyskapunk removed the ignore_errors. The 2 last jobs test the regular use case and the fallback in case there are operators missing.

tonyskapunk (approved):
LGTM, thanks!

PR #48, opened 2023-11-14
Change versioning

tonyskapunk:
The current versioning is considered pre-release. Switching to use a slightly different format to make every release a patch release.

The version of the patch release is now the epoch of the latest commit date merged into main

Before: 0.2.1-{%Y%m%D%H%M}git{SHA}
Now: 0.2.{%s}+{%Y%m%D%H%M}.git{SHA}

The + delimits the version from metadata for the release

> NOTE: With this new version format I'm seeing 500/400 errors from galaxy when testing. Despite the errors the collection is published but this affects the state of the GHA. It has been reported in https://forum.ansible.com/t/http-500-400-errors-when-publishing-collections/2380

> NOTE: Issue opened for the 500 error https://github.com/ansible/galaxy/issues/3287

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,12 +11,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Add release, date and commit hash to version in galaxy.yml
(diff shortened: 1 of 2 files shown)


thekad (approved):
:ship:

PR #54, opened 2023-11-21
Remove tnf backwards compatibility

ramperher:
Only keeping compatibility with tnf >= v4.4.0

Including internal docs change: https://github.com/dci-labs/dallas-internal-docs/pull/122


ramperher:
All tests are working fine. The job that is failing is because I tried with tnf v4.3.0, which is lower than v4.4.0, so it has to fail.

Other jobs tested v4.5.5, v4.4.0 and HEAD.

sebrandon1 (approved):

roles/cnf_cert/templates/tnf_config.yml.j2
```diff
@@ -65,23 +38,6 @@ targetCrdFilters:
 {% endif %}
 {% endfor %}
 
-{% if test_network_function_version is version('v4.4.0', '<=') %}
```
manurodriguez, line 68: This condition checks for tnf <= v4.4.0, and if we remove it then I guess is something we really do not need right?  just curious since this PR is just keeping compatibility for  >= v4.4.0

manurodriguez (approved):
Thanks, LGTM

PR #55, opened 2023-11-21
switch to a versioning based on the UNIX epoch

fredericlepied:
To be in sync with the versionning for Ansible Galaxy.

Depends-On: https://softwarefactory-project.io/r/c/dci-packaging/+/30166

### DIFF
--- a/ansible-collection-redhatci-ocp.spec
+++ b/ansible-collection-redhatci-ocp.spec
@@ -3,9 +3,9 @@
 %global forgeurl https://github.com/%{org}/%{repo}
 
 Name:           %{repo}
-Version:        0.2.0
-Release:        1.VERS%{?dist}
-Summary:        Redhat OCP Collection for ansible
+Version:        0.2.EPOCH
+Release:        VERS%{?dist}
+Summary:        Red Hat OCP CI Collection for Ansible
 
 License:        GPL-2.0-or-later and Apache-2.0
 URL:            %{forgeurl}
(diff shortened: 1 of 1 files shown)


fredericlepied:
recheck

fredericlepied:
recheck

fredericlepied:
recheck

.github/workflows/publish.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           z=$(TZ=UTC0 git show ${{ github.sha }} --no-patch --format=%cd --date=format-local:'%s');
           ts=$(TZ=UTC0 git show ${{ github.sha }} --no-patch --format=%cd --date=format-local:'%Y%m%d%H%M');
           sha=$(echo "${{ github.sha }}" | cut -c1-7);
-          sed -i -r "s/^(version: .*)\.0$/\1.${z}+${ts}.git${sha}/" galaxy.yml;
+          sed -i -r "s/^(version: .*)\.EPOCH$/\1.${z}+${ts}.git${sha}/" galaxy.yml;
```
tonyskapunk, line 19: I don't think this is needed because the patch version is in galaxy.yml

fredericlepied: you are right. what about modifying the galaxy.yml to have .EPOCH at the end to be consistent?

tonyskapunk: It would prevent someone from building it locally and it will also get in the way of the CI https://github.com/redhat-cip/dci-openshift-agent/blob/master/test-runner#L466

tonyskapunk (approved):
LGTM

PR #57, opened 2023-11-21
Add sanity check on PRs

tonyskapunk:
Using sanity checks as CI testing is a good practice for community package collections[1].
Add initially without stopping CI to find the errors that need to be fixed an once they are cleaned, enable them.

[1] https://docs.ansible.com/ansible/latest/community/collection_contributors/collection_requirements.html#ci-testing

Test-Hints: no-check


.github/workflows/sanity-checks.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/setup-python@v3
+        with:
+          python-version: 3.11
+
+      - name: Install ansible-base ${{ matrix.ansible }}
+        run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
+      
+      # Always succeeding sanity test until all errors are fixed
+      - name: Run sanity test
+        run: ansible-test sanity -v --docker --color --coverage --failure-ok
```
fredericlepied, line 32: could we run it twice once on the PR branch and once on the  main branch and compare the number of  ERROR lines to be sure we are always reducing the number?

fredericlepied (changes requested):

.github/workflows/sanity-checks.yml
tonyskapunk: The GHA is now counting the errors and showing them in the summary: https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/6965977130

I could add also the workflow in the merge of the main branch, at least would give us an idea on errors there vs a PR, would that cover what you have in mind?

fredericlepied: I pushed a commit with a rebased branch to show what I would like to see.

fredericlepied:
I put back the draft status as I added a regression and an improvement in the last commit just to test the process. We must remove this commit before merging.

fredericlepied (approved):

fredericlepied (approved):

PR #66, opened 2023-11-27
check_resource | Remove reference to OCPBUGS-17112 and generalize wait-mcp

ramperher:
- OCPBUGS-17112 has been closed, so removing the references to that issue
- Running a new retry in case no nodes are in SchedulingDisabled status is not harmful, we can do it anyway.


manurodriguez (approved):
Thanks, LGTM

PR #67, opened 2023-11-27
install-config.j2: support disableCertificateVerification for irmc (#…

tonyskapunk:
baremetal-deploy sync


manurodriguez (approved):
:+1:

PR #70, opened 2023-11-28
Improve sanity test summary

tonyskapunk:
Make more clear the list of fixed improvements in the change and fix the way to count regressions.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -54,7 +54,7 @@ jobs:
           echo "\`\`\`diff" >> ${GITHUB_STEP_SUMMARY}
           diff -u0 branch.errors main.errors | grep '^-[^-]' | tee -a ${GITHUB_STEP_SUMMARY}
           echo "\`\`\`" >> ${GITHUB_STEP_SUMMARY}
-          if grep -q '^-[^-]' <<< $(diff -u0 branch.errors main.errors); then
+          if diff -u0 branch.errors main.errors | grep -q '^-[^-]'; then
              echo "> Fix the regression errors listed above" | tee -a ${GITHUB_STEP_SUMMARY}
              exit 1
           fi
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -54,7 +54,7 @@ jobs:
           echo "\`\`\`diff" >> ${GITHUB_STEP_SUMMARY}
           diff -u0 branch.errors main.errors | grep '^-[^-]' | tee -a ${GITHUB_STEP_SUMMARY}
           echo "\`\`\`" >> ${GITHUB_STEP_SUMMARY}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -48,14 +48,14 @@ jobs:
...
-          diff -u0 branch.errors main.errors | grep '^+[^+]' | tee -a ${GITHUB_STEP_SUMMARY}
+          diff -u0 branch.errors main.errors | grep '^+[^+]' | sed -e 's/ERROR/FIXED/' | tee -a ${GITHUB_STEP_SUMMARY}
           echo "\`\`\`" >> ${GITHUB_STEP_SUMMARY}
           echo "## Regressions are listed below" | tee -a ${GITHUB_STEP_SUMMARY}
           echo "\`\`\`diff" >> ${GITHUB_STEP_SUMMARY}
           diff -u0 branch.errors main.errors | grep '^-[^-]' | tee -a ${GITHUB_STEP_SUMMARY}
           echo "\`\`\`" >> ${GITHUB_STEP_SUMMARY}
-          if ! diff -q main.errors branch.errors;  then
-             echo "> Fix the regressions listed above (starting with a '-')" | tee -a ${GITHUB_STEP_SUMMARY}
+          if [[ $(diff -u0 branch.errors main.errors | grep -c '^-[^-]') != 0 ]];  then
```
fredericlepied, line 57: why not just checking the status of the grep command?

tonyskapunk: Because the comparison has to be on the diff.

Is possible the branch.errors still contain errors from main.errors, those are the improvements.

fredericlepied: even with a pipe, it works as it checks the status of the last command.

tonyskapunk: Got it, I did the grep -q, same effect as the pipe

.github/workflows/pr.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           echo "\`\`\`diff" >> ${GITHUB_STEP_SUMMARY}
           diff -u0 branch.errors main.errors | grep '^-[^-]' | tee -a ${GITHUB_STEP_SUMMARY}
           echo "\`\`\`" >> ${GITHUB_STEP_SUMMARY}
-          if [[ $(diff -u0 branch.errors main.errors | grep -c '^-[^-]') != 0 ]];  then
+          if grep -q '^-[^-]' <<< $(diff -u0 branch.errors main.errors); then
```
thekad, line 57: that's an interesting way to redirect the output, I've always used something like `grep foo <(command)` because that creates a temp file descriptor and treats it as a file

tonyskapunk: No problem doing this, but I submitted the change before reading your previous comment, sending it this way now

fredericlepied (approved):

PR #73, opened 2023-11-29
Fix sanity issues for deprecated-api role

tkrishtop:


tkrishtop:
discarding transition from Python3 to Python2, will open another PR to fix the sanity issues in the role

PR #77, opened 2023-12-04
Remove metadata from patch version

tonyskapunk:
Metadata in the patch version is not supported by ansible galaxy client version 2.9[1].
While the metadata is supported in SemVer, the ansible galaxy client filters out and only allows strict versions to be installed. This is not an issue in 2.10+ but we provide support for ansible 2.9

[1] https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/galaxy/collection.py#L348-L350

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,12 +11,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Add epoch from commit date as patch version and metadata in galaxy.yml
+      - name: Add epoch from commit date as patch version in galaxy.yml
         run: >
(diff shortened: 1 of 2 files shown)


thekad:
a version `0.3.0-20231221042253` would still be valid, no? you'd only lose the git commit metadata, but the timestamp is still appropriate as the revision part of the patch, iirc

tonyskapunk:
> 0.3.0-20231221042253

This is [valid semver](https://jubianchi.github.io/semver-check/#/version/0.3.0-20231221042253) but is a pre-release version, that is not supported in 2.9 either, i.e. you can't install a pre-release in 2.9

fredericlepied (approved):

PR #78, opened 2023-12-04
Always delete temp artifacts

betoredhat:
Notify is not been executed, and may fill the /tmp directory with temporary artifacts


betoredhat:
check dallas ocp-4.14-vanilla

roles/prune_catalog/tasks/main.yml
```diff
@@ -49,7 +49,7 @@
 
     - name: "Download stable opm client"
       vars:
-        ocp_clients_url: "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/opm-linux.tar.gz"
+        ocp_clients_url: "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/opm-linux.tar.gzs"
```
manurodriguez, line 52: Beto, is this a typo? I do not see the file `opm-linux.tar.gzs` with an `s` at the end

betoredhat: Sorry, that was part of my test to make it fail and have the always executed. Thanks!

betoredhat:
check dallas ocp-4.14-vanilla

roles/prune_catalog/tasks/main.yml
```diff
@@ -126,11 +126,14 @@
           -t {{ pc_destination_catalog }}
           -f {{ pc_tmp_dir }}/configs.Dockerfile .
       register: build_result
-      changed_when: build_result.rc != 0
-      notify:
-        - "Delete temp directory"
```
tonyskapunk, line 131: I think it was not called because the `changed_when` condition was never met unless failed, now that the `changed_when` is set for success, it should trigger the notification, or did you face any issue that way?

betoredhat: Yes, that was the issue. But IMO it is better not with the block

betoredhat: The previous job failure is not related to this change.

tonyskapunk (approved):
:+1:

PR #82, opened 2023-12-05
Remove MCE annotation

betoredhat:
Confirmed that setting the MCE version to install is not needed. The Multiclusterhub will select the version of the MCE operator


manurodriguez (approved):
LGTM

betoredhat:
check dallas ocp-4.14-acm-hub

PR #93, opened 2023-12-13
bugfixes for Agent Based Installer

nocturnalastro:
Cherry picking bug fix from crucible. 

Test-Hints: assisted-abi

### DIFF
--- a/roles/generate_agent_iso/defaults/main.yml
+++ b/roles/generate_agent_iso/defaults/main.yml
@@ -2,7 +2,6 @@ generated_dir: "{{ repo_root_path }}/generated"
 manifests_dir: "{{ generated_dir }}/{{ cluster_name }}"
 download_agent_dest_file: "{{ discovery_iso_name }}"
 download_dest_path: "{{ iso_download_dest_path | default('/opt/http_store/data') }}"
-config_file_path: /tmp/wip/config
 arch: x86_64
 version_filter: "[?(openshift_version == '{{ openshift_version }}') && (cpu_architecture == '{{ arch }}')]"
 release_image: "{{ (assisted_installer_release_images | json_query(version_filter))[0].url }}"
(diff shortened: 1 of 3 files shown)


nocturnalastro:
@arjuhe Can you check this for me.

thekad (dismissed):

fredericlepied (changes requested):
Could we make sure the DCI job for assisted-abi works?

ramperher (commented):
The job is failing in the task just after the one that has been changed in generate_agent_iso: https://www.distributed-ci.io/jobs/6caac908-d818-4f82-beb8-5a143d1675ae/jobStates?sort=date&task=1fe56a9d-3744-4907-bc09-d773d4ff85d9

I can see this:

```
"level=error msg=failed to write asset (Agent Installer ISO) to disk: cannot generate ISO image due to configuration errors", "level=fatal msg=failed to fetch Agent Installer ISO: failed to fetch dependency of \"Agent Installer ISO\": failed to generate asset \"Agent Installer Artifacts\": cpio: write too long"
```

Don't know if it's related to pullsecrets, I'd relaunch a job to see if it happens again, and if true, then I believe this should be reviewed.

nocturnalastro:
I think I see the issue. `XDG_RUNTIME_DIR: "{{ config_file_path }}"` also needs to change to `XDG_RUNTIME_DIR: "{{ generated_dir }}"`. I guess config_file_path wasn't cleaned up when we tested.

nocturnalastro:
I haven't been able to recreate the issue. However I have a suspicion that is the issue so I'll add a commit to change that an see if that corrects the issue.

nocturnalastro:
Looks like I was right. I'll get that fix merged back into crucible (for while we still haven't converted over to the collection).

fredericlepied (approved):

ramperher (approved):
LGTM

PR #96, opened 2023-12-13
Enabling ansible-lint action

thekad:
Setting it to be optional for now. Also adding the requirements.yml needed for all roles in this repo


thekad:
@tonyskapunk @fredericlepied PTAL I think this is what we talked about during our sync

fredericlepied (approved):

PR #98, opened 2023-12-15
Disable GHA annotations from the script

thekad:
ansible-lint has automatic GHA detection that annotates source code on each PR. This feature is still in Beta, disabling for now and will reenable later

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -33,9 +33,11 @@ jobs:
           git checkout -b branch
           git fetch --unshallow origin main
           cmd="ansible-lint --parseable --nocolor --profile=shared"
+          unset GITHUB_ACTIONS
           $cmd > branch.output || echo "Branch linting output written"
           git checkout main
           $cmd > main.output || echo "Main linting output written"
+          export GITHUB_ACTIONS=true
           set +ex
           diff -u0 branch.output main.output > diff.output || echo "Wrote diff file"
           echo "## Improvements over main branch:" | tee -a ${GITHUB_STEP_SUMMARY}
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -33,22 +33,24 @@ jobs:
           git checkout -b branch
           git fetch --unshallow origin main
           cmd="ansible-lint --parseable --nocolor --profile=shared"
+          # don't want to annotate through GHA
           unset GITHUB_ACTIONS
           $cmd > branch.output || echo "Branch linting output written"
           git checkout main
           $cmd > main.output || echo "Main linting output written"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -33,22 +33,30 @@ jobs:
...
-          grep '^-[^-]' diff.output | sed -e 's/^-/-ERROR: /' | tee -a ${GITHUB_STEP_SUMMARY}
+          grep '^-' diff.output | sed -e 's/^-/-ERROR: /' | tee -a ${GITHUB_STEP_SUMMARY}
           echo '```' >> ${GITHUB_STEP_SUMMARY}
-          if grep -q '^-[^-]' diff.output; then
+          if grep -q '^-' diff.output; then
              echo "> Fix regressions listed above" | tee -a ${GITHUB_STEP_SUMMARY}
-             exit 1
+             regressions="$(grep '^-' diff.output | wc -l | cut -f 1 -d ' ')"
+             improvements="$(grep '^+' diff.output | wc -l | cut -f 1 -d ' ')"
+             # if ratio of regressions > improvements then fail this
```
fredericlepied, line 56: counting doesn't implement a quality bar as you can change the regressions without noticing.

fredericlepied (changes requested):
Do not count in the quality bar code.

.github/workflows/pr.yml
thekad: without counting or having a ratio this is a boolean check, either it passes 100% or it doesn't. Are you OK with this outcome?

fredericlepied: It should not fail more what was failing in the main branch.

fredericlepied:
can you rebase your change as I think the diff is coming because your branch is outdated compared to main?

fredericlepied:
In fact, that is not the case. I don't understand why it is different with no code change :-(

thekad:
> In fact, that is not the case. I don't understand why it is different with no code change :-(

No idea either, I am updating from branch from the GH UI see if that makes a difference

fredericlepied:
> > In fact, that is not the case. I don't understand why it is different with no code change :-(
> 
> No idea either, I am updating from branch from the GH UI see if that makes a difference

It was fixed in main with Tatiana. I think this change is not needed anymore.

thekad:
> > > In fact, that is not the case. I don't understand why it is different with no code change :-(
> > 
> > 
> > No idea either, I am updating from branch from the GH UI see if that makes a difference
> 
> It was fixed in main with Tatiana. I think this change is not needed anymore.

I have updated the description to the PR to just disable GHA annotations for now

fredericlepied (approved):

PR #100, opened 2023-12-19
Generate more logs for tnf

ramperher:
Test-Args-Hints: -e enable_nfs_storage=true -e nfs_server=192.168.16.10 -e nfs_path=/exports/external-provisioner
Test-App-Hints: tnf_test_example
Test-App-Args-Hints: -e {"dci_components_by_query":["type:tnf_test_example"]} -e tnf_enable_hugepages=true -e do_cnf_cert=true -e {"tnf_config":[{"namespace":"test-cnf","targetpodlabels":["environment=test"],"targetoperatorlabels":["operators.coreos.com/mongodb-enterprise.test-cnf="],"target_crds":[{"nameSuffix":"crdexamples.test-network-function.com","scalable":"false"}],"exclude_connectivity_regexp":""},{"namespace":"production-cnf","targetpodlabels":["environment=production"],"targetoperatorlabels":[],"target_crds":[{"nameSuffix":"crdexamples.test-network-function.com","scalable":"false"}],"exclude_connectivity_regexp":""}]} -e {"tnf_labels":"common,telco,extended"} -e {"tnf_env_vars":{"TNF_NON_INTRUSIVE_ONLY":"false","TNF_LOG_LEVEL":"debug"}}

PR #105, opened 2023-12-21
merge queue support

fredericlepied:
Depends-On: https://softwarefactory-project.io/r/c/dci-openshift-agent/+/30479

### DIFF
--- a/.github/workflows/dci-merge.yml
+++ b/.github/workflows/dci-merge.yml
@@ -1,0 +1,27 @@
+---
+
+name: DCI
+on:
+  merge_group:
+jobs:
+
+  dci-job:
+    name: "DCI"
+    runs-on: bos2
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: DCI Job
+        run: |
+          set -ex
+          git fetch origin ${{ github.event.merge_group.base_sha }}
+          git fetch origin ${{ github.event.merge_group.head_sha }}
+          if git diff --name-only ${{ github.event.merge_group.base_sha }} ${{ github.event.merge_group.head_sha }} | grep -E 'roles/|plugins/'; then
+            ./hack/dci-merge.sh
+          else
+            echo "No code change"
+          fi
+
+...
--- a/hack/dci-merge.sh
+++ b/hack/dci-merge.sh
@@ -1,0 +1,148 @@
+#!/usr/bin/env bash
+#
+# Copyright (C) 2023 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


.github/workflows/dci-merge.yml
```diff
@@ -0,0 +1,26 @@
+---
+
+name: DCI
```
tonyskapunk, line 3: I think we should name this DCI Merge, so we can see unique workflow names in the checks:
https://github.com/redhatci/ansible-collection-redhatci-ocp

hack/dci-merge.sh
```diff
@@ -0,0 +1,148 @@
...
+GH_HEADERS=(
+    "Accept: application/vnd.github.v3+json"
+    "Authorization: token ${GITHUB_TOKEN}"
+)
+
+# Lookup the merge commits and get their PR descriptions to detect Test-Hints: strings
+COMMIT=HEAD
+VIRT=
+while true; do
+    PR=$(git log -1 "$COMMIT"|grep -oP 'Merge pull request #\K\d+')
```
tonyskapunk, line 45: What about using: 
```suggestion
    PR=$(git log --merges "$COMMIT"|grep -m1 -oP 'Merge pull request #\K\d+')
```

Since we only care about merges

hack/dci-merge.sh
```diff
@@ -0,0 +1,148 @@
...
+                APP_ARGS=$(sed -n -e "s/^\s*Test-App-Args-Hints:\s*//pi" <<< "$DESC")
+            fi
+
+            # stop at the first valid Test-Hints: string
+            break
+        fi
+    else
+        break
+    fi
+    COMMIT="${COMMIT}^"
```
tonyskapunk, line 136: If using in the top `git log --merges` we won't need this, and we reduce the odds to have a non-merge commit here

.github/workflows/dci-merge.yml
fredericlepied: no we cannot at it needs to be the same name to satisfy the required jobs.

hack/dci-merge.sh
fredericlepied: yes but I only want the PRs that are part of this merge group so I want to stop when not finding a merge commit.

hack/dci-merge.sh
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env bash
+#
+# Copyright (C) 2023 Red Hat, Inc.
```
ramperher, line 3: 2024

ramperher (commented):
I have a question. When is `dci-merge.yml` called? Only when merging the PR?
I'm saying this because I discovered [this issue](https://issues.redhat.com/browse/CILAB-1375) when testing https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/100; TLDR the hints were not applied successfully in BOS2. Now that I see the code from dci-merge.sh, I was wondering if, by launching that script for each dci-check-change job launched from PRs, this would be solved, since it looks like it's handling the hints accordingly.

fredericlepied:
recheck

tonyskapunk (approved):
LGTM

PR #110, opened 2024-01-02
Add sos_report role

tonyskapunk:
Generate SOS report from a list of OCP nodes, supports connected and disconnected OCP clusters.
A SOS report is helpful to debug issues in a node. It contains information about the underlying OS, packages, modules and logs. It is in some cases requested for certain type of bugs.

### DIFF
--- a/roles/sos_report/tasks/sos-reports.yml
+++ b/roles/sos_report/tasks/sos-reports.yml
@@ -1,0 +1,41 @@
+---
+- name: Generate SOS report for nodes
+  vars:
+    sos_report_registry: "{{ sos_report_image | regex_replace('([^/]+)/.*', '\\1') }}"
+    sos_report_image_name: "{{ sos_report_image | regex_replace('[^/]+/(.*)', '\\1') }}"
+  ansible.builtin.shell: >
+    {{ sos_report_oc_path }} debug --image={{ sos_report_image }} node/{{ node_name }} --
+    bash -c 'echo -e REGISTRY='{{ sos_report_registry }}'\\nIMAGE='{{ sos_report_image_name }}'> /host/root/.toolboxrc';
+    {{ sos_report_oc_path }} debug --image={{ sos_report_image }} node/{{ node_name }} --
+    chroot /host
+    toolbox
+    sos report --batch {{ sos_report_options }}
+  async: 600
+  poll: 0
+  register: report
+  loop: "{{ sos_report_nodes }}"
+  loop_control:
+    loop_var: node_name
+  changed_when: true
+
+- name: Check SOS report status
+  ansible.builtin.async_status:
+    jid: "{{ result.ansible_job_id }}"
+  loop: "{{ report.results }}"
+  loop_control:
+    loop_var: "result"
+  register: sos_async_results
+  until: sos_async_results.finished
+  retries: 20
+  delay: 30
+
+- name: Extract SOS report for nodes
+  vars:
+    tarball: "{{ async.stdout | regex_findall('/host/var/tmp/sosreport-.*.tar.xz') | first }}"
+  ansible.builtin.shell: >
+    {{ sos_report_oc_path }} debug --image={{ sos_report_image }} node/{{ async.result.node_name }} --
+    bash -c 'cat {{ tarball }}' > {{ sos_report_dir }}/{{ tarball | basename }}
+  loop: "{{ sos_async_results.results }}"
+  loop_control:
+    loop_var: "async"
+  changed_when: true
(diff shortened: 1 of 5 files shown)


pierreblanc (approved):
Nice feature Tony !
Lvgtm

pierreblanc (dismissed):

ramperher (commented):
This feature looks really nice @tonyskapunk , but, has it been tested?

roles/sos_report/tasks/sos-reports.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: Generate SOS report for node {{ node_name }}
+  vars:
+    sos_report_registry: "{{ sos_report_image | regex_replace('([^/]+)/.*', '\\1') }}"
+    sos_report_image_name: "{{ sos_report_image | regex_replace('[^/]+/(.*)', '\\1') }}"
+  shell: >
+    {{ sos_report_oc_path }} debug --image={{ sos_report_image }} node/{{ node_name }} --
+    bash -c 'echo -e REGISTRY='{{ sos_report_registry }}'\\nIMAGE='{{ sos_report_image_name }}'> /host/root/.toolboxrc';
+    {{ sos_report_oc_path }} debug --image={{ sos_report_image }} node/{{ node_name }} --
```
capolrik, line 9: why do you need to run the shell cmd into 2 different 'oc debug' instead of one? Is it not possible to add the .toolboxrc and run the 'toolbox' cmd at the same time?

capolrik (approved):
Very nice

roles/sos_report/tasks/sos-reports.yml
tonyskapunk: certainly can be done in one shot :stuck_out_tongue_winking_eye:  but then it's required to keep the `'` along the full command which makes harder for reading and for handling `sos_report_options` when the command is passed to the shell.

roles/sos_report/tasks/sos-reports.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: Generate SOS report for node {{ node_name }}
+  vars:
+    sos_report_registry: "{{ sos_report_image | regex_replace('([^/]+)/.*', '\\1') }}"
+    sos_report_image_name: "{{ sos_report_image | regex_replace('[^/]+/(.*)', '\\1') }}"
+  shell: >
```
thekad, line 6: FQCN shell module

roles/sos_report/tasks/sos-reports.yml
```diff
@@ -0,0 +1,41 @@
...
+  async: 600
+  poll: 0
+  register: report
+  loop: "{{ sos_report_nodes }}"
+  loop_control:
+    loop_var: node_name
+  changed_when: true
+
+- name: Check SOS report status
+  async_status:
```
thekad, line 22: FQCN async_status module

roles/sos_report/tasks/sos-reports.yml
```diff
@@ -0,0 +1,41 @@
...
+    loop_var: "result"
+  register: sos_async_results
+  until: sos_async_results.finished
+  retries: 20
+  delay: 30
+
+- name: Extract SOS report for node {{ async.result.node_name }}
+  vars:
+    tarball: "{{ async.stdout | regex_findall('/host/var/tmp/sosreport-.*.tar.xz') | first }}"
+  shell: >
```
thekad, line 35: FQCN shell module

roles/sos_report/tasks/sos-reports.yml
```diff
@@ -0,0 +1,41 @@
...
+  loop_control:
+    loop_var: "result"
+  register: sos_async_results
+  until: sos_async_results.finished
+  retries: 20
+  delay: 30
+
+- name: Extract SOS report for node {{ async.result.node_name }}
+  vars:
+    tarball: "{{ async.stdout | regex_findall('/host/var/tmp/sosreport-.*.tar.xz') | first }}"
```
thekad, line 34: if you're finding the first tarball found, you have to clean the tarballs at some point either before you run the sos report (preferred) or after you've processed the tarball

thekad (changes requested):

roles/sos_report/tasks/sos-reports.yml
tonyskapunk: > if you're grabbing the first tarball found, 

This `first` is to extract the unique element out of the list, see this is a loop, so it does on every iteration.

> you have to clean the tarballs at some point either before you run the sos report (preferred) or after you've processed the tarball

Not necessarily, the tarballs have a timestamp in the name that makes them unique.

thekad: what's the current behavior if you run sos_report through this module multiple times? is it safe? how big are ,SOS report files?

tonyskapunk: > what's the current behavior if you run sos_report through this module multiple times? is it safe? how big are SOS report files?

A tarball is created for each node defined, each tarball is unique to the run. 

Don't understand the relevance of the size of the tarballs as that is not  under the control of this role.

pierreblanc: It is maybe impact the storage of the DCI infrastructure :D

tonyskapunk: > It is maybe impact the storage of the DCI infrastructure :D

Not for this role, though, since the role does not upload the sos reports to DCI, that would be a concern for the agent that picks these tarballs into DCI :smile:

thekad: I'm not trying to be difficult, I was just directing attention to the possible failure scenarios on this role, if there are no failure scenarios then cool by me.

thekad (approved):

PR #112, opened 2024-01-03
Clear cache for ansible-lint step

tkrishtop:


tkrishtop:
Closing this PR as the issue is not in the old "Compute ansible-lint score" that is doing fine but in a new "Ansible-lint Check" introduced today.

PR #125, opened 2024-01-08
DCI Job: send a QUEUED message to provide more feedback

fredericlepied:
- sync the code check between maybe-dci-check-change and dci-merge.sh

### DIFF
--- a/hack/maybe-dci-check-change
+++ b/hack/maybe-dci-check-change
@@ -26,20 +26,22 @@ URL="$1"
 GITHUB_REPO_TOKEN=$($BASEDIR/get-config-entry "https://github.com/$proj" github_token "$GITHUB_TOKEN")
 
 if [ -n "$GITHUB_REPO_TOKEN" ]; then
-   AUTH=(-H "Authorization: token $GITHUB_REPO_TOKEN")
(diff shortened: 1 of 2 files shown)


hack/maybe-dci-check-change
```diff
@@ -26,20 +26,22 @@ URL="$1"
...
-   AUTH=(-H "Authorization: token $GITHUB_REPO_TOKEN")
+   HEADERS=(-H "Accept: application/vnd.github+json" -H "Authorization: token $GITHUB_REPO_TOKEN")
 else
-   AUTH=()
+   HEADERS=(-H "Accept: application/vnd.github+json")
 fi
 
 pr=$(sed -e 's@.*/pull/'@@ <<< $URL)
 proj=$(sed -e 's@https://github.com/\(.*\)/pull/.*@\1@i' <<< $URL)
+json=$(curl -s -H "Accept: application/vnd.github.v3+json" "${HEADERS[@]}" https://api.github.com/repos/$proj/pulls/$pr)
```
thekad, line 36: does this call need both Accept headers? or can you just use vnd.github.v3+json e.g. move it to the default headers up?

fredericlepied: good catch. it is already in the HEADERS var.

PR #129, opened 2024-01-08
Move default ipv6 file to default dir

tonyskapunk:
For older versions, 4.8 and below, the use of dualstack requires a FeatureGate manifest that is not in the default location, causing a failure in the install.

Here an example: https://www.distributed-ci.io/jobs/54348a4a-07c0-4dcc-8fa7-4c2b4e9eea55/jobStates?sort=date&task=dbe37c91-a949-4adc-928d-a737572403a7

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/7a5fce3bfe734127465b9ed12f9ef9ca27e60b66/roles/installer/tasks/50_extramanifests.yml#L4


thekad:
@tonyskapunk I can see it succeeded with 4.14, but you want to test this with <=4.8 to make sure it works, yes?

tonyskapunk:
> @tonyskapunk I can see it succeeded with 4.14, but you want to test this with <=4.8 to make sure it works, yes?

Correct, I'm requesting a partner to test this for us

tonyskapunk:
Talked to partner and they confirmed not needing this for 4.7. No need to get this fixed, closing

PR #131, opened 2024-01-09
Honor warnings from ansible-lint

tonyskapunk:
CI for ansible-lint PRs is ignoring the warnings from ansible-lint. Creating another section to highlight the warnings without failing the check.
Including a new warning var-naming[no-role-prefix] while we decide to use this standard or keep the current one.
Through this mechanism we can remove the warning list once we want to start truly enforcing them.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -10,7 +10,8 @@ exclude_paths:
   - plugins
 
 warn_list:
-  - yaml[line-length]
   - ignore-errors
   - no-changed-when
+  - var-naming[no-role-prefix]
+  - yaml[line-length]
 ...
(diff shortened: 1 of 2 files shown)


ramperher (approved):
LGTM

tkrishtop (approved):
We discussed the change with @tonyskapunk, and I believe we'll get some benefits once this PR is merged

thekad:
I wish there was a way to configure the format for the prefix because I do think it is important to have them be unique since ansible only has (effectively) a global scope. I like where ansible/ansible-lint#3451 is moving but not possible at the moment.

thekad (approved):

PR #135, opened 2024-01-13
Remove tested image form root account

betoredhat:
The images tested using oscap-podman aren't deleted, which can fill up the filesystem of the host used for the testing.

### DIFF
--- a/roles/preflight/tasks/test_run_health_check.yml
+++ b/roles/preflight/tasks/test_run_health_check.yml
@@ -54,6 +54,14 @@
     chdir: "{{ preflight_container_artifacts.path }}"
   become: true
 
+- name: Remove the tested image
+  shell:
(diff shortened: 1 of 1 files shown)


roles/preflight/tasks/test_run_health_check.yml
```diff
@@ -54,6 +54,14 @@
     chdir: "{{ preflight_container_artifacts.path }}"
   become: true
 
+- name: Remove the tested image
+  shell:
+    cmd: >
+      podman rmi
+      --force
+      {{ current_operator_image }}
+  become: true
```
tonyskapunk, line 63: It would be good to know why it is needed to run as root, let's create a card for that

tonyskapunk (approved):
LGTM

PR #140, opened 2024-01-19
Replace repo_root_path with dci_cluster_configs_dir pinned in DOA

tkrishtop:
Test-Hints: assisted-abi

Depends-On: https://github.com/dci-labs/inventories/pull/195

This is an attempt to fix the bug described [here](https://issues.redhat.com/browse/CILAB-1422)


tkrishtop:
retest

tkrishtop:
recheck

nocturnalastro:
We have to think of backward compatibility with pre-existing crucible inventories. What about defaulting `repo_root_path` to `dci_cluster_configs_dir`. then in the inventory validation we can check if either are defined. 
That way dci can use `dci_cluster_configs_dir` and we don't break anyone's inventories.

tkrishtop:
> We have to think of backward compatibility with pre-existing crucible inventories. What about defaulting `repo_root_path` to `dci_cluster_configs_dir`. then in the inventory validation we can check if either are defined.

Thank you @nocturnalastro, I see the point. Let's do as you propose, let me push the changes.

tkrishtop:
> > We have to think of backward compatibility with pre-existing crucible inventories. What about defaulting `repo_root_path` to `dci_cluster_configs_dir`. then in the inventory validation we can check if either are defined.
> 
> Thank you @nocturnalastro, I see the point. Let's do as you propose, let me push the changes.

Hi @nocturnalastro, I made the changes you requested and tested on assisted-abi installation: https://www.distributed-ci.io/jobs/ff803bb8-a088-464f-b40e-872981a2b26d/jobStates?sort=date 

Could you please have a look one more time?

tonyskapunk:
I'm not sure the roles should use a dci variable coming from the dci openshift agent. Would it be possible to make `repo_root_path: dci_cluster_configs_dir` as a var when called from the agent instead?

ramperher:
> I'm not sure the roles should use a dci variable coming from the dci openshift agent. Would it be possible to make `repo_root_path: dci_cluster_configs_dir` as a var when called from the agent instead?

Yes, I agree with @tonyskapunk , maybe it's better to set up that variable when calling these roles from the agent.

tkrishtop:
> I'm not sure the roles should use a dci variable coming from the dci openshift agent. Would it be possible to make `repo_root_path: dci_cluster_configs_dir` as a var when called from the agent instead?

Hi @tonyskapunk thanks for the idea, makes sense. Let me try it out.

thekad:
agreed with @tonyskapunk these roles are not DCI specific so they shouldn't rely on DCI variables. There's two ways you can do this:
* do what Tony said and alter it on the calling playbook (easier, cleaner)
* Add an extra variable `default_repo_root_path` and alter that one (somewhat dirtier, but more flexible maybe)
I'd go with Tony's way, and only adjust these roles in a non-DCI way.

tkrishtop:
> agreed with @tonyskapunk these roles are not DCI specific so they shouldn't rely on DCI variables. There's two ways you can do this:
> 
> * Do what Tony said and alter it on the calling playbook (easier, cleaner)
> * Add an extra variable `default_repo_root_path` and alter that one (somewhat dirtier, but more flexible maybe)
> 
> I'd go with Tony's way, and only adjust these roles in a non-DCI way.

agree, I'm testing with the current way just to ensure that all the dependencies are taken into account. Once it's green, I'll move the repo_root_path definition in the agent during the roles' invocation.

tkrishtop:
check dallas ocp-4.14-vanilla-abi

PR #141, opened 2024-01-19
Make operator-sdk job work again

tkrishtop:
TestDallasWorkload: operator-sdk

### DIFF
--- a/roles/operator_sdk/tasks/build_scorecard_image.yml
+++ b/roles/operator_sdk/tasks/build_scorecard_image.yml
@@ -26,7 +26,7 @@
   when: job_id is defined
 
 - name: Build Operator-SDK image
-  ansible.builtin.command:
+  ansible.builtin.shell:
     cmd: >
       podman buildx build
       -t {{ operator_sdk_img }}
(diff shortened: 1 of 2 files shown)


tkrishtop:
check workload operator-sdk

tkrishtop:
@thekad Jorge, why did we change from `shell` to `command` everywhere? We are facing a situation where the parsing of multiline commands does not work as intended because of `&&`. The straightforward solution is to revert this task from command back to shell. Would that be acceptable, or should we avoid using shell for some other reasons?

tkrishtop:
check workload operator-sdk

tkrishtop:
check workload operator-sdk

tkrishtop:
recheck

tkrishtop:
Tested green: https://www.distributed-ci.io/jobs/1f6979f2-cdf8-4486-a25d-929aa2b0455d/tests 
Would appreciate a review

tkrishtop:
@fredericlepied could you please approuve? that's a `command -> shell` fix for the operatot-sdk job that we have discussed

PR #145, opened 2024-01-22
Fix release name

betoredhat:
Using the local role variable that actually represents the release version that is been mirrored.


tonyskapunk (approved):
LGTM!

thekad (approved):
ah yes, this makes sense! :smile:

PR #147, opened 2024-01-23
Delete TNF build artifacts

betoredhat:
- Delete dangling containers and images used to build TNF

### DIFF
--- a/roles/cnf_cert/tasks/pre-run.yml
+++ b/roles/cnf_cert/tasks/pre-run.yml
@@ -111,6 +111,7 @@
       ansible.builtin.shell: |
         set -x
         podman build -t {{ tnf_image }} \
+        --force-rm \
         --build-arg TNF_VERSION={{ tnf_version_image }} \
         --build-arg TNF_SRC_URL=https://github.com/test-network-function/cnf-certification-test \
         --build-arg OPENSHIFT_VERSION={{ ocp_version_full }} .
--- a/roles/cnf_cert/tasks/teardown.yml
+++ b/roles/cnf_cert/tasks/teardown.yml
@@ -1,5 +1,17 @@
 ---
 # Deleting resources created by the CNF Cert Suite to support the execution
+- name: Prune Buildah images and containers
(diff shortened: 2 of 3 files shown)


roles/cnf_cert/tasks/teardown.yml
```diff
@@ -1,5 +1,17 @@
 ---
 # Deleting resources created by the CNF Cert Suite to support the execution
+- name: Prune Buildah images and containers
```
ramperher, line 3: Note the teardown tasks are not executed in the DCI teardown stage, they're executed just after cnf_cert/tests.yml. If the job fails in the tests.yml, then these tasks will not be executed. That's why the "Remove local tnf image" task was just after the execution of tnf.

I think there are two options here:

- Put again the image removal tasks after tnf execution in tests.yml
- Think on a more sophisticated way of implementing this, e.g. call teardown.yml in DCI teardown stage and include the logic in main.yml to select what to do (maybe this is the most interesting one)

betoredhat: Thanks @ramperher. I want to hold on a bit on running the buildah cleanup as it is going to remove the dangling artifacts even those that were not created by the CNF image build. May be better to run it from our cleanup scripts.  Let's see if the  `--force-rm` during the image build helps a bit. 

Cheers

roles/cnf_cert/tasks/pre-run.yml
```diff
@@ -111,6 +111,7 @@
       ansible.builtin.shell: |
         set -x
         podman build -t {{ tnf_image }} \
+        --force-rm \
```
thekad, line 114: are you sure this does what you think it does? this flag defaults to true

thekad:
```
   --force-rm
       Always remove intermediate containers after a build, even if the  build  fails  (de‐
       fault true).
```

betoredhat: well, I didn't realize about the default. Will close this. Thanks

PR #148, opened 2024-01-24
verify_tests: restored file name in fail task

nsilla:
The verify_tests task "Fail if not all test results are as expected" used to include the name of the affected file but it got removed in [this commit](https://github.com/redhatci/ansible-collection-redhatci-ocp/commit/848f809caf5edfccfd543dd82d74196b99560908#diff-d637e677ca97f8c3fc558dcf18c514027e2176b0c3a5808d8ca2c6eaecad6eae).

We need the name of the file in order to classify properly a failed job with the appropriate jira tag.

### DIFF
--- a/roles/verify_tests/tasks/parse_junit_file.yml
+++ b/roles/verify_tests/tasks/parse_junit_file.yml
@@ -34,7 +34,7 @@
       ansible.builtin.set_fact:
         actual_results: "{{ item | redhatci.ocp.junit2dict }}"
 
-    - name: Fail if not all test results are as expected
+    - name: "Fail if not all test results are as expected for {{ item | basename }}"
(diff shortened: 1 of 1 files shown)


thekad (approved):
This is the rule I followed when I removed it https://ansible.readthedocs.io/projects/lint/rules/name/#name
```
name[template] - Jinja templates should only be at the end of 'name'. This helps with the
    identification of tasks inside the source code when they fail. The use of templating
    inside name keys is discouraged as there are multiple cases where the rendering of the
    name template is not possible.
```
It may not be failing because now it is at the end of the name, even though it is discouraged, or maybe I just removed all templates in names, can't remember. As long as we follow the rules I am fine.

PR #151, opened 2024-01-30
Allow port configuration for assisted installer setup

ramperher:
Test-Hints: assisted

In this [issue](https://issues.redhat.com/browse/CILAB-1338), while testing Assisted on-prem in our labs, it was failing because the port used by `assisted_service_gui` was colliding with another service we run.

This change allows to provide the ports for `assisted_service_gui` and `assisted_service_image_service` externally with new variables.

### DIFF
--- a/roles/setup_assisted_installer/defaults/main.yml
+++ b/roles/setup_assisted_installer/defaults/main.yml
@@ -5,6 +5,9 @@ assisted_service_image: "{{ assisted_service_image_repo_url }}/assisted-service@
 assisted_service_gui_image: "{{ assisted_service_image_repo_url }}/assisted-installer-ui@{{ image_hashes.gui }}"
 assisted_service_image_service_image: "{{ assisted_service_image_repo_url }}/assisted-image-service@{{ image_hashes.image_service }}"
 
+assisted_service_gui_port: 8080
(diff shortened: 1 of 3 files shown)


tkrishtop (commented):
Great dev, thanks @ramperher ! It would be nice to rename the PR following the best practices to keep it under 50 chars, start with a capital letter, do not use underscores in the description, etc. Something like "Allow port configuration for assisted installer setup"

ramperher:
> Great dev, thanks @ramperher ! It would be nice to rename the PR following the best practices to keep it under 50 chars, start with a capital letter, do not use underscores in the description, etc. Something like "Allow port configuration for assisted installer setup"

Sure, done!

tkrishtop (dismissed):
LGTM, thank you @ramperher! Maybe you could also do a follow-up PR to open these ports in the firewall

nocturnalastro:
@ramperher Defiantly worth doing. I think you will need to change this as well as
https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/setup_assisted_installer/templates/configmap.yml.j2#L8

nocturnalastro (approved):

PR #156, opened 2024-02-01
Add replication support

betoredhat:


roles/ocp_logging/README.md
```diff
@@ -27,6 +27,7 @@ This role only works on OCP 4.10 and newer and with a S3 compatible storage back
 | ol_endpoint                            | undefined                     | Yes         | Object Storage endpoint                       |
 | ol_region                              | undefined                     | Yes         | Object Storage region                         |
 | ol_loki_size                           | undefined                     | Yes         | Loki Deployment Size. See [Sizing](https://docs.openshift.com/container-platform/4.13/logging/cluster-logging-loki.html#deployment-sizing_cluster-logging-loki) for more details |
+| ol_replication_factor                  | undefined                     | Yes         | Replication factor                            |
```
tonyskapunk, line 30: Would it make sense to set a default? It avoids breaking previous uses of this role by not making it required, thoughts?

betoredhat: Make sense. I was testing if this setting would allow us to control the number of pods for each Loki component but it was not the case. So closing this. 

Thanks

PR #158, opened 2024-02-06
Ignore error during image removal

betoredhat:
- Should be safe to ignore errors. Sometimes this may fail if a contaner referencing the image is already removed.


manurodriguez (approved):
LGTM

thekad (approved):

tkrishtop (approved):
thank you @betoredhat

PR #160, opened 2024-02-08
Introduce the sideload-kernel role

lack:
Test-Hints: sno
Test-Args-Hints: -e dci_sno_sideload_kernel_uri=http://server01.partnerci.bos2.lab:8080/pub/rt/kernel-rt-core-5.14.0-421.el9.x86_64.rpm

### DIFF
--- a/roles/sideload_kernel/README.md
+++ b/roles/sideload_kernel/README.md
@@ -1,0 +1,73 @@
+# Sideload-kernel utility role
+
+This role will side-load a given realtime kernel onto an OpenShift SNO instance for
+development and testing purposes.  It will replace the stock standard kernel, or
+the stock realtime kernel if that has been enabled via MachineConfig or
+NodeTuningOperator CRs.
+
+Note: Only supported for side-loading custom realtime kernels, and only tested
+on SNO installations.
+
+## Variables
+
+| Variable                    | Default     | Required | Description           |
+| --------------------------- | ----------- | -------- | --------------------- |
+| sideload_kernel_uri         | undefined   | Yes      | The full URI to the kernel-rt-core rpm package to be installed, or the special value `"reset"` to reset back to the original kernel. |
+| sideload_kernel_namespace   | `"default"` | No       | The namespace where the job and pods run that conduct the change. Must be privileged. |
+| sideload_kernel_force       | `false`     | No       | Forces re-creation of the kubernetes job even if no changes occurred. |
+| sideload_kernel_job_timeout | `15`        | No       | The amount of time to wait for the sideload operation to complete (in minutes) |
+| sideload_kernel_base_image  | `"ubi9"`    | No       | The image used to run the script on the cluster. |
+| k8s_auth                    | `{}`        | No       | See the "Authentication" section below. |
+
+## Requirements
+
+- python3-kubernetes (or [kubernetes python library](https://pypi.org/project/kubernetes/))
+
+## Authentication
+
+The steps taken by this role require proper kubernetes authentication be set up
+on the ansible host (or localhost) for the cluster in question.  This may be
+done in 3 ways:
+
+- If the ansible host has a valid kubeconfig in ~/.kube/config, this will be
+  used by default.
+- You can set the appropriate environment variables via the `k8s_auth` role
+  variable. These will be named K8S_AUTH_* and are outlined in [kubernetes.core.k8s](https://galaxy.ansible.com/ui/repo/published/kubernetes/core/content/module/k8s/)
+
+## Usage example
+
+- Side-load a kernel, assuming ~/.kube/config is authenticated to
+  the proper kubernetes cluster:
+
+```yaml
+- name: Sideload a kernel
+  ansible.builtin.include_role:
+    name: redhatci.ocp.sideload_kernel
+  vars:
+    sideload_kernel_uri: "https://example.com/packages/kernel/5.14.0/417.el9/x86_64/kernel-rt-core-5.14.0-417.el9.x86_64.rpm"
+```
+
+- Side-load a kernel, specifying the path to an alternative
+  kubeconfig file:
+
+```yaml
+- name: Sideload a kernel
+  ansible.builtin.include_role:
+    name: redhatci.ocp.sideload_kernel
+  vars:
+    sideload_kernel_uri: "https://mymirror.local/kernels/kernel-rt-core-5.14.0-417.el9.x86_64.rpm"
+    k8s_auth:
+      K8S_AUTH_KUBECONFIG: /var/lib/clusterauth/kubeconfig
+```
+
+- Reset back to the original kernel
+
+```yaml
+- name: Reset to the custom standard kernel
+  ansible.builtin.include_role:
+    name: redhatci.ocp.sideload_kernel
+  vars:
+    sideload_kernel_uri: "reset"
+    k8s_auth:
+      K8S_AUTH_KUBECONFIG: /var/lib/clusterauth/kubeconfig
+```
--- a/roles/sideload_kernel/files/flip_kernel
+++ b/roles/sideload_kernel/files/flip_kernel
@@ -1,0 +1,184 @@
+#!/bin/bash -eu
+#
+# Note: This script is designed to run inside a privileged OCP container that
+# has been chroot'd into the host filesystem, as it deals directly  with
+# rpm-ostree.
+#
+# Upon execution, one of 3 results will happen
+# - It will return 0 if the desired kernel is installed and active
+# - It will return 1 if there is any error
+# - It will reboot the system if the requested kernel is not yet active
+#
+# Run this script with '-h' for usage instructions.
+#
+
+NONRT=(kernel{-core,-modules,-modules-core,-modules-extra})
+RT=(kernel-rt{-core,-modules,-modules-core,-modules-extra})
+PREFETCH=/tmp
+
+replace_standard_kernel() {
+    local -a install
+    for pkg in "$@"; do
+        install+=(--install "${pkg}")
+    done
+    rpm-ostree override remove kernel "${NONRT[@]}" "${install[@]}"
+}
+
+reinstall_kernel_packages() {
+    local -a install
+    for pkg in "$@"; do
+        install+=(--install "${pkg}")
+    done
+    rpm-ostree uninstall --all "${install[@]}"
+}
+
+replaced_base_kernel() {
+    rpm-ostree status --json | jq -r '.deployments[0]."requested-base-removals"[]' | grep -q kernel-core
+}
+
+is_rpmostree_staged() {
+    [[ $(rpm-ostree status --json | jq '.deployments[0].staged') == "true" ]]
+}
+
+is_pkg_installed() {
+    local basename
+    basename=$(basename "$1" .rpm)
+    rpm -q "$basename" >/dev/null
+}
+
+reboot_if_staged() {
+    rpm-ostree status
+    if is_rpmostree_staged; then
+        # TODO: There is a race with MCD which may run 'rpm-ostree cleanup -p'
+        # on us and blow away our staged changes before they take effect.
+        # Attempt to make the window smaller by rebooting immediately
+        echo "Rebooting!"
+        systemctl reboot
+    else
+        echo "No changes staged!"
+        return 1
+    fi
+}
+
+inject_kernel_rpms() {
+    if is_pkg_installed "$1"; then
+        echo "$(basename "$1") is already installed; nothing to do!"
+        uname -r
+        return 0
+    fi
+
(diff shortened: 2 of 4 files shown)


roles/sideload_kernel/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Variables
+
+| Variable                     | Default                               | Required | Description                                                                                    |
+| ---------------------------- | ------------------------------------- | -------- | ---------------------------------------------------------------------------------------------- |
+| sideload_kernel_uri          | undefined                             | Yes      | Full URI to the kernel-rt-core...rpm package to be installed.                                  |
+| sideload_kernel_namespace    | default                               | No       | The namespace where the job and pods run that conduct the change. Must be privileged.          |
+
+## Requirements
+
+- [kubernetes python library](https://pypi.org/project/kubernetes/)
```
thekad, line 20: should list RPM package names as well

roles/sideload_kernel/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Note: Only supported for side-loading custom realtime kernels, and only tested
+on SNO installations.
+
+## Variables
+
+| Variable                     | Default                               | Required | Description                                                                                    |
+| ---------------------------- | ------------------------------------- | -------- | ---------------------------------------------------------------------------------------------- |
+| sideload_kernel_uri          | undefined                             | Yes      | Full URI to the kernel-rt-core...rpm package to be installed.                                  |
+| sideload_kernel_namespace    | default                               | No       | The namespace where the job and pods run that conduct the change. Must be privileged.          |
```
thekad, line 16: There's another variable `sideload_kernel_force`  in the defaults file

roles/sideload_kernel/files/flip_kernel
```diff
@@ -0,0 +1,166 @@
...
+NONRT=(kernel{-core,-modules,-modules-core,-modules-extra})
+RT=(kernel-rt{-core,-modules,-modules-core,-modules-extra})
+PREFETCH=/tmp
+
+replace_standard_kernel() {
+    local -a install
+    for pkg in "$@"; do
+        install+=(--install "${pkg}")
+    done
+    sudo rpm-ostree override remove kernel "${NONRT[@]}" "${install[@]}"
```
thekad, line 12: You can get rid of the assorted `sudo` calls in this script if you force it to run as root e.g. adding `if $( id -u ) -neq 0; then exit 1; fi` or similar at the very top of the script. I believe this script should be run as root anyway.

roles/sideload_kernel/files/flip_kernel
```diff
@@ -0,0 +1,166 @@
...
+reinstall_kernel_packages() {
+    local -a install
+    for pkg in "$@"; do
+        install+=(--install "${pkg}")
+    done
+    sudo rpm-ostree uninstall --all "${install[@]}"
+}
+
+replaced_base_kernel() {
+    rpm-ostree status --json | jq -r '.deployments[0]."requested-base-removals"[]' | grep -q kernel-core
```
thekad, line 24: if the script requires `jq` then add it to the requirements

roles/sideload_kernel/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
+---
+- name: Validate required variables
+  ansible.builtin.assert:
+    that:
+      - sideload_kernel_uri | string
+
+- name: Install python dependencies
```
thekad, line 7: don't install python packages via pip, install the required RPM instead, or, if there is no RPM, at least install this in a python virtual environment

thekad: This would also be a good place to ensure `jq` is installed

roles/sideload_kernel/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
...
+        name: flip-kernel
+        namespace: "{{ sideload_kernel_namespace }}"
+      spec:
+        backoffLimit: 3
+        # TODO: Add a selector for multi-node cluster cases?
+        template:
+          spec:
+            containers:
+              - command:
+                  - /bin/sh
```
thekad, line 63: did you test this with `/bin/sh`? I think I saw a few BASH-isms in the script

thekad: oh I see, it does call `/tmp/flip_kernel` at end so it ends up running through whatever shebang it has

thekad (changes requested):
Good stuff! there's a few things I noticed

roles/sideload_kernel/files/flip_kernel
lack: The script runs in the container (or more accurately the chroot to the host running the container), so it's not required to be present on the ansible host.

roles/sideload_kernel/tasks/main.yml
lack: I was curious about this, and just generally how we are supposed to handle dependencies of these ansible roles...

I see other roles that also use community.kubernetes.k8s without any explicit installation of the required python libaries.  Is there a better way to do this?

roles/sideload_kernel/tasks/main.yml
lack: But maybe using bash here too will remove any confusion :)

roles/sideload_kernel/files/flip_kernel
lack: The script was originally designed to run by hand, and envolved out of something that was information as non-root and effective as root.  But it's true that since the meat of the script needs root, the whole script should. I'll do that.

roles/sideload_kernel/files/flip_kernel
lack: I'll add a comment to that effect at the top of the script so it's more clear.

roles/sideload_kernel/tasks/main.yml
lack: I tried a version that uses ansible.builtin.package to install the RPM, with a fallback to pip... but that requires becoming root on the host, which seems a bit heavy considering nothing else in this role requires it...

Maybe I should remove this and rely on people reading the dependencies section of the README?

roles/sideload_kernel/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+      ansible.builtin.package:
+        name: python3-kubernetes
+        state: present
+  rescue:
+    - name: Fallback to pip installation
+      ansible.builtin.pip:
+        name: kubernetes
+
+- name: Sideload-kernel job namespace
+  kubernetes.core.k8s:
```
tonyskapunk, line 20: Please switch to community.kubernetes, we don't have support for kubernetse.core.k8s yet as newer versions only support newer versions of ansible and we support 2.9
```suggestion
  community.kubernetes.k8s:
```

roles/sideload_kernel/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Side-load a kernel directly from brew, assuming ~/.kube/config is
+  authenticated to the proper kubernetes cluster:
+
+```yaml
+- name: Sideload a specific brew kernel
+  ansible.builtin.include_role:
+    name: redhatci.ocp.sideload_kernel
+  vars:
+    sideload_kernel_uri: "https://download.eng.bos.redhat.com/brewroot/vol/rhel-9/packages/kernel/5.14.0/417.el9/x86_64/kernel-rt-core-5.14.0-417.el9.x86_64.rpm"
```
tonyskapunk, line 45: May be worth not using internal domains/paths and redact with something like example.com/path/to/kernel-rt-core-5.14.0-417.el9.x86_64.rpm or simliar

roles/sideload_kernel/files/flip_kernel
```diff
@@ -0,0 +1,184 @@
+#!/bin/bash
+#
+# Note: This script is designed to run inside a priviliged OCP conatiner that
```
tonyskapunk, line 3:
```suggestion
# Note: This script is designed to run inside a privileged OCP container that
```

roles/sideload_kernel/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+                args:
+                  - "-c"
+                  - "install /script/flip_kernel /host/tmp && chroot /host /tmp/flip_kernel ${KERNEL_URI}"
+                env:
+                  - name: KERNEL_URI
+                    valueFrom:
+                      configMapKeyRef:
+                        name: sideload-kernel
+                        key: KERNEL_URI
+                image: ubi9
```
tonyskapunk, line 79: Can you make this one a variable with ubi9 being the default? this way we can make it work in a disconnected environment by passing the Fully Qualified Repository Name of the local registry

tonyskapunk (commented):
Hi @lack thanks for your contribution!

roles/sideload_kernel/README.md
```diff
@@ -0,0 +1,73 @@
...
+## Variables
+
+| Variable                    | Default     | Required | Description           |
+| --------------------------- | ----------- | -------- | --------------------- |
+| sideload_kernel_uri         | undefined   | Yes      | The full URI to the kernel-rt-core rpm package to be installed, or the special value `"reset"` to reset back to the original kernel. |
+| sideload_kernel_namespace   | `"default"` | No       | The namespace where the job and pods run that conduct the change. Must be privileged. |
+| sideload_kernel_force       | `false`     | No       | Forces re-creation of the kubernetes job even if no changes occurred. |
+| sideload_kernel_job_timeout | 15          | No       | The amount of time to wait for the sideload operation to complete (in minutes) |
+| sideload_kernel_base_image  | `"sideload_kernel_job_timeout9"`    | No       | The image used to run the script on the cluster. |
+| k8s_auth                    | undefined   | No       | See the "Authentication" section below. |
```
ramperher, line 20: this variable has a default variable in fact, which is `{}`

roles/sideload_kernel/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+              - name: host
+                hostPath:
+                  path: /
+                  type: Directory
+              - name: script
+                configMap:
+                  name: sideload-kernel
+  environment: "{{ k8s_auth }}"
+
+- ansible.builtin.debug: # noqa name[missing]
```
ramperher, line 104: Missing task name?

ramperher (commented):
Some comments from my side

tonyskapunk (commented):
The [failures in the sanity checks](https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/7855555067?pr=160#summary-21438751408) are discussed in https://github.com/ansible/ansible/issues/78346

The use case in this role is valid, I suggest to bring it up in that issue as another use case

roles/sideload_kernel/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Variables
+
+| Variable                    | Default     | Required | Description           |
+| --------------------------- | ----------- | -------- | --------------------- |
+| sideload_kernel_uri         | undefined   | Yes      | The full URI to the kernel-rt-core rpm package to be installed, or the special value `"reset"` to reset back to the original kernel. |
+| sideload_kernel_namespace   | `"default"` | No       | The namespace where the job and pods run that conduct the change. Must be privileged. |
+| sideload_kernel_force       | `false`     | No       | Forces re-creation of the kubernetes job even if no changes occurred. |
+| sideload_kernel_job_timeout | 15          | No       | The amount of time to wait for the sideload operation to complete (in minutes) |
+| sideload_kernel_base_image  | `"sideload_kernel_job_timeout9"`    | No       | The image used to run the script on the cluster. |
```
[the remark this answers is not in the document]
lack: That's what I get for writing code on my phone 🤦

roles/sideload_kernel/tasks/main.yml
lack: Intentional.  I've never seen the point in having names for debug tasks, since they just print a thing that's more useful than any static name I could come up with for the task.

That said, I wouldn't mind naming it, if that's an important convention.

lack:
I changed the bash shebang to `#!/bin/bash -eu` which should hopefully pass the sanity check.

roles/sideload_kernel/tasks/main.yml
thekad: you can have a validate section at the top of main.yml (or include a validate.yml or whatever) to make sure you've got all the things you need to run the sideload script and bail early if not with a useful message.

fredericlepied:
@thekad are we good with this PR? It is working in DCI now.

fredericlepied (approved):

tonyskapunk (approved):
LGTM, thanks for all the work in here!

PR #161, opened 2024-02-09
Evaluate based on requested VFs in the PF range

tonyskapunk:
Instead of evaluating by the total amount of VFs (num_vfs), compare against the requested PF range per node policy. Default to num_vfs when no range has been defined.


tonyskapunk:
check dallas ocp-4.14-vanilla example-cnf-blue

ramperher (dismissed):
LGTM

manurodriguez (approved):
LGTM

PR #164, opened 2024-02-12
Replace podman_container with podman run to fix pids_limit issue

tkrishtop:


tkrishtop:

manurodriguez (commented):
Looks good to me, but just for curiosity what is the issue with `pids_limit` in the podman collection?

tkrishtop:
> Looks good to me, but just for curiosity what is the issue with `pids_limit` in the podman collection?

It looks like the multiple downloads exceed the default limit for the number of sub-processes that the container process could spawn. In our case, we think we set unlimited by using `-1` but I believe that we did not because [ `pids_limit=-1` means "unlimited" for docker and "default" for before-2021 podman](https://github.com/containers/podman/pull/11794). 

We also have a problem setting this value properly when using the podman_container module
```
# code
containers.podman.podman_container:
-- snip --
pids_limit: -1
-- snip --

# error
[WARNING]: The value -1 (type int) in a string field was converted to '-1' (type string). 
If this does not look like what you expect, quote the entire value to ensure it does not change. 
```

So the fix I'm trying to implement is
- Replace podman_container module with cmd/podman run
- Set pids_limit to 0

[Here you have more job links and details](https://issues.redhat.com/browse/CILAB-1363).

tkrishtop:
[The build is working](https://www.distributed-ci.io/jobs/dce6ec67-85ba-43f2-b9ad-1c88d377028d/jobStates?sort=date&task=0057137f-a45f-4479-b927-2a18d07182b2), ready for a review.

manurodriguez (approved):
Thanks for the details, I was wondering if  `pids_limit: "{{ -1 | int }}"` could work, that might get rid of the warning about the type string, but I do not know if the podman container collection will actually use it.

tonyskapunk:
> We also have a problem setting this value properly when using the podman_container module
> 
> ```
> # code
> containers.podman.podman_container:
> -- snip --
> pids_limit: -1
> -- snip --
> 
> # error
> [WARNING]: The value -1 (type int) in a string field was converted to '-1' (type string). 
> If this does not look like what you expect, quote the entire value to ensure it does not change. 
> ```
> 

This is not an error, is a warning, it's telling that the value passed is an int and requires a string, but it casts a string for you. Doc [1]states it needs to be a string, to avoid the warning simply do

```yaml
     pids_limit: "-1"
```


[1] https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-pids_limit

roles/opcap_tool/tasks/build.yml
```diff
@@ -9,20 +9,15 @@
...
-    volume:
-      - "{{ opcap_dir.path }}/opcap/:/opcap:Z"
+  # Launching a container to build the opcap binary.
+  # The directory where the repository has been cloned is bound as a volume.
+  # The 'Z' option is for SELinux compatibility.
+  command:
+    cmd: >
+      podman run --name go-builder --rm -it
+      --log-level info --pids-limit=0
+      -v {{ opcap_dir.path }}/opcap/:/opcap:Z
```
tonyskapunk, line 19: For the selinux part, since it is a temp file, the right context could be set before running the container and avoid switching to podman

We have some other roles that do something similar that should be useful, a couple of examples:

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/ceca24e3871bbc26b9ed85d6ee59e9d28adb32ce/roles/ocp_on_libvirt/tasks/setup.yml#L75-L83

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/ceca24e3871bbc26b9ed85d6ee59e9d28adb32ce/roles/mirror_ocp_release/tasks/download.yml#L31-L36

tonyskapunk (changes requested):

roles/opcap_tool/tasks/build.yml
tkrishtop: Thanks for the advice, but I'm not fixing selinux part here, I'm fixing the go build timeout. We could keep your advice for later on.

tkrishtop:
> Thanks for the details, I was wondering if `pids_limit: "{{ -1 | int }}"` could work, that might get rid of the warning about the type string, but I do not know if the podman container collection will actually use it.

My main idea - to switch pids_limit from -1 to 0 - was wrong. In Dallas, we have podman 4.6.1 which [was released in 2023](https://github.com/containers/podman/blob/main/RELEASE_NOTES.md). That means that the fix to consider `-1` as `0` for `pids_limit` should be already here, as [it was merged in 2021](https://github.com/containers/podman/pull/11794/files).

I'm going to close this PR and think about the proper fix.

tkrishtop:
> > We also have a problem setting this value properly when using the podman_container module
> > ```
> > # code
> > containers.podman.podman_container:
> > -- snip --
> > pids_limit: -1
> > -- snip --
> > 
> > # error
> > [WARNING]: The value -1 (type int) in a string field was converted to '-1' (type string). 
> > If this does not look like what you expect, quote the entire value to ensure it does not change. 
> > ```
> 
> This is not an error, is a warning, it's telling that the value passed is an int and requires a string, but it casts a string for you. Doc [1]states it needs to be a string, to avoid the warning simply do
> 
> ```yaml
>      pids_limit: "-1"
> ```
> 
> [1] https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-pids_limit

noted

PR #165, opened 2024-02-12
Reduce the data to transfer to fix RPC error during TNF Git clone

tkrishtop:


tkrishtop:

tkrishtop:
hi @ramperher thank you for running the test, [everything seems to work fine on the first try, we even have no retries involved as usually for the HTTPS clone](https://www.distributed-ci.io/jobs/91a5924e-a085-4329-ae41-06462a608900/jobStates?sort=date&task=715548d1-5fbb-4532-abd0-d99c2a3a20b8).

roles/cnf_cert/defaults/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 tnf_stage: "tests"
 tnf_check_commit_sha: false
-test_network_function_repo: https://github.com/test-network-function/cnf-certification-test
+test_network_function_repo: "git@github.com:test-network-function/cnf-certification-test.git"
```
manurodriguez, line 4: don't we need SSH keys to pull with `git`? 
```
[f5@f5-jumphost ~]$ git clone git@github.com:test-network-function/cnf-certification-test.git
Cloning into 'cnf-certification-test'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
It works great when I use a server with an account that has a SSH key associated in github

tkrishtop: That's a good point. I checked some other places where we clone with the git module, it's everywhere via HTTPS + no credentials. 

Let's stick to that, I'll fallback to HTTPS and add an increased idle timeout then
```
environment:
  GIT_HTTP_MAX_REQUESTS: "4"
  GIT_HTTP_IDLE_TIMEOUT: "3600" 
```

tonyskapunk: > Let's stick to that, I'll fallback to HTTPS and add an increased idle timeout then

Stick to ssh or https?

IMO, we should try to keep it https since the repo is public, otherwise we are making it hard to consume

ramperher:
> hi @ramperher thank you for running the test, [everything seems to work fine on the first try, we even have no retries involved as usually for the HTTPS clone](https://www.distributed-ci.io/jobs/91a5924e-a085-4329-ae41-06462a608900/jobStates?sort=date&task=715548d1-5fbb-4532-abd0-d99c2a3a20b8).

I haven't run the job :sweat_smile: maybe it was our github-stream logic that automatically runs a tnf job if it detects a change in cnf_cert role.

I also agree with Manuel and Tony, I'd use https if possible

roles/cnf_cert/defaults/main.yml
tkrishtop: I'll fallback to HTTPS and add an increased idle timeout then

tkrishtop:
> > hi @ramperher thank you for running the test, [everything seems to work fine on the first try, we even have no retries involved as usually for the HTTPS clone](https://www.distributed-ci.io/jobs/91a5924e-a085-4329-ae41-06462a608900/jobStates?sort=date&task=715548d1-5fbb-4532-abd0-d99c2a3a20b8).
> 
> I haven't run the job 😅 maybe it was our github-stream logic that automatically runs a tnf job if it detects a change in cnf_cert role.
> 
> I also agree with Manuel and Tony, I'd use https if possible

Sure, I switched back to https and increased the idle timeout trying to avoid RPC issues. [The job is green so we could merge and see if that will help with dailies](https://www.distributed-ci.io/jobs/1bcc6862-6696-4341-9aae-5159aa77fea7/jobStates?sort=date&task=bd88f3cd-aa8a-4f02-9c43-d47d157b51b2).

roles/cnf_cert/tasks/pre-run.yml
```diff
@@ -155,12 +155,15 @@
         version: "{{ test_network_function_version }}"
         dest: "{{ tnf_dir }}"
         force: true
+      environment:
+        GIT_HTTP_MAX_REQUESTS: "4"
+        GIT_HTTP_IDLE_TIMEOUT: "3600"
```
manurodriguez, line 160: my Google fu is too basic that I couldn't find these variables 😛, I was just wondering what it means. Is this 4 retries of potentially 1 hour timeout each ? Or 4 request in a full hour, if the later then 15 mins looks good to me.

manurodriguez (approved):

roles/cnf_cert/tasks/pre-run.yml
betoredhat: @tkrishtop I think we are only interested in the latest commit and do not care about the whole history. This will significantly reduce the data that needs to be transferred. 

Do you think using shallow here will work?

 depth: 1

Create a shallow clone with a history truncated to the specified number or revisions. The minimum possible value is 1, otherwise ignored. Needs git>=1.9.1 to work correctly.

```
$ git clone --depth=1 --origin origin https://github.com/test-network-function/cnf-certification-test
Cloning into 'cnf-certification-test'...
remote: Enumerating objects: 448, done.
remote: Counting objects: 100% (448/448), done.
remote: Compressing objects: 100% (398/398), done.
remote: Total 448 (delta 85), reused 190 (delta 24), pack-reused 0
Receiving objects: 100% (448/448), 716.87 KiB | 1.59 MiB/s, done.
Resolving deltas: 100% (85/85), done.
$ du -hs cnf-certification-test/
6.2M	cnf-certification-test/
```

tkrishtop: It does not seem to be a problem with the size as it is, we really have some bytes missing
```
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
error: 769 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
```
It's more poor internet connection.

I found one more idea in [my google](https://docs.gitlab.com/ee/topics/git/troubleshooting_git.html#rpc-failed-curl-92-http2-stream-0-was-not-closed-cleanly-internal_error-err-2) though:
```
RPC failed; curl 92 HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR (err 2)
This problem may be caused by a slow internet connection. 
If you use Git over HTTP instead of SSH, try one of these fixes:

Increase the POST buffer size in the Git configuration with git config http.postBuffer 52428800.
Switch to the HTTP/1.1 protocol with git config http.version HTTP/1.1.
If neither approach fixes the error, you may need a different internet service provider.
```

tkrishtop: @betoredhat and @manurodriguez thank you for the reviews! I decided not to complicate life and went with the solution of Beto, adding `depth=1`  and `reference: "{{ tnf_dir }}"` as the clone options. The main issue is a slow network while cloning the large repo, so let's see, if that may fix the issue.

tkrishtop:
With the fix: [to clone takes 2 seconds](https://www.distributed-ci.io/jobs/279fb9a4-6fba-4ea3-b79c-7f73f31275ba/jobStates?sort=date&task=efdcf66a-d0cc-478b-86d3-f8bd81e4740a)
```
TASK [redhatci.ocp.cnf_cert : Clone tnf repository]
2s
task path: /--snip--/cnf_cert/tasks/pre-run.yml:152
changed: [jumphost] => {"after": "8531700557bef9d056efa56d776dfb6d050fb124", 
"attempts": 1, "before": null, "changed": true}
```

Before the fix: [to clone takes 9 seconds](https://www.distributed-ci.io/jobs/7b4bb2c7-d000-490a-bf14-01a20d8b35c7/jobStates?sort=date&task=f2fa0318-0b59-4630-9e1b-844c3fcf0f08)
```
TASK [redhatci.ocp.cnf_cert : Clone tnf repository]
9s
task path: /--snip--/cnf_cert/tasks/pre-run.yml:152
changed: [jumphost] => {"after": "8531700557bef9d056efa56d776dfb6d050fb124", 
"attempts": 1, "before": null, "changed": true}
```

I believe we're good to merge.

manurodriguez (approved):
LGTM

betoredhat (approved):
LGTM, thanks!

PR #166, opened 2024-02-12
Add RPM dependencies from requirements

tonyskapunk:
The roles require multiple packages that are not satisfied when installing the RPM alone. Adding all the dependencies required by the roles to allow this RPM to be used independently from DCI agents.


thekad (approved):
LGTM

ramperher (approved):
LGTM

PR #168, opened 2024-02-13
Download opcap binary instead of building it from scratch

tkrishtop:


tkrishtop:

roles/opcap_tool/tasks/build.yml
```diff
@@ -15,12 +15,15 @@
...
-    pids_limit: -1
+    pids_limit: "-1"
     log_level: "info"
     command: make -C /opcap build OPCAP_VERSION={{ opcap_version }}
     detach: false
     tty: true
     rm: true
+    # Increasing the memory limit from 512m to 1g
+    # to fix recurrent 137 OOM SIGKILL during the opcap build
+    memory: "1g"
```
capolrik, line 26: Nice finding! I did not found where this OOM error was coming from

capolrik (approved):

roles/opcap_tool/tasks/build.yml
manurodriguez: Just for curiosity, where did you find that the limit was 512m? I was searching online again but I couldn't find it 🤕,  in some podman-run references I saw that if no memory is specified then the container’s memory is not limited

manurodriguez (approved):

roles/opcap_tool/tasks/build.yml
tonyskapunk: I'm sorry but I'm not sure where the 512m limit was set, I looked at the docs but I couldn't find a default value, so I'm not sure how the 512 limit was found.

Correct me if I'm wrong, but the change suggest setting a limit rather than increasing it.

[1] https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-memory
[2] https://docs.podman.io/en/latest/markdown/podman-run.1.html#memory-m-number-unit

tkrishtop: Sorry, I put all the details in the [JIRA ticket](https://issues.redhat.com/browse/CILAB-1363). 

> where did you find that the limit was 512m

I did not find about the podman, only about the [docker](https://medium.com/codex/how-to-limit-cpu-and-memory-usage-in-docker-containers-a024b429f55e), where it's indeed 512m.

tkrishtop: @manurodriguez I have measured the memory usage locally with `top` while running opcap build task:
```
# Before the tests
MiB Mem :  31838.3 total,   1063.8 free,  23522.4 used,   7252.2 buff/cache

# During the opcap build
MiB Mem :  31838.3 total,    721.3 free,  23864.4 used,   7252.6 buff/cache
```

23864.4 used - 23522.4 used = about 300MiB.

For sure, the estimation is really rough, `top` usage + locally I ran opcap build with `podman run`  instead of `podman_container` module. But overall, looks reasonable, and justifies the idea to move up with 1g.

tkrishtop:

roles/opcap_tool/tasks/build.yml
thekad: I am also not sure if there even was a limit before, by setting a limit now to 1Gb (which is reasonable) you're effectively containing the opcap run and that protecting the host system from consuming all its memory and being OOM-killed. Maybe that's what was happening before?

tkrishtop: @thekad Maybe, that's totally possible. 

There are also [more layers](https://issues.redhat.com/browse/CILAB-1363) to the story: cgroups v2 vs cgroups v1. Currently, we have cgroup v1 on our Dallas jumphost:

- cgroup v1 does not support resource claims from rootless containers
- cgroup v1 [is not supported by podman guys](https://github.com/containers/podman/issues/17582) anymore

So if speak about a good solid fix, we should probably first migrate to cgroup v2. But I'm scared even to estimate possible consequences :)

tkrishtop:

tkrishtop:

tkrishtop:

tonyskapunk (commented):
This is another approach and I think is worth evaluating, the only caveat is that the release is quite old (from may 2023), not sure if we will be missing something as we were building the latest changes in the default branch.

Do you know if there would be any issue by not using latest changes?

roles/opcap_tool/tasks/build.yml
```diff
@@ -3,30 +3,24 @@
...
+- name: Ensure bin directory exists
+  ansible.builtin.file:
+    path: "{{ opcap_dir.path }}/opcap/bin"
+    state: directory
+
+- name: Download opcap release
+  vars:
+    opcap_archi: "{{ 'linux-amd64' if ansible_architecture == 'x86_64' else 'arm64' }}"
+  ansible.builtin.get_url:
+    url: "https://github.com/opdev/opcap/releases/download/v{{ opcap_version }}/opcap-{{ opcap_archi }}"
```
thekad, line 21: this url should be a variable too

tkrishtop:
> This is another approach and I think is worth evaluating, the only caveat is that the release is quite old (from may 2023), not sure if we will be missing something as we were building the latest changes in the default branch.
> 
> Do you know if there would be any issue by not using latest changes?

The release of May 2023 is the latest.

roles/opcap_tool/tasks/build.yml
```diff
@@ -1,34 +1,29 @@
 ---
-- name: "Clone opcap repository"  # noqa: latest[git]
+- name: Clone opcap repository  # noqa: latest[git]
```
capolrik, line 2: If we download directly the binary from GH, it is not necessary to git clone the repository anymore right?

capolrik (changes requested):

roles/opcap_tool/tasks/build.yml
tkrishtop: no, nice catch! Let me remove it, too

tonyskapunk:
> > This is another approach and I think is worth evaluating, the only caveat is that the release is quite old (from may 2023), not sure if we will be missing something as we were building the latest changes in the default branch.
> > Do you know if there would be any issue by not using latest changes?
> 
> The release of May 2023 is the latest.

Right, that's what I mentioned, the question is if there would be any issue switching to older code rather than the latest as we are currently doing.

tkrishtop:
> > > This is another approach and I think is worth evaluating, the only caveat is that the release is quite old (from may 2023), not sure if we will be missing something as we were building the latest changes in the default branch.
> > > Do you know if there would be any issue by not using latest changes?
> > 
> > 
> > The release of May 2023 is the latest.
> 
> Right, that's what I mentioned, the question is if there would be any issue switching to older code rather than the latest as we are currently doing.

Then we'll simply set `opcap_version` while running the job.
I discussed with Charles - we used `0.0.1` simply because it's been almost 2 years since we updated this role, and that's fine to change to the latest one.

tkrishtop:
The change is [working fine](https://www.distributed-ci.io/jobs/7fbe9879-292b-4d91-bd0a-fbd9ff473047/jobStates?sort=date&task=3d6923bc-41db-429a-a4e3-6834df74d87c). @capolrik could you please have a look to ensure that the latest version 0.2.1 produces the same results? Thanks.

capolrik (approved):
LGTM, a good way of getting ride of the random build errors, thanks!

tonyskapunk:
Removing from the merging Queue to have an answer on my last question, if there is no issue with using an older version, then please go ahead and merge.

PR #173, opened 2024-02-16
Use json string instead

tonyskapunk:
The value expected is a string representation of a json, during load the string is interpreted as json, we need to convert it back to a string


manurodriguez (approved):
LGTM

tonyskapunk:
This is the job for the validation: https://www.distributed-ci.io/jobs/60f37b72-a270-4831-8fab-1921987d5309

PR #175, opened 2024-02-19
Delete the collector

betoredhat:
- Delete the instance if already exists, as it will fail if an instance from a previous deployment already exists.
- Delete the instance during the teardown

### DIFF
--- a/roles/setup_netobserv_stack/tasks/cleanup.yml
+++ b/roles/setup_netobserv_stack/tasks/cleanup.yml
@@ -1,23 +1,11 @@
 ---
-- name: Get PVCs in the namespace
-  community.kubernetes.k8s_info:
-    api_version: v1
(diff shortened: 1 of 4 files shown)


thekad:
I think it should only be deleted once at the beginning. Also you can just delete the namespace and it should remove all operators/pods running afaik, but pass `wait: true` to make sure before continuing.

betoredhat:
> I think it should only be deleted once at the beginning. Also you can just delete the namespace and it should remove all operators/pods running afaik, but pass `wait: true` to make sure before continuing.

Yeah, found that the flowcollector is a non-namespaced resource. After deleting it at the beginning and removing the NS resources it is working fine. 

Thanks!

ramperher (approved):
LGTM

PR #178, opened 2024-02-26
add retries in metal-lb setup tasks

manurodriguez:
These objects do not have a status section to query then we are evaluating the task output result.


pierreblanc (approved):
LGTM

PR #180, opened 2024-02-27
Wait up to 45 minutes in preflight teardown after removing image source

ramperher:
This is to try to mitigate CILAB-1230.

### DIFF
--- a/roles/preflight/tasks/teardown.yml
+++ b/roles/preflight/tasks/teardown.yml
@@ -12,8 +12,8 @@
         name: redhatci.ocp.check_resource
       vars:
         resource_to_check: "MachineConfigPool"
-        check_wait_retries: 60
-        check_wait_delay: 20
+        check_wait_retries: 90
+        check_wait_delay: 30
         check_reason: "Delete Image Sources in preflight role"
 
     - name: Remove preflight Image Source file


tkrishtop:
I believe that waiting 45 for a job that should be finished within a 20-minute timeframe is a waste of time. It's better let it fail imho.

manurodriguez:
> I believe that waiting 45 for a job that should be finished within a 20-minute timeframe is a waste of time. It's better let it fail imho.

I believe the 45 mins are taking into account the reboot of the 4 workers, and because it's one node at that time we need at least 10 mins per server : (
But I agree that the ICSP change should take a couple of minutes and the reboot might be a bug or something new? maybe we should ask in OCP channels if that reboot is expected now, otherwise we open a bug, and we leave the 45min wait just as a work-around if we can't figure this out soon, wdyt?

ramperher:
> > I believe that waiting 45 for a job that should be finished within a 20-minute timeframe is a waste of time. It's better let it fail imho.
> 
> I believe the 45 mins are taking into account the reboot of the 4 workers, and because it's one node at that time we need at least 10 mins per server : ( But I agree that the ICSP change should take a couple of minutes and the reboot might be a bug or something new? maybe we should ask in OCP channels if that reboot is expected now, otherwise we open a bug, and we leave the 45min wait just as a work-around if we can't figure this out soon, wdyt?

Yes, as I said, this would be a temp workaround to confirm that nodes are able to recover by themselves. If not, then we'd have to think on another solution.

For me it also makes sense to ask in these forunms and channels, we can keep this change opened in the meantime.

manurodriguez (approved):
We agreed to go with this time increase t see if it helps, while we see other options to to avoid the deletion and hence the reboot.

tkrishtop (approved):
We agreed to merge during the daily

PR #181, opened 2024-02-28
Set a temporary workaround to opm issues

tonyskapunk:
Stable version of opm requires glibc libraries that are not available in el8. Pinning temporarily stable 4.14 to avoid this situation.


manurodriguez (approved):
LGTM

betoredhat (approved):
LGTM

PR #183, opened 2024-03-04
Move to Preflight 1.9.0 enforcing RequiredAnnotations check

tkrishtop:


tkrishtop:
check workload preflight-green

manurodriguez (approved):
LGTM

manurodriguez:
Hi @tkrishtop I checked the preflight and tnf jobs in the PR, and I see we're really using preflight 1.9.0, but I do not see where the tests validate for RequiredAnnotations? Could you please shed some light here? :pray:

tkrishtop:
> Hi @tkrishtop I checked the preflight and tnf jobs in the PR, and I see we're really using preflight 1.9.0, but I do not see where the tests validate for RequiredAnnotations? Could you please shed some light here? 🙏

Hi @manurodriguez thank you for checking it out, I've seen also that you have helped Brandon from TNF. 

RequiredAnnotations is a new mandatory test for operators, it's in every preflight_operator_* test, for example, [here it is for simple-demo](https://www.distributed-ci.io/jobs/4a37ac17-8004-47bc-af55-08413000f13c/tests/c40512bc-b52f-43ee-afcf-a5d09db0a671). The idea of Preflight team is to describe all operator features or requirements within these annotations: does it work in disconnected env, does it support proxy, etc etc, [here is an example](https://github.com/opdev/simple-demo-operator/commit/30e67bdbc382613eb0e2a2adcefb54e753a8b6da).

cc: @pierreblanc and @nsilla as you might see some failures for your partner once this PR is merged. You'll need to add [annotations like this](https://github.com/opdev/simple-demo-operator/commit/30e67bdbc382613eb0e2a2adcefb54e753a8b6da) to make it work again.

PR #194, opened 2024-03-07
Pin sushy tools < 1.1.0 and remote qemu URI

### LINKED ISSUE #174: Sushy-tools 1.1.0 moved virtual media interfaces from system manager to system
A recent change to sushy-tools in version 1.1.0 (https://opendev.org/openstack/sushy-tools/commit/a654a033dece53b8e6eb005025e37643d61144a5) moved the virtual media interfaces from system manager to systems.  The roles/vendors/kvm/tasks/iso.yml playbook in this collection will fail when attempting to run the tasks that eject, insert, and verify virtual media as they are coded to use the system manager interface.

thekad:
Closes #174
Also lets you configure a possible remote qemu URI, you shouldn't need this regularly but the feature is there
Random style fixes

### DIFF
--- a/roles/setup_sushy_tools/defaults/main.yml
+++ b/roles/setup_sushy_tools/defaults/main.yml
@@ -1,3 +1,4 @@
+---
 cert_vars_host_var_key: "{{ inventory_hostname }}"
 sushy_tools_port: 8082
 sushy_dir: "/opt/sushy-tools"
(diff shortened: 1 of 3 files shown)


betoredhat (approved):
LGTM, thanks

p3ck (approved):
lgtm

PR #195, opened 2024-03-08
[DNM] Test Image Source creation and deletion in preflight without launching tests

ramperher:
This is just to validate https://github.com/dci-labs/example-cnf-config/pull/98 and confirm that CILAB-1230 is gone, I'll close it when finishing


ramperher:
I've tested what I needed, so closing this

PR #199, opened 2024-03-11
Adding dependencies to galaxy spec

thekad:
We have requirements.yml, but this doesn't automatically pull the required dependencies when installing via `ansible-galaxy install`. This should take care of that


thekad:
Not 100% sure if we can entirely replace `requirements.yml` with something like `ansible-galaxy install .` for now I am just copying the content from the reqs file to the galaxy metadata file.

arjuhe (approved):
👍 
Thanks for submitting this.

fredericlepied (approved):

PR #200, opened 2024-03-12
Retries for installer extract

betoredhat:
Test-Hints: force-check

### DIFF
--- a/roles/mirror_ocp_release/tasks/installer.yml
+++ b/roles/mirror_ocp_release/tasks/installer.yml
@@ -14,6 +14,10 @@
     --command=openshift-baremetal-install
     --from {{ mor_pull_url }}
     --to "{{ mor_cache_dir }}/{{ mor_version }}"
+  register: extract_cmd
+  until: extract_cmd is succeeded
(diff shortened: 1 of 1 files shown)


betoredhat:
check dallas ocp-4.14-vanilla

manurodriguez (approved):
LGTM

thekad (commented):
This looks good to me, but I'm curious what failure you saw? I think this operation is pretty straightfoward if it fails it will fail always, do you have an example of a fail?

betoredhat:
> This looks good to me, but I'm curious what failure you saw? I think this operation is pretty straightfoward if it fails it will fail always, do you have an example of a fail?

Mainly time outs reaching the registry

https://www.distributed-ci.io/jobs/067511bb-dc2b-46f3-9066-159e7c9f5b9a/jobStates
https://www.distributed-ci.io/jobs/ff332ba2-1a65-4869-a68f-31d40f0de56e/jobStates

pierreblanc (approved):
LGTM

betoredhat:
Thanks

PR #202, opened 2024-03-13
dci-merge | use libvirt IPI by default

ramperher:
Please check CILAB-1526 to understand the reason why this is needed.


ramperher:
I think the solution is the correct one now, @fredericlepied

fredericlepied (approved):

PR #203, opened 2024-03-13
Fixing galaxy.yaml

thekad:
Maybe the keys do have to be strings


tkrishtop (approved):
thank you @thekad let's try ot out.

PR #205, opened 2024-03-13
galaxy.yml: fix the version of community.kubernetes to be compatible with DCI

fredericlepied:
- make a galaxy.yml change tested in CI


manurodriguez (approved):
LGTM

PR #208, opened 2024-03-13
Ansible collection community.crypto isn't packaged yet

thekad:
Removing from the dependency list until there's an RPM to provide it


manurodriguez (approved):
Thanks

PR #209, opened 2024-03-14
cnf_cert | Move some tasks from tests to teardown stage

ramperher:
Now that a tnf job can fail in the tnf execution, we're missing the logs in the DCI job, so moving these tasks and some others that are required for cleaning the environment to the teardown playbook, which is always executed

### DIFF
--- a/roles/cnf_cert/tasks/teardown.yml
+++ b/roles/cnf_cert/tasks/teardown.yml
@@ -16,6 +16,7 @@
   containers.podman.podman_image:
     name: "{{ tnf_image }}"
     state: absent
+  ignore_errors: true
 
 # Deleting resources created by the CNF Cert Suite to support the execution
 - name: Clean CNF Cert Suite resources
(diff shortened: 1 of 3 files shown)


roles/cnf_cert/tasks/teardown.yml
```diff
@@ -1,4 +1,22 @@
...
+    mode: "0750"
+  with_fileglob:
+    - "{{ tnf_dir }}/{{ test_network_function_project_name }}/tnf_config.yml"
+    - "{{ tnf_dir }}/{{ test_network_function_project_name }}/*.log"
+
+# Clean images just after finishing the execution
+- name: Remove local tnf image
+  containers.podman.podman_image:
+    name: "{{ tnf_image }}"
+    state: absent
```
tkrishtop, line 18: what about ignoring errors here to ensure the resource cleanup in the next task?

tkrishtop (approved):
LGTM, and the job is green. Just listed a possible minor improvement

ramperher:
> from change [test-network-function/cnf-certification-test#1919](https://github.com/test-network-function/cnf-certification-test/pull/1919):
> 
> * FAILURE https://www.distributed-ci.io/jobs/0f0456cb-2996-4615-960f-f237cc09a5cd/jobStates

In this failure, we've seen the logs are gathered correctly afterwards, so this is working

tkrishtop (approved):
LGTM

PR #210, opened 2024-03-14
updated to use role_path variable

mukrishn:
Updated the set_fact tasks to use the right default value for `extramanifestsopenshift_path` and `extramanifests_path`


manurodriguez (approved):
LGTM

PR #216, opened 2024-03-18
Remove release_image and os_images from defaults

nocturnalastro:
DCI already inserts these anyway and they are a maintenance burden, crucible will be requiring these to set in the inventory via validation. 

Test-Hints: assisted-abi


arjuhe:
This will definitely require some documentation and explaining. Perhaps the documentation can have a list of examples of previous versions?

thekad (changes requested):
I agree with @nocturnalastro you'll have to add an example to the README file otherwise people will get confused about the structure of the variable. I think a copy/paste of 1 or 2 of the structures you removed should be fine?

thekad (approved):

PR #217, opened 2024-03-18
Remove validate_inventory and display_deployment_plan roles

nocturnalastro:
These roles are crucible specific so is not needed in the collection. There was also a dependency of validate_inventory to some roles, this is also only needed in crucible.

Test-Hints: assisted-abi


thekad (approved):
Triggered dci-check-change again because it didn't have the right test hints initially, but now it validated it vs Assisted conf

ramperher (approved):
LGTM

PR #220, opened 2024-03-19
Check NetworkAttachmentDefinitions when creating SriovNetworks

ramperher:
Added this check just to wait for the creation of the related NetworkAttachmentDefinitions when creating SriovNetworks.

### DIFF
--- a/roles/sriov_config/tasks/check_sriov_network.yml
+++ b/roles/sriov_config/tasks/check_sriov_network.yml
@@ -1,0 +1,12 @@
+---
+- name: Wait until NetworkAttachmentDefinition is created from SriovNetwork {{ sriov_conf.network.name }}
+  community.kubernetes.k8s_info:
+    api_version: k8s.cni.cncf.io/v1
+    name: "{{ sriov_conf.network.name }}"
+    namespace: "{{ sriov_conf.network.network_namespace }}"
+    kind: NetworkAttachmentDefinition
+  register: net_attach_def_check
+  retries: 6
+  delay: 10
+  until: net_attach_def_check.resources|length == 1
+  no_log: true
--- a/roles/sriov_config/tasks/check_sriov_network.yml
+++ b/roles/sriov_config/tasks/check_sriov_network.yml
@@ -3,7 +3,7 @@
   community.kubernetes.k8s_info:
     api_version: k8s.cni.cncf.io/v1
     name: "{{ sriov_conf.network.name }}"
-    namespace: "{{ sriov_conf.network.network_namespace }}"
+    namespace: "{{ sriov_conf.network.network_namespace | default('default') }}"
     kind: NetworkAttachmentDefinition
   register: net_attach_def_check
   retries: 6
--- a/roles/sriov_config/README.md
+++ b/roles/sriov_config/README.md
@@ -14,6 +14,7 @@ Nodes must have SR-IOV capable network interfaces.
 | ----------------------------- | --------- | ------- | --------- | -----------
 | sriov_config_file             | Yes       | String  | Undefined | The configuration file for SR-IOV Node policy and/or network.
 | sriov_config_wait_node_policy | No        | Boolean | True      | Whether or not wait for node policies to apply
+| sriov_config_wait_network     | No        | Boolean | True      | Whether or not wait for networks to apply the proper NetworkAttachmentDefinition resources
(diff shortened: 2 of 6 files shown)


manurodriguez (approved):
LGTM

ramperher:
This check looks like working fine, I'll also wait for @tonyskapunk 's opinion on this since we were discussing about this before merging

ramperher:
> from change [dci-labs/dallas-config#130](https://github.com/dci-labs/dallas-config/pull/130):
> 
> * FAILURE https://www.distributed-ci.io/jobs/59c3eccc-e719-4265-88b4-41dfc72e343f/jobStates

I've had a case where the example-cnf deployment is failed, however this check is passing. I'll hold on this change a little bit, because I want to see if we need to do some extra checks in the SRIOV side before proceeding with the workload.

roles/sriov_config/tasks/check_sriov_network.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Wait until NetworkAttachmentDefinition is created from SriovNetwork {{ sriov_conf.network.name }}
+  community.kubernetes.k8s_info:
+    api_version: k8s.cni.cncf.io/v1
+    name: "{{ sriov_conf.network.name }}"
+    namespace: "{{ sriov_conf.network.network_namespace }}"
```
tonyskapunk, line 6: Since ns is not required, default to "default" ns
```suggestion
    namespace: "{{ sriov_conf.network.network_namespace | default('default') }}"
```

tonyskapunk (commented):
Change looks good, just needs one case to be considered.

roles/sriov_config/tasks/check_sriov_network.yml
```diff
@@ -0,0 +1,12 @@
...
+    kind: NetworkAttachmentDefinition
+  register: net_attach_def_check
+  retries: 6
+  delay: 10
+  until: net_attach_def_check.resources|length == 1
+  no_log: true
```
tkrishtop, line 12: should we remove no_log here to ease the debug?

tkrishtop (approved):
LGTM

roles/sriov_config/tasks/check_sriov_network.yml
ramperher: Maybe @tonyskapunk can clarify on this, but I assume the `no_log` is placed here to avoid disclosing network information that may be sensible, right?
Honestly, the output is not really throwing relevant information, it's just saying that the resource is created and its specification can be found in the variable holding the SRIOV config, so maybe it's not a big deal to remove the `no_log`

roles/sriov_config/tasks/check_sriov_network.yml
ramperher: ack

ramperher:
Let me rework a little bit this to see if I need to add some more conditions here, since example-cnf is not working as expected with 4 SriovNetwork scenario. I'll move to Ready when finished.

roles/sriov_config/tasks/check_sriov_network.yml
tkrishtop: I believe no_log was to remove the bulky logs here, we sometimes use it by default. But here the logs are not that bulky I guess and could really help when debugging

ramperher:
The issue with example-cnf looks like to be related to race condition issues when creating the pods under test, so I think no more changes are needed in this PR, this is ready for review

ramperher:
> from change [dci-labs/dallas-config#130](https://github.com/dci-labs/dallas-config/pull/130):
> 
> * SUCCESS https://www.distributed-ci.io/jobs/42835189-8001-40de-99b2-16b0f7d2ff48/jobStates

Loadbalancer case is definitely working, I'll not launch more tests here, this is ready for review.

tonyskapunk (approved):
:+1:

roles/sriov_config/tasks/check_sriov_network.yml
ramperher: For the moment, if you agree, I'll merge this, since I have other changes depending on this that I need to merge, but we can explore this in a future change, in case we need it.
But I'd say we're hiding here the logs to avoid exposing net info that may be sensible, that's my assumption. Anyway let's revisit it in a new change if required

PR #223, opened 2024-03-20
Adding the microshift_generate_iso role

fdaencarrh:
This role is used to generate the microshift iso.

Test-Hints: no-check


roles/microshift_generate_iso/library/get_latest_rhocp_repo.py
tonyskapunk, line 1: In collections modules must live within plugins top dir[1]. For this it would be in plugins/modules.

[1] https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_structure.html#roles-directory

roles/microshift_generate_iso/tasks/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+- import_tasks: "{{ microshift_generate_iso_action }}.yml" # noqa name[missing]
```
tonyskapunk, line 2: Could you add some validation for the variable, to only allow expected values.

Also, is there any reason why to import and not include the task? Generally is preferred to include unless there's a particular reason on the import.

fdaencarrh: No particular reason, I will do the change

roles/microshift_generate_iso/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+microshift_generate_iso_action: "install"
+microshift_generate_iso_action_allowed: ['install']
```
thekad, line 3: if this is the only action allowed why even make it dynamic?

roles/microshift_generate_iso/tasks/creating_iso.yml
```diff
@@ -0,0 +1,71 @@
+---
+- name: Create the microshift installer file
+  ansible.builtin.template:
+    src: microshift-installer.toml.j2
+    dest: "{{ microshift_generate_iso_folder }}/microshift-installer.toml"
+    mode: '0644'
+  register: microshift_installer_template
```
thekad, line 7: this variable isn't used anywhere, don't register it

roles/microshift_generate_iso/tasks/creating_rhel_edge_image.yml
```diff
@@ -0,0 +1,70 @@
...
+  ansible.builtin.command:
+    composer-cli -j compose status
+  delay: 30
+  retries: 200
+  until:
+    (image_build_status.stdout | from_json | json_query(query_finished)) == "FINISHED"
+  register: image_build_status
+  vars:
+    query_running: "[].body.run[?id=='{{ build_id }}']|[0][0].queue_status"
+    query_finished: "[].body.finished[?id=='{{ build_id }}']|[0][0].queue_status"
```
thekad, line 39: you only really use `query_finished` remove the other 2 from this task

roles/microshift_generate_iso/tasks/creating_rhel_edge_image.yml
```diff
@@ -0,0 +1,70 @@
...
+  become: true
+  ansible.builtin.command:
+    composer-cli compose image {{ build_id }} --filename {{ downloaded_edge_image_path }}
+  changed_when: true
+
+- name: Change permission of the image
+  become: true
+  ansible.builtin.file:
+    path: "{{ downloaded_edge_image_path }}"
+    mode: a+r
```
thekad, line 57: you should also chown this to something like ansible_user

roles/microshift_generate_iso/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- include_tasks: "{{ microshift_generate_iso_action }}.yml" # noqa name[missing]
+  when: microshift_generate_iso_action in microshift_generate_iso_action_allowed
```
thekad, line 3: you don't need to make this dynamic, just include install.yml

roles/microshift_generate_iso/templates/kickstart.ks.j2
```diff
@@ -0,0 +1,112 @@
...
+zerombr
+clearpart --all --disklabel gpt
+part biosboot --fstype=biosboot --size=1
+part /boot/efi --fstype=efi --size=200
+part /boot --fstype=xfs --asprimary --size=800
+# Uncomment this line to add a SWAP partition of the recommended size
+#part swap --fstype=swap --recommended
+part pv.01 --grow
+volgroup rhel pv.01
+logvol / --vgname=rhel --fstype=xfs --size=10240 --name=root
```
thekad, line 33: why not use `--grow` here instead of `--size`?

roles/microshift_generate_iso/templates/kickstart.ks.j2
```diff
@@ -0,0 +1,112 @@
...
+
+# Make sure redhat user directory contents ownership is correct
+chown -R redhat:redhat /home/redhat/
+
+# Configure the firewall (rules reload is not necessary here)
+firewall-offline-cmd --zone=trusted --add-source=10.42.0.0/16
+firewall-offline-cmd --zone=trusted --add-source=169.254.169.1
+
+# Make the KUBECONFIG from MicroShift directly available for the root user
+echo -e 'export KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig' >> /root/.profile
```
thekad, line 73: You should add this to the `redhat` user profile

roles/microshift_generate_iso/tasks/creating_iso.yml
fdaencarrh: For now, I want to add the cleanup later

roles/microshift_generate_iso/templates/kickstart.ks.j2
fdaencarrh: the redhat user doesn't have access to this file.

roles/microshift_generate_iso/defaults/main.yml
tonyskapunk: I think this is part of the feedback I provided. By doing this it avoids injecting any value into `microshift_generate_iso_action` allowing only to load controlled "actions", for now is only install.yml, and not loading any other file.

roles/microshift_generate_iso/templates/kickstart.ks.j2
fdaencarrh: grow will fill all the available space. Why I should use size instead?

roles/microshift_generate_iso/tasks/main.yml
fdaencarrh: For now yes, but I want to add the cleanup later. I can remove it for now if needed

roles/microshift_generate_iso/defaults/main.yml
fdaencarrh: I simply wish to introduce another option like the cleanup. If necessary, I can remove it for now.

roles/microshift_generate_iso/templates/kickstart.ks.j2
fdaencarrh: The remainder of the volume will be used by the CSI driver for storing data

README.md
tonyskapunk, line 1: Could you include the plugin module in the plugins section (line 131), thanks!

thekad (approved):

PR #224, opened 2024-03-20
Adding community.crypto, removing requirements.yml

thekad:
The dependency is already packaged. The file requirements.yml is not used anywhere that I can see, just leaving dependency info in galaxy.yml


fredericlepied:
@thekad how can we activate the code in an install?

tonyskapunk (approved):
:+1:

thekad:
> @thekad how can we activate the code in an install?

This should be called from the `setup_selfsigned_cert` role in the assisted-pre-run playbook

recheck

thekad:
recheck

fredericlepied:
> > @thekad how can we activate the code in an install?
> 
> This should be called from the `setup_selfsigned_cert` role in the assisted-pre-run playbook

How can we activate this setup_selfsigned_cert role in a test? I suppose that's not used by default else it would have failed in the past.

thekad:
> > > @thekad how can we activate the code in an install?
> > 
> > 
> > This should be called from the `setup_selfsigned_cert` role in the assisted-pre-run playbook
> 
> How can we activate this setup_selfsigned_cert role in a test? I suppose that's not used by default else it would have failed in the past.

You're right, it's been skipped because setup_registry_service is set to false e.g. https://www.distributed-ci.io/jobs/e143f675-e758-44ce-8a3f-e259352b0dab/jobStates?sort=date&task=dac055f2-de1e-4910-b3da-c705057298f5 . I don't want to just turn on that flag on BOS2, I'll have to test this out on my own lab.

thekad:
I am having issues getting a successful end-to-end run for Assisted (my configuration + not enough disk space) but the jobs progress past the steps where the self-signed certificates are created https://www.distributed-ci.io/jobs/54c147a0-4a64-45cd-b58d-5935ced3b5e8/jobStates?sort=date&task=8b7d72fc-b1b6-4317-9d42-63369c9e8bf0 this shows that the collection is working as it would fail otherwise without the community.crypto collection installed

betoredhat (approved):
LGTM

thekad:
rebased

PR #229, opened 2024-03-26
Sushy tools may not have authentication

thekad:
The BMC user and password may be not required if you didn't set up authentication

Test-Hints: assisted-abi


nocturnalastro (approved):
Seems sensible to me :)

PR #234, opened 2024-03-28
Make webserver_caching_image consistent with setup_http_store

nocturnalastro:
Change was prompted by quay.io/centos7/httpd-24-centos7 removing the latest tag which was what was being used previously

See https://quay.io/repository/centos7/httpd-24-centos7?tab=history

Test-Hints: libvirt
Test-Args-Hints: -e cache_enabled=True

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -9,7 +9,7 @@ install_config_appends_file: install-config-appends.yml
 registry_auth_file: registry-auths.json
 disconnected_registry_user: dummy
 disconnected_registry_password: dummy
-webserver_cache_image: "quay.io/centos7/httpd-24-centos7:latest"
+webserver_cache_image: "quay.io/fedora/httpd-24:latest"
 webserver_caching_port: "{{ webserver_caching_port_container }}"
 webserver_caching_port_container: 8080
 registry_creation: false
--- a/roles/sno_installer/defaults/main.yml
+++ b/roles/sno_installer/defaults/main.yml
@@ -2,7 +2,7 @@
 # defaults file for sno_installer
 cache_enabled: true
 provision_cache_store: "{{ cache_dir }}"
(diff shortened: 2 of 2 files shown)


nocturnalastro:
@fredericlepied I dunno how these get executed so no idea what the hints should be. I'm guessing one should be sno.

tonyskapunk:
> I added it as an edit no idea if that kicks off another run or not

The edit does not trigger the CI, after the change you need to re-run the 'DCI / Maybe Schedule...' check to retrigger the execution with that hint.

But I'm afraid the already executed IPI install and the SNO will not run the change because that needs the cache to be enabled and we don't have it in BOS2.

Honestly I'm not sure if it would work enabling it, but let's try. I'm modifying the hints to pass that variable enabled :crossed_fingers:

ramperher:
> > I added it as an edit no idea if that kicks off another run or not
> 
> The edit does not trigger the CI, after the change you need to re-run the 'DCI / Maybe Schedule...' check to retrigger the execution with that hint.
> 
> But I'm afraid the already executed IPI install and the SNO will not run the change because that needs the cache to be enabled and we don't have it in BOS2.
> 
> Honestly I'm not sure if it would work enabling it, but let's try. I'm modifying the hints to pass that variable enabled 🤞

Nit: To test SNO with AI, `sno-ai` hint should be used instead of `sno`

tonyskapunk:
> SNO with AI, sno-ai hint should be used instead of sno

Does `sno-ai` uses `webserver_caching_image` ?

nocturnalastro:
Its in roles `installer` and `sno_installer` so I guess its probably not using the assisted installer.

tonyskapunk:
sno is not easy to test as it requires baremetal and our testing for sno uses VM. I switched to tes the libvirt of IPI

tonyskapunk (approved):
Alright, the test with IPI shows the use of the new image, let's get this merged.

Thanks!

PR #236, opened 2024-04-01
Unique fork

betoredhat:
Test-Hints: force-check


roles/hco_setup/tasks/deploy-hco.yml
tonyskapunk, line 1: Seems like this one was not intended, correct?

betoredhat: Ups!... thanks!

ramperher (approved):
LGTM

PR #240, opened 2024-04-04
Bump to tnf v5.0.6

manurodriguez:


pierreblanc (approved):
LGTM

manurodriguez:
In `tnf-test-cnf-green-pipeline.yml` the test `platform-alteration-tainted-node-kernel` is whitelisted, but in `tnf-test-cnf-pipeline.yml` is not, @ramperher do you know this is expected?

ramperher:
> In `tnf-test-cnf-green-pipeline.yml` the test `platform-alteration-tainted-node-kernel` is whitelisted, but in `tnf-test-cnf-pipeline.yml` is not, @ramperher do you know this is expected?

Yes, it's expected, because tnf-test-cnf-green is meant to be only run in cluster10, where we always have the kernel taint due to the Dell servers, but tnf-test-cnf is always used in cluster1-6, where we don't have kernel taints by default.

roles/cnf_cert/defaults/main.yml
```diff
@@ -23,7 +23,7 @@ tnf_image_suffix: "{{ lookup('community.general.random_string', length=16, speci
 # Provide feedback for each tnf test launched. Just copy this template in your pipeline file and fill the value for the
 # tests you want to provide feedback. You don't need to include the tests that you don't want to provide feedback.
 # To insert line breaks, use \\n (you need to escape it). Same for quotes and so on.
-# List of tests updated until tnf v5.0.5
+# List of tests updated until tnf v5.0.6
 tnf_feedback:
```
ramperher, line 27: tnf_feedback must be updated for this version to include some new tests added for tnf v5.0.6:
```
operator-crd-openapi-schema
operator-semantic-versioning
```

ramperher (changes requested):
I was waiting for this release to make an additional change according to the typical ones, just pointing you to the modification we need to make.

sebrandon1 (approved):

roles/cnf_cert/defaults/main.yml
sebrandon1: Just a heads up, have some additional `operator-` tests in flight and we weren't going to bump the minor version to v5.1.x until they were all available for use.

ramperher: > Thanks for the heads up, I see the tests are passing in both TNF jobs, so I was wondering what could be a relevant feedback to add? Perhaps something about possible failure and introduction in 5.0.6?

It's just to add an entry in the tnf_feedback default value, which is an empty string for all tests available. This is mainly to have a kind of template that can be used afterwards to fill in the tests that are really failing

manurodriguez: Thanks for the clarification, fixed in the last commit!

ramperher (approved):
Now it's perfect :)

PR #241, opened 2024-04-05
Remove mor_build unused variable in mirror_ocp_release role

rh-gvincent:
mor_build is required but not used in the role


tonyskapunk (approved):
Thanks, is great to remove unused code!

PR #242, opened 2024-04-05
Use mor_oc variable instead of oc_tool_path in mirror_ocp_release

rh-gvincent:
`mor_oc` variable describe in the readme is not used. We are using `oc_tool_path` instead.

### DIFF
--- a/roles/mirror_ocp_release/tasks/download.yml
+++ b/roles/mirror_ocp_release/tasks/download.yml
@@ -11,7 +11,7 @@
 - name: "Extract artifacts from release image"
   ansible.builtin.command:
     cmd: >
-      {{ oc_tool_path }} adm release extract
+      {{ mor_oc }} adm release extract
       --registry-config {{ mor_auths_file }}
       --tools
       --from {{ mor_pull_url }}
--- a/roles/mirror_ocp_release/tasks/installer.yml
+++ b/roles/mirror_ocp_release/tasks/installer.yml
@@ -9,7 +9,7 @@
 
 - name: "Extract installer command from release image"
   ansible.builtin.command: >
-    {{ oc_tool_path }} adm release extract
+    {{ mor_oc }} adm release extract
     --registry-config={{ mor_auths_file }}
     --command=openshift-baremetal-install
     --from {{ mor_pull_url }}


rh-gvincent:
Maybe we want to add backward compatibility and copy oc_tool_path into mor_oc if setted ?
wdyt ?

ramperher:
> Maybe we want to add backward compatibility and copy oc_tool_path into mor_oc if setted ? wdyt ?

In fact, in the agent, mor_oc is defined to be oc_tool_path:

```
[raperez@raperez-thinkpadp1gen3 dci-openshift-agent]$ grep -r mor_oc
plays/acm-pre-run.yml:        mor_oc: "{{ oc_tool_path }}"
plays/acm-pre-run.yml:        mor_oc: "{{ oc_tool_path }}"
plays/assisted-pre-run.yml:          mor_oc: "{{ oc_tool_path }}"
plays/pre-run.yml:    mor_oc: "{{ oc_tool_path }}"
plays/pre-run.yml:        mor_oc: "{{ oc_tool_path }}"
```

So I suppose it's fine to do the change

ramperher (approved):
LGTM

rh-gvincent:
I was thinking about partner or other consumer of this role

tonyskapunk (approved):
> Maybe we want to add backward compatibility and copy oc_tool_path into mor_oc if setted ? wdyt ?

I think the way you have it right now is the right one as the role requires `mor_oc`.

tkrishtop:
> I was thinking about partner or other consumer of this role


> `mor_oc` variable describe in the readme is not used. We are using `oc_tool_path` instead.

if you use oc_tool_path everywhere, why do you change oc_tool_path with moc_oc? That's a bit confusing :)

ramperher:
> > I was thinking about partner or other consumer of this role
> 
> > `mor_oc` variable describe in the readme is not used. We are using `oc_tool_path` instead.
> 
> if you use oc_tool_path everywhere, why do you change oc_tool_path with mor_oc? That's a bit confusing :)

I suppose it's mainly to decouple the agents from the collections. The roles from collections should not depend on variables created in the agent.

PR #244, opened 2024-04-05
Document and improve get logs from namespace role

tonyskapunk:
The role was using variables defined in the dci-openshift-app-agent, removing those references and allowing it to be called independently. Create its documentation.

Goes in hand with #242 to make the roles independent from the DCI agents.

Test-Hints: no-check


roles/get_logs_from_namespace/tasks/main.yml
```diff
@@ -1,46 +1,70 @@
...
+    namespace: "{{ glfn_ns }}"
   register: pod_list
 
 - name: "Print pods names"
   ansible.builtin.debug:
-    msg: "pod_list: {{ pod_list | json_query('resources[*].metadata.name') }}"  # noqa: jinja[invalid]
+    msg: "pod_list: {{ pod_list | json_query('resources[*].metadata.name') }}"
 
-- name: "Get logs from each pods"  # noqa: jinja[invalid]
+- name: "Get logs from each pods"
```
betoredhat, line 43:
```suggestion
- name: "Get logs from each pod"
```

ramperher (approved):
LGTM

PR #247, opened 2024-04-08
Remove semver lib dependency

pierreblanc:
The semver lib is provided by an EPEL rpm package.
We want to remove external dependency.

PR #249, opened 2024-04-09
add a unit test for ocp_compatibility.py

fredericlepied:


fredericlepied:
@tkrishtop if you want to add a real test, it would be great ;-)

tests/unit/requirements.txt
```diff
@@ -0,0 +1,2 @@
+junit_xml
+semver
```
tkrishtop, line 2: @fredericlepied we're currently re-writing ocp_compatibility.py with @pierreblanc to get rid of semver: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/247/files

tkrishtop: would it be fine if we first merge another PR and then re-test yours?

tkrishtop (approved):

.github/workflows/pr.yml
```diff
@@ -113,13 +113,17 @@ jobs:
             sed -e 's/3.5 //g')
           for version in $PY_VERS; do
             ansible-test sanity --verbose --docker --python $version --color --coverage --failure-ok
+            ansible-test units --verbose --docker --python $version --color --coverage || :
```
tonyskapunk, line 116: Is there a reason why the `--failure-ok` may not be enough here?

It is changed here in the coming `branch` but is not changed in the `main` branch.

fredericlepied: Good catch! Yes, there is a reason: it doesn't work the units sub-command.

fredericlepied: that's also there on the `main` branch...

tkrishtop:
> @tkrishtop if you want to add a real test, it would be great ;-)

@fredericlepied, yes, we discussed this with @pierreblanc yesterday. We'll add a real output listing some deprecated API from a DCI job. We currently use this output to test locally

tkrishtop (approved):
let's merge this change first, before merging #247, as we discussed during the daily. This way, we'll check how the check works.

.github/workflows/pr.yml
tonyskapunk: Nevermind, I did look it wrong. All good :+1:

tonyskapunk (approved):
LGTM

PR #251, opened 2024-04-09
Use ansible k8s module to get cluster version

tonyskapunk:
Avoid relying in a binary to get the cluster version, instead get the version from the cluster itself through the ansible modules.


betoredhat (approved):
LGTM

thekad (approved):
:sheep: it

PR #261, opened 2024-04-13
Create a custom rule module for ansible-lint

thekad:
Ansible lint allows you to write custom rules and add them to the list of things you want to check. Adding custom rules to start enforcing naming convention guidelines in the repo.
Also reworked `hack/ansible-lint.sh` to be used in the same way both in GHA and in your development environment.

### DIFF
--- a/hack/ansible-lint.sh
+++ b/hack/ansible-lint.sh
@@ -17,33 +17,32 @@
 # under the License.
 #
 #~Usage:
-#~  ansible-lint.sh [-f] [-p PROFILE]
+#~  ansible-lint.sh [-b BIN] [-d] [PATH]
 #~
 #~Options:
-#~  -c      Remove .linted markers
-#~  -p      Use the "production" ansible-lint profile
-#~  -n      Don't clear screen after linting each role
+#~  -b BIN  Path to the `ansible-lint` binary (default: $PATH/ansible-lint)
+#~  -d      Download requirements declared in galaxy.yml (default: False)
+#~Arguments:
+#~  PATH    (Optional) path to lint, defaults to the top level directory
 
 function usage(){
     grep "^#~" "$0" | sed -e "s/#~//"
 }
 
+set -x
+
 TOPDIR="$(git rev-parse --show-toplevel)"
 CWD="$PWD"
-CLEAN=""
-PROFILE="shared"
-CLEAR="yes"
+BIN="$(which ansible-lint)"
+OFFLINE="--offline"
 
-while getopts ":cpn" o; do
+while getopts ":b:d" o; do
     case $o in
-        c)
-            CLEAN="yes"
+        b)
+            BIN="$OPTARG"
             ;;
-        p)
-            PROFILE="production"
-            ;;
-        n)
-            CLEAR=""
+        d)
+            OFFLINE=""
             ;;
         *)
             usage
@@ -52,29 +51,22 @@ while getopts ":cpn" o; do
     esac
 done
 
-# Be verbose from this point forward
-set -x
+shift $((OPTIND-1))
+EXTRA_ARGS="$@"
+
+if ! test -x $BIN; then
+    echo "You need to install ansible-lint. Ideally from the source code branch"
+    echo "Example:"
+    echo -e "pip install 'ansible-lint[lock] @ git+https://github.com/ansible/ansible-lint@v6'"
+    exit 1
+fi
 
 cd "$TOPDIR"
 
-if [ "$CLEAN" == "yes" ]; then
-    rm -fv roles/*/.linted
-fi
-
-for role in roles/*; do
-    if [ "$CLEAR" == "yes" ]; then
-        clear
-    fi
-    echo "$role"
-    if [ -f "$role/.linted" ]; then
-        continue
-    else
-        if ansible-lint --profile="$PROFILE" "$role"; then
-            touch "$role/.linted"
-        else
-            break
-        fi
-    fi
-done
+$BIN \
+    $OFFLINE \
+    --force-color \
+    --parseable \
+    $EXTRA_ARGS
 
 cd "$CWD"
--- a/hack/rules/redhat_ci.md
+++ b/hack/rules/redhat_ci.md
@@ -1,0 +1,73 @@
+# redhat-ci
+
+This rule checks for compliance with some coding conventions in the OCP
+collection
+
+The `redhat-ci` rule has the following checks:
+
+- `redhat-ci[no-role-prefix]` - Variables used within a role should have a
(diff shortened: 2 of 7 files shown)


thekad:
@tonyskapunk @fredericlepied @redhatci/crucible  ptal and let me know what you think. This rule should enforce what we have been doing already

thekad:
Mostly I want your input in the logic behind the prefix

hack/ansible-lint.sh
```diff
@@ -17,33 +17,35 @@
...
-        c)
-            CLEAN="yes"
-            ;;
         p)
             PROFILE="production"
             ;;
-        n)
-            CLEAR=""
+        b)
+            BINARY="$OPTARG"
```
p3ck, line 45: Should this be BIN instead of BINARY?  Otherwise I don't see how the which check below will work?

thekad:
> There are a few places in the crucible roles where variables are shared across roles `fetched_dest` for instance, how will that be handled?
> 
> perhaps having a prefix of `global_` or `shared_` might be useful for them
> 
> or do we put the overhead on the caller rename?

This might be a good idea, allowing you to set global scope facts (this would only be allowed for set_fact) so e.g.
1. You import role A
2. Inside role A you `set_fact global_cool_fact`
3. You import role B, passing `vars: b_cool_fact: global_cool_fact` and only referencing `b_cool_fact` inside role B as I would like to avoid abusing the global scope a lot

Keep in mind that rules here can be deactivated with `noqa: redhat-ci[no-role-prefix]` just like any other rule, that should be the exception

hack/rules/redhat_ci.md
```diff
@@ -0,0 +1,65 @@
...
+  1. If there's a single word, use the first $SHORT characters
+  1. Finally, if there are 2 or more words, build an acronym from the split
+     words
+
+  Examples:
+  | Role Name               | Prefix    |
+  |-------------------------|-----------|
+  | `sno`                   | `sno_`    |
+  | `installer`             | `insta_`  |
+  | `validate_http_store`   | `vhs_`    |
```
tonyskapunk, line 30: Would it be possible for this custom profile to include the var-naming role-prefix, just in case any role is already following that? otherwise it will need to switch back to use these rules.
```Shell
  | Role Name       | Prefix           |
  |-----------------|------------------|
  | custom_role_foo | custom_role_foo_ |
```

hack/ansible-lint.sh
tonyskapunk, line 1: For debugging could you please add -x to the shebang?

hack/ansible-lint.sh
```diff
@@ -52,29 +49,25 @@ while getopts ":cpn" o; do
...
-        continue
-    else
-        if ansible-lint --profile="$PROFILE" "$role"; then
-            touch "$role/.linted"
-        else
-            break
-        fi
-    fi
-done
+ansible-lint \
```
tonyskapunk, line 67: Right before this line could you do an `env` 
I got the feeling that we are seeing different behaviors because something in the environment

thekad:
@tonyskapunk I have added what you requested here https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/8841457894/job/24278589639?pr=261 I also found the same issue you'd seen before, the file `roles/acm_setup/tasks/main.yml` had a non FQCN task see diffence vs https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/8841230920/job/24278001753
I don't understand why this is happening but I have created an issue ansible/ansible-lint#4122

tonyskapunk:
After long troubleshooting I found a way to reproduce the same behavior as in the CI (GHA).

In my env I couldn't replicate the results from the CI. @thekad you also can't reproduce it, we both had the same amount of findings (thousands).

But isolating the environment helped me to reproduce this:

```Shell
WARNING  Listing 3 violation(s) that are fatal
roles/acm_setup/tasks/main.yml:29:3: syntax-check[unknown-module]: couldn't resolve module/action 'k8s_info'. This often indicates a misspelling, missing collection, or incorrect module path.
roles/vendors/supermicro/tasks/disk.yml:56: warning[outdated-tag]: Replaced outdated tag 'var-spacing' with 'jinja', replace it to avoid future errors
roles/vendors/zt/tasks/power_off.yml:9: warning[outdated-tag]: Replaced outdated tag 'unnamed-task' with 'name', replace it to avoid future errors

-- snip ---
                     Rule Violation Summary                      
 count tag                          profile rule associated tags 
     1 syntax-check[unknown-module] min     core, unskippable    
     2 warning[outdated-tag]                core, experimental   
```


To reproduce:

```Shell
podman run --rm -it -v ${PWD}:/tst quay.io/centos/centos:stream9
```

> Careful with selinux, might complain...


Once in the container, install the needful:

```Shell
## python
dnf install python3.11 python3-pip -y

## ansible-lint
pip install ansible-lint

## requirements
pip install junitparser junit_xml jmespath
```

Reproduce the current CI:

```Shell
ansible-lint --parseable --nocolor --profile=shared -v >isol.log 2>&1
```

Is still unclear to me why as soon as we add a correction in `roles/acm_setup/tasks/main.yml` in this environment, then the same behavior we see "outside" takes place :shrug: 

```Diff
index 7a69f377..744b5f69 100644
--- a/roles/acm_setup/tasks/main.yml
+++ b/roles/acm_setup/tasks/main.yml
@@ -27,7 +27,7 @@
         disableHubSelfManagement: "{{ hub_disable_selfmanagement }}"
 
 - name: "Wait for search-postgres deployment"
-  k8s_info:
+  community.kubernetes.k8s_info:
kind: Deployment
     namespace: "{{ hub_namespace }}"
     name: search-postgres
```

ansible-lint results:
```Shell
Failed: 637 failure(s), 1102 warning(s) on 861 files. Profile 'shared' was required, but 'min' profile passed.
```

I'm of the opinion to merge this change but without the fix in acm_setup, and tackle the issues in different changes, otherwise we are going to block every new PR.

roles/acm_setup/tasks/main.yml
```diff
@@ -27,7 +27,7 @@
         disableHubSelfManagement: "{{ hub_disable_selfmanagement }}"
 
 - name: "Wait for search-postgres deployment"
-  k8s_info:
+  community.kubernetes.k8s_info:
```
tonyskapunk, line 30: I think is better to get this fixed in a different change to avoid the explosion of issues that could prevent others to merge changes

tonyskapunk: After our sync, we agreed to include this one as upcoming changes will compare to this, so we shouldn't see much issues, despite being there :+1:

tonyskapunk (approved):
:+1:

PR #265, opened 2024-04-18
Use podman_image module to remove root-owned image

tkrishtop:
Test-Hints: no-check


ramperher (approved):
LGTM

PR #269, opened 2024-04-23
Add a role to manage firewalld

thekad:
Mostly doing it for the ability to turn masquerading off, but adding some basic FW features to it.


ramperher (approved):
Looks nice to me, I know there are some installer roles where it's performed some firewalld config (e.g. assisted), maybe we can leverage on this new role as next steps for these installers.

PR #270, opened 2024-04-23
Allow other GH protocols

betoredhat:
Test-App-Args-Hints: -e git_protocol=https -e do_chart_verifier=true -e github_token_path=/opt/cache/token.txt -e partner_name=fredco -e partner_email=telcoci -e {"dci_charts":[{"name":"samplechart","chart_file":"/opt/cache/charts/samplechart/samplechart-0.1.1.tgz","values_file":["/opt/cache/charts/samplechart/values.yaml","/opt/cache/charts/samplechart/values2.yaml"],"deploy_chart":true,"create_pr":false},{"name":"remotechart","chart_file":"http://jumpbox.partnerci.bos2.lab/charts/samplechart-0.1.2.tgz","values_file":["http://jumpbox.partnerci.bos2.lab/charts/values1.yaml","http://jumpbox.partnerci.bos2.lab/charts/values2.yaml"],"deploy_chart":true,"create_pr":true},{"name":"remotechart","chart_file":"http://jumpbox.partnerci.bos2.lab/charts/samplechart-0.1.2.tgz","deploy_chart":true,"create_pr":false}]}

### DIFF
--- a/roles/create_pr/README.md
+++ b/roles/create_pr/README.md
@@ -13,6 +13,7 @@ partner_email                       | undefined | true/false                  | 
 target_repository                   | undefined | true                        | GitHub repository where the pull request will be created.
 product_type                        | undefined | true                        | Product type. Either "helmchart" or "operator".
 publish_chart                       | true      | false                       | For helm charts, disable adding the Chart as part of the pull request.
+git_protocol                        | ssh       | false                       | Protocol for GitHub communication (ssh/https)
 
 Those are the common variables used by both certification project.
 It includes tasks to generate an SSH key needed to push to GitHub repository and add it to the GitHub account.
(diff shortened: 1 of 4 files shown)


roles/create_pr/README.md
```diff
@@ -13,6 +13,7 @@ partner_email                       | undefined | true/false                  |
 target_repository                   | undefined | true                        | GitHub repository where the pull request will be created.
 product_type                        | undefined | true                        | Product type. Either "helmchart" or "operator".
 publish_chart                       | true      | false                       | For helm charts, disable adding the Chart as part of the pull request.
+git_protocol                        | ssh       | false                       | Protocol for GitHub communication (ssh/https)
```
manurodriguez, line 16: maybe a nice to have is to include a validation in the main.yml to make sure `git_protocol` contains either `ssh` or `https`, wdyt?

manurodriguez (commented):
Thanks Beto, just to understand, does this means that despite we are using `https` protocol now, we still need to create the SSH key pair and do the key setup right? 
https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/create_pr/tasks/main.yml#L49-L75

roles/create_pr/README.md
thekad, line 16: don't you already have this fragment from the `target_repository` variable? you can infer it by doing something like `git_protocol: "{{ target_repository | urlsplit('scheme') }}"`

PR #277, opened 2024-04-26
Delete hcp cluster only if exists

betoredhat:
- Trying to delete the cluster will fail if it actually does not exists
- Checking if cluster exists before attempt to remove it

Test-Hints: no-check


tonyskapunk (approved):
:+1:

PR #283, opened 2024-04-30
acm_setup: hub_os_images must be optional

nsilla:
hub_os_images must be optional to work on connected environments.

Test-Hints: no-check


betoredhat (approved):
LGTM

PR #284, opened 2024-05-02
Use container to extract rhcos from installer

tonyskapunk:
Starting 4.16 (nightlies) openshift BM install uses el9 libraries that does not run under el8, due to FIPS compliance. Using a container for this task reduces the need of the ansible controller using el9.

---

- [x]  TestDallas: ocp-4.16-vanilla  -> https://www.distributed-ci.io/jobs/ebdbbff6-4826-4ebc-969d-c73628245826?sort=date&task=12f08727-6a1b-464c-9d50-f54d981560e6

---
Test-Hints: no-check

### DIFF
--- a/roles/mirror_ocp_release/tasks/facts.yml
+++ b/roles/mirror_ocp_release/tasks/facts.yml
@@ -7,12 +7,27 @@
   register: release_image
   changed_when: false
 
-- name: "Write rhcos.json from command (>= 4.8)"
+- name: Write rhcos.json from command (4.16+)
+  when: mor_version is version("4.16", ">=")
+  containers.podman.podman_container:
+    name: extract-rhcos
+    image: quay.io/centos/centos:stream9
+    command:
+      - /bin/bash
+      - -c
+      - "/ocp/openshift-baremetal-install coreos print-stream-json > /ocp/rhcos-test.json"
+    rm: true
+    volumes:
+      - "{{ mor_cache_dir }}/{{ mor_version }}:/ocp"
+
+- name: "Write rhcos.json from command (4.8 - 4.15)"
   ansible.builtin.shell: >
     {{ mor_cache_dir }}/{{ mor_version }}/openshift-baremetal-install coreos print-stream-json >
     {{ mor_cache_dir }}/{{ mor_version }}/rhcos.json
   when:
+    - mor_version is version("4.16", "<")
     - mor_version is version("4.8", ">=")
+  changed_when: true
 
 # TODO: Remove this block when 4.7 is no longer supported
 - name: "Download rhcos.json (< 4.8)"
--- a/roles/mirror_ocp_release/tasks/facts.yml
+++ b/roles/mirror_ocp_release/tasks/facts.yml
@@ -15,7 +15,7 @@
     command:
       - /bin/bash
       - -c
-      - "/ocp/openshift-baremetal-install coreos print-stream-json > /ocp/rhcos-test.json"
+      - "/ocp/openshift-baremetal-install coreos print-stream-json > /ocp/rhcos.json"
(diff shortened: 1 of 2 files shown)


roles/mirror_ocp_release/tasks/facts.yml
```diff
@@ -7,12 +7,27 @@
   register: release_image
   changed_when: false
 
-- name: "Write rhcos.json from command (>= 4.8)"
+- name: Write rhcos.json from command (4.16+)
+  when: mor_version is version("4.16", ">=")
```
betoredhat, line 11: The binary from the previous version can be executed using centos:stream9. 

What about using the container approach for all versions > 4.8?

betoredhat: We addressed this comment offline. All good.

betoredhat (dismissed):
LGTM

roles/mirror_ocp_release/tasks/facts.yml
```diff
@@ -7,12 +7,27 @@
...
+  containers.podman.podman_container:
+    name: extract-rhcos
+    image: quay.io/centos/centos:stream9
+    command:
+      - /bin/bash
+      - -c
```
manurodriguez, line 17: From our discussion in Slack, I thought the final fix was using a `|` after the `-c`?
```
          - -c
          - |
            /ocp/openshift-baremetal-install coreos print-stream-json > /ocp/rhcos-test.json
```

roles/mirror_ocp_release/tasks/facts.yml
```diff
@@ -7,12 +7,27 @@
...
+      - "/ocp/openshift-baremetal-install coreos print-stream-json > /ocp/rhcos-test.json"
+    rm: true
```
manurodriguez, line 19: I guess there is no need of these options right?
```
        interactive: no
        tty: yes
        detach: yes
        debug: true
```

roles/mirror_ocp_release/tasks/facts.yml
tonyskapunk: Ah, good observation, I thought about it as well. But I realized that the truly important part is to have the command *and* the redirect in the same line, I tested without the pipe and got the expected result.

roles/mirror_ocp_release/tasks/facts.yml
tonyskapunk: Correct, those are not required

tonyskapunk:
> from change #284:
> 
>     * SUCCESS https://www.distributed-ci.io/jobs/14ab55ee-ac73-4e23-8d70-7e135d931e28/jobStates

This comes from a test in Bos2 with libvirt, I made the change to test it in the disconnected environment, where it is used. That is in the queue

roles/mirror_ocp_release/tasks/facts.yml
betoredhat: yeah, no pipe is needed. I tested locally and works

ramperher (dismissed):
LGTM, waiting for the dallas test to finally confirm it's working as expected

tonyskapunk:
check dallas ocp-4.16-vanilla

tonyskapunk:
recheck

tonyskapunk:
recheck

tonyskapunk:
This change has been validated here: https://www.distributed-ci.io/jobs/ebdbbff6-4826-4ebc-969d-c73628245826?sort=date&task=12f08727-6a1b-464c-9d50-f54d981560e6

The job will fail and that's expected as the failure will be later in the run.

betoredhat (approved):
nice, lgtm!

PR #285, opened 2024-05-02
Allow to pass a target org for the operators to mirror

betoredhat:
- It is needed in case operators need to be loaded at a separate registry path.

Test-Hints: no-check


betoredhat:
Operators have been mirrored to the proper path at the local registry.

ramperher (approved):
LGTM

PR #292, opened 2024-05-07
Add general role-collection of useful GitOps tasks

sshnaidm:
Sometimes there are useful automation tasks that we can use, but they don't deserve a role. Collect them in one role and then use as `tasks_from` of specific file.

### DIFF
--- a/roles/gitops_operations/README.md
+++ b/roles/gitops_operations/README.md
@@ -1,0 +1,35 @@
+gitops_operations
+=================
+
+This Ansible role contains varios tasks for GitOps operations.
+
+Requirements
+------------
+
+* Access to an OpenShift cluster
+* Properly configured kubeconfig file for cluster access
+
+Role Variables
+--------------
+
+* `gops_kubeconfig_file`: (string) The file path of kubeconfig file for cluster. Default: undefined, If not provided, and no other connection options are provided, the Kubernetes client will attempt to load the default configuration file from ~/.kube/config. Can also be specified via `K8S_AUTH_KUBECONFIG` environment variable.
+* `gops_app_name`: (string) The name of Application in GitOps.
+* `gops_namespace`: (string) The namespace to use.
+
+Example Playbook
+----------------
+
+* Sync repo for ArgoCD if autosync is disabled
+
+```yaml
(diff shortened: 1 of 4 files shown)


roles/gitops_operations/README.md
```diff
@@ -0,0 +1,35 @@
+gitops_operations
+=================
+
+This Ansible role contains varios tasks for GitOps operations.
```
ramperher, line 4: typo: various

roles/gitops_operations/README.md
```diff
@@ -0,0 +1,35 @@
...
+* Properly configured kubeconfig file for cluster access
+
+Role Variables
+--------------
+
+* `gops_kubeconfig_file`: (string) The file path of kubeconfig file for cluster. Default: undefined, If not provided, and no other connection options are provided, the Kubernetes client will attempt to load the default configuration file from ~/.kube/config. Can also be specified via `K8S_AUTH_KUBECONFIG` environment variable.
+* `gops_app_name`: (string) The name of Application in GitOps.
+* `gops_namespace`: (string) The namespace to use.
+
+Example Playbook
```
ramperher, line 19: Rather than "Example Playbook", you may have a chapter here to describe the different operations that are included in this role. E.g. sync_repo, since it's the only one currently available

roles/gitops_operations/README.md
```diff
@@ -0,0 +1,35 @@
...
+```yaml
+    - hosts: localhost
+      tasks:
+
+        - name: Force sync repo for "clusters" Applicaion in ArgoCD
+          include_role:
+            name: gitops_operations
+          tasks_from: sync_repo
+          vars:
+            gops_kubeconfig_file: /path/to/kubeconfig
```
ramperher, line 33: instead of providing the kubeconfig as variable and then using it on each task of the playbook, you may pass the KUBECONFIG env var in the following way:


```
         - name: Force sync repo for "clusters" Applicaion in ArgoCD
          include_role:
            name: gitops_operations
          tasks_from: sync_repo
          vars:
            gops_app_name: clusters
          apply:
            environment:
              KUBECONFIG: /path/to/kubeconfig

```

PR #214, opened 2024-03-15
Switch to kubernetes.core collection

tonyskapunk:
The kubernetes.core is the new name of the community.kubernetes ansible collection. community.kubernetes has been deprecated.

PR #295, opened 2024-05-08
Add kubernetes.core as dependency

sshnaidm:
Use 2.3.2 fixed version to prevent possible breakages

---

TestBos2: virt

### DIFF
--- a/ansible-collection-redhatci-ocp.spec
+++ b/ansible-collection-redhatci-ocp.spec
@@ -21,6 +21,7 @@ Requires: ansible-collection-community-general
 Requires: ansible-collection-community-kubernetes
 Requires: ansible-collection-community-libvirt
 Requires: ansible-collection-containers-podman
+Requires: ansible-collection-kubernetes-core
 Requires: git
 Requires: jq
 Requires: podman
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -103,6 +103,7 @@ dependencies:
   "community.kubernetes": "1.2.1"
   "community.libvirt": "*"
   "containers.podman": "*"
+  "kubernetes.core": "2.3.2"
 
 # The URL of the originating SCM repository
 repository: https://github.com/redhatci/ansible-collection-redhatci-ocp


galaxy.yml
```diff
@@ -103,6 +103,7 @@ dependencies:
   "community.kubernetes": "1.2.1"
   "community.libvirt": "*"
   "containers.podman": "*"
+  "kubernetes.core": "3.0.0"
```
tonyskapunk, line 106: This should match the RPM package version available (currently [2.3.2](https://softwarefactory-project.io/r/plugins/gitiles/ansible-collection-kubernetes-core)) so both installation methods (RPM and collections) use the same versions

ansible-collection-redhatci-ocp.spec
tonyskapunk, line 1: Let's bump the RPM version.

tonyskapunk (changes requested):
There's some related work in #214, but I like that we can make one step at a time and  move closer to kubernetes core in here.

galaxy.yml
sshnaidm: Ok, let's use 2.3.2 then. Seems like latest Ansible binary depends on kubernetes.core < 3, and it can't be updated.

tonyskapunk:
The [dev#pull-requests](https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/docs/dev.md#pull-requests) doc contains information on how to test accordingly the changes, in this case it should be enough to validate that the change  is not breaking anything :).

recheck

tonyskapunk:
recheck

sshnaidm:
@tonyskapunk is there anything still blocking this?

tonyskapunk:
> @tonyskapunk is there anything still blocking this?

Yes, it requires a bump to the RPM and the collections, more than happy to help if needed.

sshnaidm:
@tonyskapunk sorry, I'm not sure what you mean. What is bumping RPM and collection?

tonyskapunk:
Ah, my bad, sorry for not being clear,  please increase the minor version for the RPM spec and the galaxy config file:

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/9547bf3aba2e10569aa8e17b096c1ee6bd111590/galaxy.yml#L12

and

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/9547bf3aba2e10569aa8e17b096c1ee6bd111590/ansible-collection-redhatci-ocp.spec#L6

PR #298, opened 2024-05-09
HCP - Patch registry overrides

betoredhat:
- Patch the hypershfit operator deployment with registry overridden in format source=target, this allows the ignition server to get the corresponding image digests required during the deployment.

Test-Hints: no-check


betoredhat:
Not required

PR #302, opened 2024-05-14
Adding a new RHOAI role

thekad:
This will manage the Red Hat Openshift AI operator stack

Test-Hints: no-check
TestDallas: test


thekad:
@tonyskapunk @ylamgarchal @fdaencarrh  PTAL

thekad:
This is a sample output of the role running in the App agent https://www.distributed-ci.io/jobs/b4fba74d-b532-4599-bbd3-47c839d67ac5/jobStates?sort=date

ramperher (commented):
Thanks for this good work! I have some general questions:
- I suppose it's tested in the app-agent because it's probably the easiest way of doing it using hooks, but, should we include some support in the dci-openshift-agent to install this operator, such as we currently do with other operators?
- If that's true, should we reuse OLM-related roles to install the operator, rather than creating new playbooks that does the same?

thekad:
> Thanks for this good work! I have some general questions:
> 
> * I suppose it's tested in the app-agent because it's probably the easiest way of doing it using hooks, but, should we include some support in the dci-openshift-agent to install this operator, such as we currently do with other operators?

I don't think that is necessary, this is not a whole new feature and the end result will be something that cannot be easily testable (you can only verify it got installed)

> * If that's true, should we reuse OLM-related roles to install the operator, rather than creating new playbooks that does the same?

I toyed with the idea of calling `olm_operator` but for now we can leave as is and we can swap it later

ylamgarchal:
LGTM

ylamgarchal (approved):
LGTM

thekad:
recheck

betoredhat:
retest

tonyskapunk:
recheck

Issue #296, opened 2024-05-08
kubeconfig requirements and limitations in roles


tonyskapunk:
Opening this issue to discuss and understand the limitations/requirements on kubeconfig for k8s related roles.

### Problem

It has been expressed that some environments have limitations on the use of ENV variables, preventing them to use roles that require  KUBECONFIG ENV var to be used in a role.

### Proposal

We would like to use this issue to understand any kind of limitation or requirement on the use of k8s related roles to make them more accessible to all. Please document any limitation/requirement in this issue.

---

cc: @sshnaidm

lack:
Given that we'll probably often be using `community.kubernetes` or one day `kubernetes.core`, we should come up with some idiom that makes it easy for our roles to use that collection, and it may be prudent to make sure our way of receiving credentials matches the various ways that collection works.

In the sideload_kernel role, for example, I came up with this:
- https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/sideload_kernel/README.md#authentication

Basically my role expects a variable called `k8s_auth` which is expected to be a dictionary whith values that correspond to the individual environment variables that the underlying `community.kubernetes` collection expects.

The reason for this kind of design is that it is hard to pass environment variables directly to roles, but not hard for a role to take a variable and put it in the environment any time we call into that collection.  For an example, see this:
- https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/sideload_kernel/tasks/main.yml#L7-L13

tonyskapunk:
> The reason for this kind of design is that it is hard to pass environment variables directly to roles, but not hard for a role to take a variable and put it in the environment any time we call into that collection.

It is possible to pass ENV to a role. I found it is the same amount of effort to *define* ENV or var in a role, but I see some advantages with the ENV.

As a variable:
```yaml {.numberLines}
- name: Role Foo
  vars:
    k8s_auth:
      K8S_AUTH_KUBECONFIG: /var/lib/clusterauth/kubeconfig
  ansible.builtin.include_role:
    name: redhatci.ocp.foo
```

It requires all the task that rely on the variable to use `environment: "{{ k8s_auth }}"`

As an environment variable:

```yaml
- name: Role Foo
  ansible.builtin.include_role:
    name: redhatci.ocp.foo
    apply:
      environment:
        K8S_AUTH_KUBECONFIG: /var/lib/clusterauth/kubeconfig
```

[...]

PR #305, opened 2024-05-15
Add option to specify kubeconfig in olm_operator

sshnaidm:
Add optional parameter `olm_operator_kubeconfig` to specify exact kubeconfig file to use. If not specified, defined or it's empty, then the previous default behaviour keeps working.

### DIFF
--- a/roles/olm_operator/README.md
+++ b/roles/olm_operator/README.md
@@ -6,6 +6,7 @@ Role to deploy an OLM-based operator.
 
 Name                        | Required  | Default                | Description
 --------------------------- |-----------|------------------------|--------------------------------------
+olm_operator_kubeconfig     | No        | empty                  | Path to kubeconfig file to use. If not specified, then omitted.
 channel                     | No        | \<defaultChannel\>     | The default channel of the operator is calculated and used when undefined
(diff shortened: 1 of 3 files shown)


thekad:
@sshnaidm we usually do apply the kubeconfig env when including the role, something like:
```yaml
- ansible.builtin.include_role:
    name: olm_operator
    vars:
      <config>
    apply:
      environment:
        KUBECONFIG: <path to KC>
```

Which should work unless more than 1 KC file is used in the role, would that work for you?

sshnaidm:
> @sshnaidm we usually do apply the kubeconfig env when including the role, something like:
> 
> ```yaml
> - ansible.builtin.include_role:
>     name: olm_operator
>     vars:
>       <config>
>     apply:
>       environment:
>         KUBECONFIG: <path to KC>
> ```
> 
> Which should work unless more than 1 KC file is used in the role, would that work for you?

We discuss this in issue: #296

sshnaidm:
@tonyskapunk @thekad is there anything blocking this PR?

tonyskapunk:
@sshnaidm before actually doing the changes we need to reach consensus  in #296.  This is good as a PoC and a validation that things won't break, but once #296 is solved we can proceed with this one.

PR #311, opened 2024-05-21
Extract commit ID with ttp instead of shell and awk

reihl:
Now in case the command failed the error appear immediately, also the output is more easy to read and understand

### DIFF
--- a/roles/installer/tasks/23_rhcos_image_paths.yml
+++ b/roles/installer/tasks/23_rhcos_image_paths.yml
@@ -2,15 +2,30 @@
 - name: RHCOS image path (pre 4.8)
   block:
     - name: Get COMMIT_ID
-      shell: |
-        /usr/local/bin/{{ installer_cmd }} version | grep '^built from commit' | awk '{print $4}'
-      register: commit_id
+      changed_when: false
+      ansible.builtin.command: |
+        /usr/local/bin/{{ installer_cmd }} version
+      register: installer_cmd_output
+      tags: rhcospath
+
+    - name: Parse openshift_baremetal_install command with TTP
+      ansible.utils.cli_parse:
+        text: "{{ installer_cmd_output['stdout'] }}"
+        parser:
+          name: ansible.utils.ttp
+          template_path: "{{ role_path }}/templates/cmd_openshift-baremetal-install.ttp"
+        set_fact: installer_openshift_baremetal_install_fact
+      tags: rhcospath
+
+    - name: Set commit_id
+      ansible.builtin.set_fact:
+        commit_id: "{{ installer_openshift_baremetal_install_fact[0][0].commit }}"
       tags: rhcospath
 
     - name: Get the URLs
       set_fact:
         offline_url: "{{ webserver_url }}/{{ version }}/rhcos.json"
-        online_url: "https://raw.githubusercontent.com/openshift/installer/{{ commit_id.stdout }}/data/data/rhcos.json"
+        online_url: "https://raw.githubusercontent.com/openshift/installer/{{ commit_id }}/data/data/rhcos.json"
 
     - name: Get RHCOS JSON File
       uri:
--- a/ansible-collection-redhatci-ocp.spec
+++ b/ansible-collection-redhatci-ocp.spec
@@ -28,6 +28,7 @@ Requires: python3-jmespath
 Requires: python3-netaddr
 Requires: python3-pyyaml
 Requires: skopeo
+Requires: ttp
 
 %description
 %{summary}.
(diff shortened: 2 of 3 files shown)


fredericlepied:
retest

roles/installer/tasks/23_rhcos_image_paths.yml
```diff
@@ -2,15 +2,30 @@
 - name: RHCOS image path (pre 4.8)
   block:
     - name: Get COMMIT_ID
-      shell: |
-        /usr/local/bin/{{ installer_cmd }} version | grep '^built from commit' | awk '{print $4}'
-      register: commit_id
```
tonyskapunk, line 7: Could fail sooner:
```yaml
    - name: Get COMMIT_ID
      shell: |
        set -xeo pipefail
        /usr/local/bin/{{ installer_cmd }} version | grep '^built from commit' | awk '{print $4}'
      register: commit_id
    
```

tonyskapunk (commented):
I agree that is best to avoid shell whenever is possible, but I have some thoughts for this change.

- This is a change for a deprecated version of OCP
- We are adding a dependency (if there is a package for ttp)
- The change removes the shell but ends up executing the command to parse it later

I added a suggestion to fail it sooner, if that would make the experience better, wdyt?

roles/installer/tasks/23_rhcos_image_paths.yml
reihl: I agree that this approach is better now, I removed the ttp requirement, however, the test didn't visit the tasks we changed so, I can't tell if the requirement in ansible.util take an action. Can I suggest another way to parse the output? I think we can still use command without pipe and get the information we want with some ansible filters

tonyskapunk: > he test didn't visit the tasks we changed

Correct, it is because this is only applicable for OCP 4.8 or older.  Are you consuming OCP 4.8 or anything older?

> Can I suggest another way to parse the output? I think we can still use command without pipe

Sure

ramperher (commented):
Can we test this with OCP 4.7? Because this is the target version for this change.

reihl:
> Can we test this with OCP 4.7? Because this is the target version for this change.

I can run it on my ci machine, is it enough?

fredericlepied:
retest

manurodriguez:
> TestDallas: ocp-4.17-vcp

We probably mean `ocp-4.7-vcp`?

tonyskapunk:
> > TestDallas: ocp-4.17-vcp
> 
> We probably mean `ocp-4.7-vcp`?

I talked to @reihl and we will be testing with 4.8, I've made the change in the body of the PR

---

recheck

tonyskapunk:
> Can we test this with OCP 4.7? Because this is the target version for this change.

Good point, checking this one with @reihl

tonyskapunk:
The last failure is due to a maintenance going on in RDO project, I'll retry once it's done.

ramperher:
> > > TestDallas: ocp-4.17-vcp
> > 
> > 
> > We probably mean `ocp-4.7-vcp`?
> 
> I talked to @reihl and we will be testing with 4.8, I've made the change in the body of the PR

But the change only affects to OCP 4.7, doesn't it? According to this when condition:

```
  when: (release_version.split('.')[0]|int == 4) and (release_version.split('.')[1]|int <= 7)
```

reihl:
> > > > TestDallas: ocp-4.17-vcp
> > > 
> > > 
> > > We probably mean `ocp-4.7-vcp`?
> > 
> > 
> > I talked to @reihl and we will be testing with 4.8, I've made the change in the body of the PR
> 
> But the change only affects to OCP 4.7, doesn't it? According to this when condition:
> 
> ```
>   when: (release_version.split('.')[0]|int == 4) and (release_version.split('.')[1]|int <= 7)
> ```

Correct, I also change the when condition to be more readable with ansible.builtin.version filter. We face it because something that ignore we don't need rhcos.json for version newer then 4.7 and this is the reason I fix it.

tonyskapunk:
> But the change only affects to OCP 4.7, doesn't it? According to this when condition:

@ramperher correct, I pinged @reihl to see if he can share logs as he is using OCP 4.8, but the change in here shouldn't affect 4.8, so I'm not sure how exactly this is going to fix something in 4.8

tonyskapunk:
>  We face it because something that ignore we don't need rhcos.json for version newer then 4.7 and this is the reason I fix it.

Could you please elaborate a bit more here?

For 4.8 and above a different set of tasks are performed to use `oc  coreos print-stream-json`, if you're using 4.8 this is the tasks that should get you what's needed without having to go to a commit as in 4.7-.

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/2f66e0f84b3037236cb408aafd57fb8d79b35f9e/roles/installer/tasks/23_rhcos_image_paths.yml#L39-L43

reihl:
retest

PR #312, opened 2024-05-22
Move to Preflight 1.9.6 making DeployableByOlm to validate packagename

tkrishtop:


ramperher (approved):
LGTM

PR #314, opened 2024-05-22
Remove packages from RDO repo

tonyskapunk:
The RDO repo has been removed after a maintenance for removal of centos8[0][1], switching to install the pyghmi package through the available repositories (not available in CentOS) python3-crypt is not needed by the latest python3-pyghmi package.

Removed the ipmi_power module that required those specific packages as it breaks "talking" to vBMC. Switching to ipmitool that works and removes the need of keeping those deprecated packages.

[0] https://lists.rdoproject.org/archives/list/dev@lists.rdoproject.org/thread/NTG4PWBTQ6KBD2JRGNUVMCFYMVX4LODP/
[1] https://lists.rdoproject.org/archives/list/dev@lists.rdoproject.org/thread/Q3AD6A5B3RM5ENAN6MCXICCVTQJYNLTD/

---

Test-Hints: no-check

PR #315, opened 2024-05-23
Fix warning messages  in "Power off hosts via IPMI" task by changing state to string

reihl:


roles/installer/tasks/59_power_off_cluster_servers.yml
```diff
@@ -19,12 +19,12 @@
         - "{{ groups.workers | default([]) }}"
 
     - name: Power off hosts via IPMI
-      ipmi_power:
+      community.general.ipmi_power:
```
manurodriguez, line 22: @reihl in case you haven't seen this, today @tonyskapunk modified this task to use `ansible.builtin.command` module instead of any ansible ipmi module because of some deprecated python packages, please see #314

reihl:
Changed by other PR (314)

PR #323, opened 2024-05-29
Patch body format

shaior:
Test-Hints: no-check


fredericlepied:
retest

roles/jenkins_job_launcher/tasks/main.yml
```diff
@@ -16,6 +16,7 @@
     user: "{{ jjl_username }}"
     password: "{{ jjl_token }}"
     force_basic_auth: true
+    body_format: form-urlencoded
```
natifridman, line 19: Is this something needed for every job or just for a specific one? If not for all, maybe it's better to add a variable with something default

natifridman (approved):
LGTM

PR #324, opened 2024-05-29
Add dry-run to mirror_catalog role

tonyskapunk:
In some cases is useful to only generate the manifests without actually mirroring images of a catalog. With this new option, mc_manifest_only is possible to only generate the manifest files.

---

- [x] TestDallas: ocp-4.15-vanilla - https://www.distributed-ci.io/jobs/f0531dd6-8b81-4250-954e-28bd8f29b405/jobStates

---

Test-Hints: no-check

### DIFF
--- a/roles/mirror_catalog/tasks/main.yml
+++ b/roles/mirror_catalog/tasks/main.yml
@@ -56,6 +56,9 @@
       {% endif %}
       --max-components=3
       --continue-on-error={{ mc_continue_on_error }}
+      {% if mc_manifest_only | bool %}
+      --manifests-only=true
+      {% endif %}
       {% if mc_pullsecret is defined %}
       --registry-config {{ mc_pullsecret }}
       {% endif %}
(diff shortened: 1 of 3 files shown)


roles/mirror_catalog/tasks/main.yml
```diff
@@ -56,6 +56,9 @@
       {% endif %}
       --max-components=3
```
betoredhat, line 57: This is unrelated to your change, but considering the layout of the prega repo we will need to increase this to 4.

tonyskapunk: I'll add a way to modify this then

betoredhat (approved):
LGTM, thanks!

tonyskapunk:
The failure is unrelated to this change, as I was testing a connected environment.

The test for this change is in the successful job on OCP 4.15 https://www.distributed-ci.io/jobs/f0531dd6-8b81-4250-954e-28bd8f29b405/jobStates

tonyskapunk:
recheck

PR #326, opened 2024-05-30
Aligning the ZTP related roles with CI ztphub role

nsilla:
@sshnaidm crated the [PR 262](https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/262) contributing the role ztphub, which overlaps with some of the already existing roles.
This PR is to align the existing roles with the requirements observed in the ztphub role.

### DIFF
--- a/roles/acm_setup/tasks/main.yml
+++ b/roles/acm_setup/tasks/main.yml
@@ -181,6 +181,8 @@
 
 - name: "Update the provisioning resource to watch all Namespaces"
   community.kubernetes.k8s:
+    merge_type:
+      - merge
     definition:
       apiVersion: metal3.io/v1alpha1
       kind: Provisioning
@@ -188,6 +190,7 @@
         name: provisioning-configuration
       spec:
         watchAllNamespaces: true
+        disableVirtualMediaTLS: true
 
 - name: "Get cluster version"
   community.kubernetes.k8s_info:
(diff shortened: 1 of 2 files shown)


roles/acm_setup/tasks/main.yml
```diff
@@ -181,13 +181,16 @@
...
+    merge_type:
+      - merge
     definition:
       apiVersion: metal3.io/v1alpha1
       kind: Provisioning
       metadata:
         name: provisioning-configuration
       spec:
         watchAllNamespaces: true
+        disableVirtualMediaTLS: true
```
nsilla, line 193: @betoredhat, I'd like to know you opinion on whether this can be left like this or it should be made optional.

betoredhat: @nsilla sure, let's add a variable to control this. I haven't seen many environments where the BMCs have verified TLS certificates.

betoredhat (approved):
LGTM

roles/acm_setup/tasks/main.yml
```diff
@@ -258,6 +261,10 @@
       metadata:
         name: agent
         namespace: multicluster-engine
+        {% if hub_skip_verify_tls %}
```
reihl, line 264: Should be ternary and not Jinja2 format

PR #316, opened 2024-05-23
Change when release_version to be more readable

reihl:
I replaced the place I found with ansible.builtin.version

Test-Hints: no-check

PR #327, opened 2024-05-30
Fix version validation as the pull request #316 broke the version che…

arnagaram:
…ck as the format is values such as 'latest-4.3' or 'latest-4.4'

---

- [x] OCP 4.12 (disconnected) -  https://www.distributed-ci.io/jobs/0b7fca44-7463-41c1-8eb1-ab7faffb5d47/jobStates
- [x] OCP 4.15 (disconnected) - https://www.distributed-ci.io/jobs/60f00124-3404-4e18-ad00-cb37a07838de
- [x] OCP 4.15 (connected) - https://www.distributed-ci.io/jobs/494e1bae-55f9-4f3e-9a91-38e91ecdf0a1/jobStates

---

Test-Hints: no-check

### DIFF
--- a/roles/node_prep/tasks/10_validation.yml
+++ b/roles/node_prep/tasks/10_validation.yml
@@ -295,37 +295,42 @@
   tags: validation
 
 - name: Validations for IPv6 VIPs in OCP >= 4.12
+  when:
+    - release_version is ansible.builtin.version('4.12', '>=')
+    - ipv6_enabled | bool
+    - dualstack_baremetal | bool
+    - dualstack_vips | bool
   block:
     - name: Verify DNS records for Wildcard (Ingress) IPv6 VIP
-      set_fact:
-        ingressvip6: "{{ lookup('dig', 'foo.apps.{{ cluster |quote }}.{{ domain | quote }}.', 'qtype=AAAA' ) }}"
+      ansible.builtin.set_fact:
+        ingressvip6: "{{ lookup('dig', 'foo.apps.{{ cluster |quote }}.{{ domain | quote }}.', 'qtype=AAAA' ) }}"  # noqa: redhat-ci[no-role-prefix]
       when: ((ingressvip6 is undefined) or (ingressvip6|length == 0))
       tags:
         - always
         - validation
 
     - name: Verify DNS records for API IPv6 VIP
-      set_fact:
-        apivip6: "{{ lookup('dig', 'api.{{ cluster |quote }}.{{ domain | quote }}.', 'qtype=AAAA' ) }}"
+      ansible.builtin.set_fact:
+        apivip6: "{{ lookup('dig', 'api.{{ cluster |quote }}.{{ domain | quote }}.', 'qtype=AAAA' ) }}"  # noqa: redhat-ci[no-role-prefix]
       when: ((apivip6 is undefined) or (apivip6|length == 0))
       tags:
         - always
         - validation
 
     - name: Display API IPv6 VIP
-      debug:
+      ansible.builtin.debug:
         msg: "The API IPv6 VIP is {{ apivip6 }}"
         verbosity: 2
       tags: validation
 
     - name: Display Ingress IPv6 VIP
-      debug:
+      ansible.builtin.debug:
         msg: "The Wildcard (Ingress) IPv6 VIP is {{ ingressvip6 }}"
         verbosity: 2
       tags: validation
 
     - name: Fail if incorrect API IPv6 VIP
-      fail:
+      ansible.builtin.fail:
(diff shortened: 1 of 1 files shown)


roles/node_prep/tasks/10_validation.yml
```diff
@@ -184,7 +184,7 @@
         - always
         - validation
   when:
-    - version is ansible.builtin.version('4.12', '>=')
+    - version.split('-')[-1] is ansible.builtin.version('4.12', '>=')
```
tonyskapunk, line 187: I'm afraid that this won't work for non-GA builds, e.g.

- Nightly builds, e.g. `4.16.0-0.nightly-2024-05-23-173505`
- RC builds, e.g. `4.16.0-rc.2`
- EC builds, e.g. `4.16.0-ec.6`

tonyskapunk (commented):
@arnagaram thanks for your contribution and bringing awareness of this.  Do you know if this is  the only instance that requires this change? #316 had more places where the conditionals were modified and I'm afraid we will need to revert it to have back the support to version formats as you mentioned which include: latest-4.Y, stable-4.Y, etc.

roles/node_prep/tasks/10_validation.yml
arnagaram: So it should not be an issue as of the task `Ge` Release.txt File` should fix the rest of issues. This is the only place where the version provided in the playbook is actually used for version checking. Based on the defined format of var `verrion` it should be fine.

arnagaram: Without this fix we will have the following issue:

`The error was: Version comparison failed: '<' not supported between instances of 'str' and 'int'\`
 
If the value of the version is set as `latest-4.14` for example

tonyskapunk: In that case, can we move the validation block *after* setting `release_version`? that way we keep the validation always against the `release_version` and not the requested `version`. That is moving:

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/e0977416b11102815693f263819c6c3dc40a63c9/roles/node_prep/tasks/10_validation.yml#L141-L190

To the current line 347:

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/e0977416b11102815693f263819c6c3dc40a63c9/roles/node_prep/tasks/10_validation.yml#L346-L348

arnagaram: Yeah we could also do that and always use `release_version` in every validation. It's just that I do not fully know the whole context of if its placed down there for a reason or its just a left over for refactoring.

tonyskapunk: It should be OK to move that validation after, I don't know either why that validation was placed in there, I can only assume because the task before that was also API VIP related, but I see other validations after the set of `relase_version` so I don't expect any issue with that. We will perform tests on it for the case of explicit  versions and for 4.12 and above.

arnagaram: I have now made the changes to always use `release_version` to validate.

tonyskapunk: Thanks, I'm performing a few tests with explicit versions. It will take some time. If you could validate the "latest-4.Y" form in the meantime, it would be great.

tonyskapunk:
@arnagaram there were some [issues caught by the CI for ansible-lint, as shown here](https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/9303772144?pr=327#summary-25608058211)

I've fixed them in my last commit, I'll continue to run the tests.

tonyskapunk:
@arnagaram I also see your commit is not signed, it is a [requirement to sign commits](https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/docs/dev.md#sign-commits) PLMK if there is any trouble with it.

tonyskapunk:
retest

tonyskapunk:
retest

tonyskapunk:
While the last testing job has not finished, I see it has passed the code changed in this PR without issues: https://www.distributed-ci.io/jobs/60f00124-3404-4e18-ad00-cb37a07838de?sort=date.

I'm moving it to merge when ready, but it will be pushed back as the first commit is not signed.

@arnagaram please follow the instructions to sign the commit so we can merge this change. Thanks again for your contribution!

tonyskapunk (approved):

tonyskapunk:
@arnagaram thanks for signing your commit, I see my commit was dropped, I'm cherry-picking it on top of your branch to have the ansible-lint fixed.

tonyskapunk (approved):
LGTM

tonyskapunk:
recheck

PR #331, opened 2024-06-03
Move to CNF Cert Suite v5.1.1, including new way of launching the certsuite

ramperher:
Test-Hints: no-check

---------

- Move to tnf v5.1.1 as default.
- New certsuite tool is proposed to launch the tnf container. Moving to that approach for this version - https://github.com/test-network-function/cnf-certification-test/blob/main/cmd/certsuite/run/run.go
- Keeping the old way of launching the cert suite as backwards compatibility, to be removed at some point.
- Setting common steps for both ways of launching the certification suite (create config/result dir, put the files in the same place...)

---------

Tests:

- [x] tnf-test-cnf-green - unstable - passed - https://www.distributed-ci.io/jobs/3ef38548-2a14-447c-9065-c0bf5a85d24c/jobStates?sort=date
- [x] tnf-test-cnf - v5.1.1 - passed - https://www.distributed-ci.io/jobs/c4650a1c-736b-4eeb-a68d-8b212866f12c/jobStates?sort=date
- [x] tnf-test-cnf - v5.1.0 - passed - https://www.distributed-ci.io/jobs/1b1e860b-0726-4b54-aa09-6303133466a0/jobStates?sort=date


ramperher:
Change is ready for review but I'll hold on the change till verifying what's happening with DeployableByOLM preflight test launched by tnf, which is now failing.

roles/cnf_cert/tasks/logging.yml
```diff
@@ -67,8 +67,15 @@
...
+  vars:
+    log_file_name: >
+      {% if test_network_function_version is version("v5.1.0", "<=") and test_network_function_version != "HEAD" %}
+        dci-tnf-execution.log
+      {% else %}
+        certsuite.log
+      {% endif %}
   ansible.builtin.lineinfile:
-    path: "{{ tnf_logs_location }}/dci-tnf-execution.log"
+    path: "{{ tnf_logs_location }}/{{ log_file_name | trim }}"
```
manurodriguez, line 78: LGTM, just to confirm. looks like in TNF v5.1.1 we only have `certsuite.log` log right? in the job with v5.1.0 version I see both, the `certsuite.log` and `dci-tnf-execution.log`

manurodriguez (dismissed):

roles/cnf_cert/tasks/logging.yml
ramperher: Yes, it was renamed at some point to `certsuite.log`, so, now that for v5.1.1 I'm proposing to directly use the container with `podman run`, then I've omitted the capture of logs in the file called `dci-tnf-execution.log`, that's why this file is not generated in 5.1.1, and with `certsuite.log`, it should be enough to check what we need to check in this task.
Also, there are some conditionals in the code to ensure backwards compatibility. When all consumers of this role move to, at least, 5.1.1, I'll start removing the support of the old code, as I did some months ago with 4.x.x.

roles/cnf_cert/tasks/logging.yml
```diff
@@ -67,8 +67,15 @@
 # this feature is enabled (disabled by default).
 # This check can only be done if CNF Cert was executed correctly (so claim.json file exists).
 - name: Confirm we are testing the correct CNF Cert code
+  vars:
+    log_file_name: >
+      {% if test_network_function_version is version("v5.1.0", "<=") and test_network_function_version != "HEAD" %}
```
reihl, line 72: Can we switch it to ternary?
"{{ (test_network_function_version is version("v5.1.0", "<=") and test_network_function_version != "HEAD") | ternary('dci-tnf-execution.log', 'certsuite.log')}}"

roles/cnf_cert/tasks/tests.yml
```diff
@@ -11,37 +23,85 @@
...
 
+# For v5.1.1 in advance, keep using some environment variables to
+# keep the definition of tnf_env_vars, but would be deprecated at some point.
+- name: Run the CNF Test Suite (tnf >= v5.1.1 and HEAD)
+  when: test_network_function_version is version("v5.1.1", ">=") or test_network_function_version == "HEAD"
+  vars:
+    tnf_partner_repo: "{{ (dci_local_registry | length) | ternary(dci_local_registry, 'quay.io') }}/testnetworkfunction"
+    support_image: "debug-partner:{{ support_image_version }}"
+  environment: "{{ tnf_env_vars }}"
+  ansible.builtin.shell: |
```
reihl, line 75: Why not use the podman module?

ramperher: Could be used, but since running this podman container requires a bunch of arguments and parameter, I preferred to use the shell module as a first step, similarly to what we do with [preflight](https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/preflight/tasks/test_preflight_check_container_one_image.yml#L70-L96).
Let me finish the testing and I can try to explore that. Could be a good improvement here.

roles/cnf_cert/tasks/logging.yml
ramperher: I started doing that in that way because I planned to use more than two values but finally I just used two alternatives, so yes, it's fine for me to move to ternary.

roles/cnf_cert/tasks/tests.yml
ramperher: @reihl , I tried here to use `podman` module: https://www.distributed-ci.io/jobs/2143a51b-307a-4fbe-8073-9f00b4b5c8ca/jobStates, but it's a bit tricky right now to do this transition. The execution doesn't like to work right now by using just exactly the same arguments I'm using with `shell`, and also, this would require a loop task to check the status of the container to confirm it's working fine.
I think it's better to keep this as it is right now, just to follow the trend that we already follow with preflight, as I said before, but since this role would require a refactor at some point, I'll take care of addressing that point in that moment. Is it fine for you?

reihl: Yes, we tried and face problems that we can avoid when we using shell. I just want to put this modules in our consideration. 
Having said that, can we consider using set_fact and convert the container command to a variable?

roles/cnf_cert/tasks/logging.yml
ramperher: I've finally kept dci-tnf-execution.log file for tnf v5.1.1 too, so I've removed this new code and I'll keep using the old one.

roles/cnf_cert/tasks/logging.yml
ramperher: I've finally kept dci-tnf-execution.log file for tnf v5.1.1 too, so I've removed this new code and I'll keep using the old one.

roles/cnf_cert/tasks/tests.yml
ramperher: Sure, I'll have the usage of podman module into consideration when addressing the next refactor of this role. I think it will be benefitial, but as I said, as this new way is also a fix for a current problem, and the integration of podman module would require to add more complexity to the role, I'd prefer to postpone this.

Regarding using set_fact, sure, I'll define the command in a variable (mainly with `vars`, as it's only used in this task) and give a try.

roles/cnf_cert/tasks/tests.yml
```diff
@@ -11,37 +23,94 @@
...
     {% if tnf_dockercfg_path is defined %}-c {{ tnf_dockercfg_path }} \{% endif %}
-    -i {{ tnf_image }} &> {{ test_network_function_project_name }}/dci-tnf-execution.log
+    -i {{ tnf_image }} &> result_files/dci-tnf-execution.log
   args:
     chdir: "{{ tnf_dir }}"
   ignore_errors: "{{ tnf_exec_ignore_errors | bool }}"
 
+# For v5.1.1 in advance, keep using some environment variables to
+# keep the definition of tnf_env_vars, but would be deprecated at some point.
+- name: Cpmmon tasks for CNF Cert Suite execution (tnf >= v5.1.1 and HEAD)
```
capolrik, line 69: *common

roles/cnf_cert/tasks/tests.yml
```diff
@@ -11,37 +23,94 @@
...
+        --network host \
+        -v {{ tnf_dir }}/config_files:/usr/tnf/config:Z \
+        -v {{ tnf_dir }}/result_files:/usr/tnf/results:Z \
+        {{ tnf_image }} \
+        {{ container_command }}
+      register: tnf_output
+      no_log: true
+
+    - name: Save output in a file
+      ansible.builtin.copy:
```
capolrik, line 110: If you are using the `no_log` option above, you may want to use it here as well.

roles/cnf_cert/defaults/main.yml
```diff
@@ -2,7 +2,7 @@
 tnf_stage: "tests"
 tnf_check_commit_sha: false
 test_network_function_repo: "https://github.com/test-network-function/cnf-certification-test"
-test_network_function_version: v5.1.0
+test_network_function_version: v5.1.1
 test_network_function_project_name: cnf-certification-test
```
capolrik, line 6: Is this variable still used?

ramperher: Yes, if no `test_network_function_version` is provided in the pipeline, it's used this one by default: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/cnf_cert/tasks/pre-run.yml#L152-L169, which is what happens with `tnf-test-cnf` pipeline

roles/cnf_cert/tasks/tests.yml
ramperher: Not really required because the output cannot be seen in this task but I've done it just in case, thanks!

roles/cnf_cert/tasks/tests.yml
ramperher: fixed

tkrishtop (approved):
LGTM, thanks @ramperher

roles/cnf_cert/tasks/tests.yml
ramperher: This is done, now the command is provided in a separate variable.

ramperher:
Tests are finished and all feedback has been integrated, I'll wait for the `local/check` to be fixed by DCI team before merging, also waiting for some more comments and feedback.

roles/cnf_cert/tasks/tests.yml
```diff
@@ -11,37 +23,96 @@
...
-    tnf_exec_ignore_errors: "{{ (test_network_function_version is version('v5.0.4', '>=') or test_network_function_version == 'HEAD') | ternary('false', 'true') }}"
+    tnf_exec_ignore_errors: "{{ (test_network_function_version is version('v5.0.4', '>=')) | ternary('false', 'true') }}"
   environment: "{{ tnf_env_vars | combine({'TNF_PARTNER_REPO': tnf_partner_repo, 'TNF_IMAGE': tnf_image, 'SUPPORT_IMAGE': support_image}) }}"
   ansible.builtin.shell: |
     set -x
 
-    ./run-tnf-container.sh -k {{ kubeconfig_path }} \
-    -t {{ tnf_dir }}/{{ test_network_function_project_name }} \
-    -o {{ tnf_dir }}/{{ test_network_function_project_name }} \
+    ./run-tnf-container.sh -k {{ tnf_dir }}/config_files/kubeconfig \
```
tonyskapunk, line 57: Maybe like this since it's already inside `tnf_dir` ?
```suggestion
    ./run-tnf-container.sh -k config_files/kubeconfig \
```

roles/cnf_cert/tasks/tests.yml
```diff
@@ -11,37 +23,96 @@
...
+    -t {{ tnf_dir }}/config_files \
+    -o {{ tnf_dir }}/result_files \
```
tonyskapunk, line 59: Maybe?

```suggestion
    -t config_files \
    -o result_files \
```

tonyskapunk (approved):
LGTM

Left a couple of suggestions, don't forget to squash your changes before merging, thanks!

roles/cnf_cert/tasks/tests.yml
ramperher: I think I've been passing the arguments using the absolute path because I had issues with relative paths in the past. But I'll test this.

roles/cnf_cert/tasks/tests.yml
ramperher: Same, I'll test this to confirm it works fine.

roles/cnf_cert/tasks/tests.yml
ramperher: As commented in https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/331#issuecomment-2149295177, we cannot use relative paths for the arguments passed to `run-tnf-container`, so I will keep the code in the same status if you're fine with this.
I'll also keep `chdir: "{{ tnf_dir }}"` just to shortcut the call to `./run-tnf-container.sh` script, which is placed in that directory

roles/cnf_cert/tasks/tests.yml
ramperher: As commented in https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/331#issuecomment-2149295177, we cannot use relative paths for the arguments passed to `run-tnf-container`, so I will keep the code in the same status if you're fine with this.
I'll also keep `chdir: "{{ tnf_dir }}"` just to shortcut the call to `./run-tnf-container.sh` script, which is placed in that directory

ramperher:
Change is ready for review again, launching again the DCI jobs to check everything is fine. Could you take a look at this? @tkrishtop @tonyskapunk thanks!

tonyskapunk (approved):
Thanks for taking a look and testing the suggestion, :+1:

tonyskapunk:
recheck

PR #332, opened 2024-06-05
Produce IDMS for 4.14+

tonyskapunk:
ImageContentSourcePolicies is going to be deprecated, its replacement is ImageDigestMirrorSets, starting 4.14 the release mirror allows IDMS, switching to its consumption.

---
- [x] IPI OCP-4.14 - https://www.distributed-ci.io/jobs/9a95817f-0656-45e4-9edb-1b0c09241ec7 
- [x] Assisted OCP-4.14 - https://www.distributed-ci.io/jobs/42ab9a77-14d0-4ae4-9f50-b9161020f02f

---
Test-hints: no-check


ramperher (commented):
I've seen there are some more places where `oc adm release mirror` is used in the collections, are they also affected by this?

```
$ grep -r "oc adm release mirror"
roles/populate_mirror_registry/tasks/populate_registry.yml:          "{{ binaries_tool_path }}"/oc adm release mirror
roles/populate_mirror_registry/tasks/populate_registry.yml:          "{{ binaries_tool_path }}"/oc adm release mirror
roles/installer/tasks/15_disconnected_registry_create.yml:        /usr/local/bin/oc adm release mirror
roles/mirror_ocp_release/tasks/registry.yml:        {{ mor_cache_dir }}/{{ mor_version }}/oc adm release mirror
```

In fact, the code that is changed is under a rescue block, that was not really triggered in the last DCI job, so I don't know if there's a good way to test this.

roles/mirror_ocp_release/tasks/registry.yml
```diff
@@ -27,12 +27,15 @@
...
   rescue:
     - name: "Mirror release images to local registry"
-      ansible.builtin.command: >
+      ansible.builtin.command: >-
         {{ mor_cache_dir }}/{{ mor_version }}/oc adm release mirror
         --registry-config={{ mor_auths_file }}
         --from={{ ocp_release_data['container_image'] | quote }}
         --to-release-image={{ mor_registry_url }}/{{ mor_registry_path }}:{{ mor_version }}
         --to={{ mor_registry_url }}/{{ mor_registry_path }}
+        {% if mor_version is version("4.14", ">=") %}
```
manurodriguez, line 36: Thanks Tony, from our conversation about this yesterday, AI still uses `ImageContentSourcePolicy` in 4.14, so I wonder if this option would print both outputs of `ImageContentSourcePolicy` and `ImageDigestMirrorSet` so we can decide which one to use at the end?

betoredhat:
> I've seen there are some more places where `oc adm release mirror` is used in the collections, are they also affected by this?
> 
> ```
> $ grep -r "oc adm release mirror"
> roles/populate_mirror_registry/tasks/populate_registry.yml:          "{{ binaries_tool_path }}"/oc adm release mirror
> roles/populate_mirror_registry/tasks/populate_registry.yml:          "{{ binaries_tool_path }}"/oc adm release mirror
> roles/installer/tasks/15_disconnected_registry_create.yml:        /usr/local/bin/oc adm release mirror
> roles/mirror_ocp_release/tasks/registry.yml:        {{ mor_cache_dir }}/{{ mor_version }}/oc adm release mirror
> ```
> 
> In fact, the code that is changed is under a rescue block, that was not really triggered in the last DCI job, so I don't know if there's a good way to test this.

An option is to force the mirroring or remove the already cached image-sources files in the cache.

roles/mirror_ocp_release/tasks/registry.yml
tonyskapunk: Great observation, that would be a problem :crying_cat_face:  This change generates only the IDMS. A better approach is to make this role to have the option of using one or the other and only allowing IDMS for 4.14 and above. Let me change this.

roles/mirror_ocp_release/tasks/registry.yml
```diff
@@ -27,12 +27,15 @@
...
   rescue:
     - name: "Mirror release images to local registry"
-      ansible.builtin.command: >
+      ansible.builtin.command: >-
         {{ mor_cache_dir }}/{{ mor_version }}/oc adm release mirror
         --registry-config={{ mor_auths_file }}
         --from={{ ocp_release_data['container_image'] | quote }}
         --to-release-image={{ mor_registry_url }}/{{ mor_registry_path }}:{{ mor_version }}
         --to={{ mor_registry_url }}/{{ mor_registry_path }}
+        {% if mor_version is version("4.14", ">=") and mor_is_type | lower == "idms" %}
```
reihl, line 36: Can it be switch to ternary?

betoredhat:
recheck

roles/mirror_ocp_release/tasks/registry.yml
tonyskapunk: This change should address it, that is IPI (4.14+) Will request IDMS (https://softwarefactory-project.io/r/c/dci-openshift-agent/+/31736), while ABI will not

tonyskapunk:
> I've seen there are some more places where `oc adm release mirror` is used in the collections, are they also affected by this?

This change is only for the` mirror_ocp_release` role, the rescue code you mentioned has been changed as in the past it was depending on one artifact (image source file) to determine whether it needed to mirror or not images, now an image is checked, mirroring is triggered if lacking AND *always* an image source is generated depending on the requested type IDMS or ICSP.

tonyskapunk:
recheck

roles/mirror_ocp_release/tasks/registry.yml
```diff
@@ -1,58 +1,57 @@
...
-        msg: "Forcing the release mirroring"
-      when:
-        - mor_force | bool
+- name: Grab generated Image Source manifest
+  ansible.builtin.shell: >
+    set -o pipefail;
+    echo -e "{{ _mor_result.stdout }}" |
+    sed -n '/apiVersion/,$p' |
+    sed -e 's/name:.*/name: release-{{ mor_version }}/'
+  register: is_release
```
reihl, line 46: I think there is better option to get the information from that variable without using shell

tonyskapunk: No doubt, but wonder if is worth the effort. I remember you shared a nifty module that would help with parsing, but the downside is that it requires installing a library thus adding a dependency on the package. I'll continue as is, since is clear and works for both cases IDMS and ICSP. Thanks

PR #335, opened 2024-06-07
dci-merge | Remove \r characters from PR description

ramperher:
Test-Hints: no-check

Despite https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/334, it's still not working fine because \r characters are not removed correctly.

### DIFF
--- a/hack/dci-merge.sh
+++ b/hack/dci-merge.sh
@@ -69,19 +69,22 @@ for PR in $PRS; do
         fi
 
         # extract TestBos2 commands
+        # Remove \r character if present
         if grep -qE "^\s*TestBos2:\s*" <<< "$DESC"; then
             # shellcheck disable=SC2001,SC2086
-            CMD+=("$(sed -ne 's/^\s*TestBos2:\s*//p' <<< $DESC)")
+            CMD+=("$(sed -ne 's/^\s*TestBos2:\s*//p' <<< $DESC | sed 's/\r$//')")
         fi
         # extract TestBos2Sno commands
+        # Remove \r character if present
         if grep -qE "^\s*TestBos2Sno:\s*" <<< "$DESC"; then
             # shellcheck disable=SC2001,SC2086
-            CMD_SNO+=("$(sed -ne 's/^\s*TestBos2Sno:\s*//p' <<< $DESC)")
+            CMD_SNO+=("$(sed -ne 's/^\s*TestBos2Sno:\s*//p' <<< $DESC | sed 's/\r$//')")
         fi
         # extract TestBos2Baremetal commands
+        # Remove \r character if present
         if grep -qE "^\s*TestBos2Baremetal:\s*" <<< "$DESC"; then
             # shellcheck disable=SC2001,SC2086
-            CMD_SNO_BM+=("$(sed -ne 's/^\s*TestBos2Baremetal:\s*//p' <<< $DESC)")
+            CMD_SNO_BM+=("$(sed -ne 's/^\s*TestBos2Baremetal:\s*//p' <<< $DESC | sed 's/\r$//')")
         fi
     fi
 done


ramperher:
Tested locally:

```
$ cat test-dci-merge.sh
set -x

DESC=$(curl -s "${GH_HEADERS[@]/#/-H}" https://api.github.com/repos/redhatci/ansible-collection-redhatci-ocp/pulls/333|jq -r .body)

if grep -qE "^\s*TestBos2:\s*" <<< "$DESC"; then
    CMD="$(sed -ne 's/^\s*TestBos2:\s*//p' <<< $DESC)"
    CMD="$(sed 's/\r$//' <<< $CMD)"
fi

echo $CMD

$ ./test-dci-merge.sh                                                                                                                                                         
+++ curl -s https://api.github.com/repos/redhatci/ansible-collection-redhatci-ocp/pulls/333
+++ jq -r .body
++ DESC='TestBos2: virt

Some of them was bug some of them change to be more ansible format and some of them changed to modules'
++ grep -qE '^\s*TestBos2:\s*'
+++ sed -ne 's/^\s*TestBos2:\s*//p'
++ CMD=$'virt\r'
+++ sed 's/\r$//'
++ CMD=virt
++ echo virt
virt

```

Now \r is removed correctly

@fredericlepied pinging you here because maybe there's a more elegant way of doing this but didn't find the way after thinking some time.

hack/dci-merge.sh
```diff
@@ -71,17 +71,23 @@ for PR in $PRS; do
         # extract TestBos2 commands
         if grep -qE "^\s*TestBos2:\s*" <<< "$DESC"; then
             # shellcheck disable=SC2001,SC2086
-            CMD+=("$(sed -ne 's/^\s*TestBos2:\s*//p' <<< $DESC)")
```
fredericlepied, line 74: why not `sed -ne 's/^\s*TestBos2:\s*//p; s/\r$//'` ?

fredericlepied (changes requested):

hack/dci-merge.sh
ramperher: @fredericlepied this is the only way I've found to do this in one single line: `sed -ne 's/^\s*TestBos2:\s*//p' <<< $DESC | sed 's/\r$//'`
Using other options like yours or `sed -ne 's/^\s*TestBos2:\s*//p' <<< $DESC | sed 's/\r$//'` are not working

manurodriguez: Interesting, at first glance both options seem to return the same but I see the option from Fred is rendering oddly:

```
[dciteam@jumphost ~]$ DESC=$(curl -s https://api.github.com/repos/redhatci/ansible-collection-redhatci-ocp/pulls/333 | jq -r .body)                          
[dciteam@jumphost ~]$ echo "$DESC"
TestBos2: virt

Some of them was bug some of them change to be more ansible format and some of them changed to modules
                                                   
[dciteam@jumphost ~]$ ONE=$(sed -ne 's/^\s*TestBos2:\s*//p' <<< $DESC | sed 's/\r$//')
[dciteam@jumphost ~]$ TWO=$(sed -ne 's/^\s*TestBos2:\s*//p' -e 's/\r$//' <<< $DESC)
[dciteam@jumphost ~]$ echo ${ONE}
virt
[dciteam@jumphost ~]$ echo ${TWO}
virt
[dciteam@jumphost ~]$ echo ${ONE}manny
virtmanny
[dciteam@jumphost ~]$ echo ${TWO}manny
manny
```

manurodriguez (approved):
LGTM

hack/dci-merge.sh
ramperher: Yes, and apart from this, it's still embedding the `\r` character, which is the reason why it's not working afterwards when trying to schedule a pipeline with that code.
I've refactored my code to do it in one line. Since there are no more reviews with alternatives to improve this, and don't know if @fredericlepied will be able to check this during this week since he's off, I'll propose to merge this and then improve the code if needed in a separate change.

ramperher:
I've refactored my code to do it in one line. Since there are no more reviews with alternatives to improve this, and don't know if @fredericlepied will be able to check this during this week since he's off, I'll propose to merge this and then improve the code if needed in a separate change.

manurodriguez (approved):
Thanks Ramon, I also believe we can go with this for now, and we can see with Fred when he is back if needed

PR #341, opened 2024-06-11
Use well-prefixed vars to call configure_ztp_gitops_apps and update docs

ramperher:
build-depends: 31703
Test-Hints: no-check

---

This is to test the whole workflow, hub +  spoke, then taking care of doing some pending refactoring work in the code.


ramperher:
recheck

ramperher:
Change is working fine, I was able to launch hub and spoke cluster in the same call to dci-pipeline. This is ready.

manurodriguez (approved):
LGTM

PR #344, opened 2024-06-12
ZTP | Create OCP pull secret from hub cluster, refactor of some code

ramperher:
Test-Hints: no-check

----

- Create OCP pull secret from hub cluster (currently it's not created, we rely on what's provided in the git repo, but we would need to update it if the hub cluster changes)
- Refactor of some code: `ssh_known_hosts` var in ConfigMap.
- Properly update policies-app-project.yaml file

I will not launch automatic tests for this, I'll run them manually since the scenario to test this is a bit tricky.

### DIFF
--- a/roles/configure_ztp_gitops_apps/tasks/main.yaml
+++ b/roles/configure_ztp_gitops_apps/tasks/main.yaml
@@ -101,13 +101,15 @@
         replace: "{{ czga_policies_namespace }}"
 
     - name: Add namespace to the policies-app-project.yaml
-      ansible.builtin.lineinfile:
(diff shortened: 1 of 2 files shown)


ramperher:
This is ready for review. @nsilla , @betoredhat , is it fine for you if, for the moment, we take the hub pull secret and create a new pull secret for the spoke cluster based on it, while we explore other alternatives? Because, if not, pipelines like Bevo's ones would not work, because we need to extract this pullsecret somehow.

nsilla (approved):
lgtm.
Thanks for the work

PR #349, opened 2024-06-13
ACM - No log for listing CSV

betoredhat:
Logging all CVS data from a cluster that has multiple operators installed will generate a big ansible log, so disabling it.

TestBos2: virt-prega


manurodriguez (approved):
LGTM

PR #351, opened 2024-06-16
Change variable vbmc_host to ansible_host to get the host that playbo…

reihl:
TestBos2: virt
…ok run against this need to run the playbook on ansible navigator container


reihl:
retest

reihl:
Found the way to bypass it

PR #353, opened 2024-06-18
Generate IDMS instruction if OCP >= 4.14

betoredhat:
The stable version supports printing mirroring instructions

Test-hints: no-check

---

- [x] OCP 4.12 - https://www.distributed-ci.io/jobs/f24a3ad8-494e-4642-b35f-52dee78624ec?sort=date&task=85b54fb1-ad50-4289-b5a7-845b37d83842
- [x]  OCP 4.14 - https://www.distributed-ci.io/jobs/2886cf52-81a4-40f9-a252-27486d6b8c8f/jobStates?sort=date&task=b377b129-b667-4124-a4d0-1faac2323620

PR #354, opened 2024-06-18
ignoring the --print-mirror-instructions for versions below 4.14

capolrik:
The option in oc adm mirror is not available before 4.14


capolrik:
I did not see https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/353 let me close this one

manurodriguez:
> I did not see #353 let me close this one

I was ok with your approach to solve this, I just saw that Beto started the other PR yesterday, we have been using the stable oc in other roles and parts of the agents, so far we haven't seen an issue yet [knocking wood : )]

PR #360, opened 2024-06-26
Add misc playbooks

betoredhat:
- Some utility playbook
- Readme updates

Test-Hints: no-check


playbooks/install_operators.yml
```diff
@@ -0,0 +1,44 @@
...
+        operator: ocs-operator
+        source: "{{ catalog_image }}"
+        namespace: redhat-operators
+        ns_labels:
+          openshift.io/cluster-monitoring: "true"
+        operator_group_spec:
+          targetNamespaces:
+            - openshift-storage
+
+    - name: Install OCS Operator
```
manurodriguez, line 34: I believe here should read `Install ACM Operator`?

playbooks/mirror_ocp.yml
```diff
@@ -0,0 +1,24 @@
+# A playbook that mirror an OCP release. The install artifacts are mirred to the
```
manurodriguez, line 1: I think there is a typo here: 
```suggestion
# A playbook that mirror an OCP release. The install artifacts are mirrored to the
```

playbooks/multibench_setup_host/README.md
```diff
@@ -0,0 +1,22 @@
+# Multi-bench playbook
+
+## Variables
+
+| Variable                  | Default                        | Type    | Required | Description                                                                          |
+|---------------------------|--------------------------------|---------|----------|--------------------------------------------------------------------------------------|
+| multibench_quay_token     | null                           | string  | true     | Path to the file which contains the credentials for accessing the container registry |
+| multibench_disconnected   | false                          | boolean | false    | Set it to 'true' if the multibench host does not have access to the Internet         |
+| multibench_local_registry | quay.io/crucible/client-server | String  | false    | Registry that will be used to pull the crucible images                               |
+| multibench_git_name       | Smith                          | String  | false    | Name displayed in on git                                                             |
```
manurodriguez, line 10: in `Name displayed in on git` maybe just saying `Name displayed in git`

playbooks/multibench_setup_host/README.md
```diff
@@ -0,0 +1,22 @@
...
+| multibench_git_email      | ansible@whatever.com           | String  | false    | email displayed in on git                                                            |
```
manurodriguez, line 11: Same here: `email displayed in git`

manurodriguez (commented):
Thanks Beto, minor comments from my side

ramperher (commented):
should we document the usage of these new playbooks? I mean, for example in install_operators, users need to provide the local registry and the port. Just to let them know that they have to include some inputs.

betoredhat:
> should we document the usage of these new playbooks? I mean, for example in install_operators, users need to provide the local registry and the port. Just to let them know that they have to include some inputs.

Thank Ramon, I added a not about replacing the placeholder vars. Thanks

tonyskapunk:
@betoredhat thanks for the change, is the intention have a place to exemplify the use of the roles through playbooks?
Or is the intention to provide these playbooks to be consumable by someone/something?

The reason of my question is because providing playbooks under `./playbooks` actually get them available in the collection, as collections not only include roles and plugins it also includes playbooks, so this change will include them in galaxy.

Looking at the majority of them it looks like they are meant to be examples. In that case I think the best place to have them is inside their own role, but I wanted to be sure the intention on this change first.

betoredhat:
> @betoredhat thanks for the change, is the intention have a place to exemplify the use of the roles through playbooks? Or is the intention to provide these playbooks to be consumable by someone/something?
> 
> The reason of my question is because providing playbooks under `./playbooks` actually get them available in the collection, as collections not only include roles and plugins it also includes playbooks, so this change will include them in galaxy.
> 
> Looking at the majority of them it looks like they are meant to be examples. In that case I think the best place to have them is inside their own role, but I wanted to be sure the intention on this change first.

@tonyskapunk The idea was to make them available for others and distribute them via the collection. We have had discussions about how to deliver specific task recipes using the roles without depending on the agent. However, I'm not sure if having users add additional inputs aligns with the purpose of the playbook section.

playbooks/install_operators.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Create a CatalogSource
+      ansible.builtin.include_role:
+        name: redhatci.ocp.catalog_source
+      vars:
+        cs_name: "{{ catalog_name }}"
+        cs_namespace: "{{ source_ns }}"
+        cs_image: "{{ catalog_image }}"
+        cs_publisher: "The publisher"
+
+    - name: Install OCS Operator
```
tonyskapunk, line 22: The problem I see with this playbook is that is very specific, it will only install OCS and ACM. I think it would be more useful being more generic, wdyt?

tonyskapunk (commented):
> The idea was to make them available for others and distribute them via the collection

Understood, then they need some changes to be able to be called in the expected way.

For their consumption, you could add a new playbooks section in the main README.md and include an example of its use like:

```yaml
- name: Installing Operators
  ansible.builtin.import_playbook: redhatci.ocp.prune_mirror
  vars:
    docker_config: /home/<user>/.docker/
    auths_file: "{{ docker_config }}/config.json"
    catalog_source: "registry.redhat.io/redhat/redhat-operator-index:v4.15"
    local_registry: "<local_registry>:<local_registry_port>"
    pruned_catalog: "{{ local_registry }}/<namespace>/<my_catalog>:latest"
    opm_auths: /home/<user>/opm-auths
    oc_bin_path: /usr/bin/oc
    prune_operators:
      - sriov-network-operator
```

betoredhat:
I will hold on until we get an agreement if this is needed and adds value.

PR #361, opened 2024-06-26
Add support for insecure registries

betoredhat:
- Allow interaction with registries using an unknown CA. Default is set as already in place in other roles.
- In catalog_prune role, allowing to override the opm_auths wrapper.

TestDallas: ocp-4.14-vanilla

### DIFF
--- a/roles/mirror_ocp_release/tasks/registry.yml
+++ b/roles/mirror_ocp_release/tasks/registry.yml
@@ -4,6 +4,9 @@
     skopeo inspect
     --no-tags
     --authfile {{ mor_auths_file }}
+    {%- if mor_allow_insecure_registry | bool %}
+    --tls-verify=false
+    {%- endif %}
     docker://{{ mor_registry_url }}/{{ mor_registry_path }}:{{ mor_version }}
   register: _mor_release_image
   failed_when: false
@@ -16,12 +19,19 @@
     --from={{ ocp_release_data['container_image'] | quote }}
     --to-release-image={{ mor_registry_url }}/{{ mor_registry_path }}:{{ mor_version }}
     --to={{ mor_registry_url }}/{{ mor_registry_path }}
+    {%- if mor_allow_insecure_registry | bool %}
+    --insecure
+    {%- endif %}
   retries: 3
   delay: 10
   register: _mor_result
   until: _mor_result.rc == 0
   changed_when: _mor_result.rc == 0
-  when: _mor_release_image.rc != 0 or mor_force | bool
+  when:
+    - _mor_release_image is defined
+    - _mor_release_image.rc is defined
+    - _mor_release_image.rc != 0
+      or mor_force | bool
 
 - name: Generate Image Source manifest
   ansible.builtin.command: >-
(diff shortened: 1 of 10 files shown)


roles/mirror_ocp_release/tasks/registry.yml
```diff
@@ -16,12 +19,19 @@
...
+    {%- if mor_allow_insecure_registry | bool %}
+    --insecure
+    {%- endif %}
   retries: 3
   delay: 10
   register: _mor_result
   until: _mor_result.rc == 0
   changed_when: _mor_result.rc == 0
-  when: _mor_release_image.rc != 0 or mor_force | bool
+  when:
```
ramperher, line 30: Shouldn't this be something like...?

```
when: (_mor_release_image is defined and _mor_release_image.rc is defined and _mor_release_image.rc != 0) or mor_force | bool
```

Because it's a bit strange for me that we're mixing `_mor_release_image and mor_force in the same check when we have more extra checks for only `_mor_release_image`

roles/prune_catalog/defaults/main.yml
```diff
@@ -5,4 +5,5 @@ pc_expire_time: 5h
 pc_opm_auths: "/usr/share/dci-openshift-agent/utils/opm-auths"
 pc_maintainer: "redhatci.ocp"
 pc_ignore_pull_errors: false
+pc_allow_insecure_registry: true
```
ramperher, line 8: to be included in the README as for the other roles?

roles/mirror_ocp_release/tasks/registry.yml
betoredhat: That looks good, will add it in next patch. Thanks

roles/prune_catalog/tasks/main.yml
```diff
@@ -11,16 +11,18 @@
...
+    docker_config_defined: "{{ DOCKER_CONFIG is defined }}"
+  ansible.builtin.command: >
+    skopeo inspect --no-tags
+    {%- if pc_allow_insecure_registry | bool %}
+    --tls-verify=false
+    {%- endif %}
+    {%- if docker_config_defined | bool %}
+    --authfile "$DOCKER_CONFIG/config.json"
+    {%- endif %}
+    docker://{{ pc_source_catalog }}
```
tonyskapunk, line 25: I'm not sure if `DOCKER_CONFIG` could be evaluated like that since is an `env` and not an ansible var.

I think this should do it:

```suggestion
- name: "Check if image is pullable - {{ pc_source_catalog }}"
  ansible.builtin.command: >
    skopeo inspect --no-tags
    {%- if pc_allow_insecure_registry | bool %}
    --tls-verify=false
    {%- endif %}
    {%- if lookup('ansible.builtin.env', 'DOCKER_CONFIG') | length %}
    --authfile "$DOCKER_CONFIG/config.json"
    {%- endif %}
    docker://{{ pc_source_catalog }}
```

manurodriguez (approved):
LGTM

betoredhat:
retest

PR #356, opened 2024-06-20
cnf_cert | Updating certsuite call for HEAD version

ramperher:
TestDallasWorkload: tnf-test-cnf tnf-test-cnf:ansible_extravars=tests_to_verify:[]

- [x] tnf stable (launched here) - https://www.distributed-ci.io/jobs/e8affb28-1e4e-442d-81cc-a8bd5b288d0f/jobStates?sort=date
- [x] tnf unstable (launched from https://github.com/test-network-function/cnf-certification-test/pull/2181) - https://www.distributed-ci.io/jobs/a8ac738d-2c8d-4fe5-ab85-aa693bdba711/jobStates?sort=date

----

- Adapting the code to the new way of launching certsuite, included in https://github.com/test-network-function/cnf-certification-test/pull/2181
- Using an `always` clause to always retrieve the output from tnf exec
- Removed extra arguments when building the tnf container image for PR testing, they seem to be no longer needed

PR #362, opened 2024-06-27
Increase CNF certsuite to use v5.1.3

manurodriguez:
Newest release is [v5.1.3](https://github.com/test-network-function/cnf-certification-test/releases/tag/v5.1.3)

---

- [x] tnf-test-cnf-green (default) - https://www.distributed-ci.io/jobs/45c2b06f-c296-42cb-9dd6-e07275b70609/jobStates
- [x] tnf-test-cnf - https://www.distributed-ci.io/jobs/19bebe4d-ba1b-40bd-a1c2-9ba27b9bc8d8/jobStates (failed in platform-alteration-tainted-node-kernel test) - NEEDS review if expected.

---

Test-Hints: no-check


ramperher (changes requested):
The tnf-test-cnf job is failing because we need to move this code we added for `HEAD` branch to also support `v5.1.3` and higher: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/356

So, we need to update these two vars: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/cnf_cert/tasks/tests.yml#L76-L77

Where: `(test_network_function_version == 'HEAD')` must be now `(test_network_function_version is version("v5.1.3", ">=") or test_network_function_version == "HEAD")`

Consequences of supporting backwards compatibility :sweat_smile:

manurodriguez:
> The tnf-test-cnf job is failing because we need to move this code we added for `HEAD` branch to also support `v5.1.3` and higher: #356

@ramperher thanks a lot for checking, I added those changes too, relaunching tests.

rdavid (approved):

sebrandon1 (approved):

roles/cnf_cert/tasks/tests.yml
```diff
@@ -66,10 +66,10 @@
 
 # For v5.1.1 in advance, keep using some environment variables to
 # keep the definition of tnf_env_vars, but would be deprecated at some point.
-- name: Common tasks for CNF Cert Suite execution (tnf >= v5.1.1 and HEAD)
-  when: test_network_function_version is version("v5.1.1", ">=") or test_network_function_version == "HEAD"
+- name: Common tasks for CNF Cert Suite execution (tnf >= v5.1.3 and HEAD)
```
ramperher, line 69: This cannot be changed, else the flow will be broken for tnf v5.1.1 and v5.1.2.

What we only need to change is what I said in my prev comment:

_So, we need to update these two vars: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/cnf_cert/tasks/tests.yml#L76-L77_

_Where:_ `(test_network_function_version == 'HEAD')` _must be now_ `(test_network_function_version is version("v5.1.3", ">=") or test_network_function_version == "HEAD")`

So, we only need to modify these two vars, that's all

ramperher (changes requested):

roles/cnf_cert/tasks/tests.yml
manurodriguez: my bad, I think I got this time : (

ramperher (approved):
Yeah, now looks good :) last job failed because the tainted-node-kernel test is not expected to fail for tnf-test-cnf, but since cluster10 was used, we have taints there. It was expected.

ramperher:
recheck

manurodriguez:
> Yeah, now looks good :) last job failed because the tainted-node-kernel test is not expected to fail for tnf-test-cnf, but since cluster10 was used, we have taints there. It was expected.

Thanks a lot for checking Ramon. I thought that was the case, but I preferred to wait for your review. : )

PR #363, opened 2024-06-27
sriov_config | Add OCPBUGS-32139 workaround

ramperher:
We have started to see [this issue](https://issues.redhat.com/browse/OCPBUGS-32139) in some SRIOV-based deployments. This is not 100% reproducible but it is happening now from time to time. Example [here](https://www.distributed-ci.io/jobs/e7c9b138-ee53-41d5-80ce-d7ccaf801f6f/jobStates?sort=date&task=f6d1f3e3-77e4-4de9-815c-5f0e8e1c003a).

Proposed workaround is retry applying SRIOV policies. Just doing this.

Test-Hints: no-check

### DIFF
--- a/roles/sriov_config/tasks/create_node_policies.yml
+++ b/roles/sriov_config/tasks/create_node_policies.yml
@@ -1,4 +1,6 @@
 ---
+# OCPBUGS-32139 workaround suggests to retry the creation of the SRIOV policy if failing.
+# Trying this during 1 minute in case any of them fails.
 - name: Create SriovNetworkNodePolicy
   community.kubernetes.k8s:
     definition: "{{ lookup('template', 'templates/sriov-network-node-policy.yml.j2') }}"
@@ -6,6 +8,10 @@
   loop_control:
     loop_var: sriov
     label: "{{ sriov.resource }}"
+  retries: 6
+  delay: 10
+  register: _sc_node_policy_retry
+  until: _sc_node_policy_retry.error is not defined
   when: sriov.node_policy is defined
 
 - name: Check for SRIOV Node Policy


ramperher:
Working fine in this deployment: https://www.distributed-ci.io/jobs/2214b5c7-bd12-4bbd-9bc9-21d810aa9c87/jobStates?sort=date&task=e05ef767-e0b1-41f9-887f-d7327848ea8f, ready for review once it finishes.

roles/sriov_config/tasks/create_node_policies.yml
```diff
@@ -6,7 +6,28 @@
   loop_control:
     loop_var: sriov
     label: "{{ sriov.resource }}"
-  when: sriov.node_policy is defined
+  when:
+    - sriov.node_policy is defined
+    - '"OCPBUGS-32139" not in dci_workarounds|default([])'
```
manurodriguez, line 11: IMO, I'm ok with this, but I was wondering if simply we add retries in this task and remove the conditions and new task with the dci_workarounds. Let's hear other opinions.

betoredhat: yeah, I would prefer to avoid using `dci_`  vars in the roles. May need to create a new playbook to look through the resources and wait for it it be created.

betoredhat: Anyway, retries are the secret sauce of most of the issues : )

manurodriguez: I know Ramon want's to do things right, but I do not want to image the number of work-arounds we would have in our code if we'll add them for every retry we have added in our plays :see_no_evil:

tonyskapunk: I agree with both, @ramperher could you remove the dci_workarounds in here? It won't hurt adding some retries into the creation of the resource.  It will simplify the task and we won't need to remove it later, win-win. Also I have strong feelings about making these roles dependent of DCI unless necessary.

ramperher:
recheck

roles/sriov_config/tasks/create_node_policies.yml
ramperher: It's updated now, could you take a look again? @manurodriguez @betoredhat @tonyskapunk

roles/sriov_config/tasks/create_node_policies.yml
```diff
@@ -1,11 +1,17 @@
...
+# Trying this during 1 minute in case any of them fails.
 - name: Create SriovNetworkNodePolicy
   community.kubernetes.k8s:
     definition: "{{ lookup('template', 'templates/sriov-network-node-policy.yml.j2') }}"
   loop: "{{ sriov_network_configs }}"
   loop_control:
     loop_var: sriov
     label: "{{ sriov.resource }}"
+  retries: 6
+  delay: 1
```
tkrishtop, line 12: Hi @ramperher I believe like this it's a 'one second' delay, not 'one minute'. To have 1 minute, you can go with
```
 delay: 60
```

ramperher: Oh, sorry, I deleted a 0 in my last refactor of the change, I wanted 6 retries with 10 seconds delay.

roles/sriov_config/tasks/create_node_policies.yml
manurodriguez: @ramperher, thanks a lot, for me a work-around it's worth adding when it involves extra tasks (restart a pod, apply a config, etc) but if it's just a retry we could afford not to add it, an in favour of your argument, there is a BUG reported, so that does not happens all the time, so it makes sense to reference it in the code, but let's discuss this in a a grooming or daily meeting to have a touch point, it's been a while we do not talk about work-arounds : )

manurodriguez (approved):
Thanks!

roles/sriov_config/tasks/create_node_policies.yml
ramperher: sure, let's do that :) closing this, thanks!

PR #364, opened 2024-06-28
Select RHEL version for oc stable binary in SNO and UPI installers

ramperher:
Since OCP 4.16 became GA, oc stable binary is now based on that release and it is expecting GLIBC library to be present, which is only delivered in RHEL9.

To keep compatibility with RHEL8 servers, using stable binary but indicating the RHEL version to use.

TestBos2Sno: sno

### DIFF
--- a/roles/sno_installer/tasks/10_get_oc.yml
+++ b/roles/sno_installer/tasks/10_get_oc.yml
@@ -101,7 +101,7 @@
   vars:
     ocp_stable_url: https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp
   ansible.builtin.unarchive:
-    src: "{{ ocp_stable_url }}/stable/openshift-client-linux.tar.gz"
+    src: "{{ ocp_stable_url }}/stable/openshift-client-linux-amd64-rhel{{ ansible_distribution_major_version }}.tar.gz"
     dest: "{{ sno_tmp_dir.path }}"
     remote_src: true
     mode: "0755"
--- a/roles/upi_installer/tasks/10_get_oc.yml
+++ b/roles/upi_installer/tasks/10_get_oc.yml
@@ -99,7 +99,7 @@
   vars:
     ocp_stable_url: https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp
   ansible.builtin.unarchive:
-    src: "{{ ocp_stable_url }}/stable/openshift-client-linux.tar.gz"
+    src: "{{ ocp_stable_url }}/stable/openshift-client-linux-amd64-rhel{{ ansible_distribution_major_version }}.tar.gz"
     dest: "{{ upi_tmp_dir.path }}"
     remote_src: true
     mode: "0755"


ramperher:
Failing task is now passing: https://www.distributed-ci.io/jobs/9c7db3f0-65fd-4a56-a4a1-680c04930015/jobStates?sort=date&task=cd5d86ce-e87b-4c4d-8a21-59b2420e7c37

roles/sno_installer/tasks/10_get_oc.yml
```diff
@@ -101,7 +101,7 @@
   vars:
     ocp_stable_url: https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp
   ansible.builtin.unarchive:
-    src: "{{ ocp_stable_url }}/stable/openshift-client-linux.tar.gz"
+    src: "{{ ocp_stable_url }}/stable-4.14/openshift-client-linux.tar.gz"
```
tkrishtop, line 104: Should it be version-dependent? Like `stable-4.15` for 4.15 etc? You can get this value from ocp_version variable

ramperher: I'm pinning this to 4.14 because we do the same for `opm`. The problem is that, if we rely on the `ocp_version` variable, this will not work for OCP >= 4.16 because of the GLIBC issue I've commented.

manurodriguez: The `stable-4.14` could be the simplest option, but I believe we could use `stable-4.15` or `stable-4.16` and explicitly use the binary for rhel8 or rhel9. Starting in 4.15 there are several binaries provided for both. for example:
 https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable-4.16/openshift-client-linux-amd64-rhel8.tar.gz
 https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable-4.16/openshift-client-linux-amd64-rhel9.tar.gz

And as I mentioned in 31882, we could determine the client to use based on an ansible fact:
```
https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-4.16/openshift-client-linux-amd64-rhel{{ ansible_distribution_major_version }}.tar.gz
```
We'll have to test to make sure that works, but it sounds like a promising option.

ramperher: That's a good option, let me try that.

roles/sno_installer/tasks/10_get_oc.yml
```diff
@@ -101,7 +101,7 @@
   vars:
     ocp_stable_url: https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp
   ansible.builtin.unarchive:
-    src: "{{ ocp_stable_url }}/stable/openshift-client-linux.tar.gz"
+    src: "{{ ocp_stable_url }}//stable/openshift-client-linux-amd64-rhel{{ ansible_distribution_major_version }}.tar.gz"
```
manurodriguez, line 104: extra `/` is causing download issues in the last job :P
```suggestion
    src: "{{ ocp_stable_url }}/stable/openshift-client-linux-amd64-rhel{{ ansible_distribution_major_version }}.tar.gz"
```

ramperher: my bad, fixed it

ramperher:
Now it's passing: https://www.distributed-ci.io/jobs/17079658-8486-4a6e-935f-ab4451e853b2/jobStates?sort=date&task=5645178e-4f3e-4d21-bbbe-fe16df684a6a

manurodriguez (approved):
Thanks, LGTM

Issue #369, opened 2024-07-05
When powering off nodes with IPMI some servers return an error when they are already off - IPI installer role


manurodriguez:
Role: `installer`
Playbook: `59_power_off_cluster_servers.yml` 

Problem
We recently moved from using Ansible ipmi module to use ipmitool command to power off nodes in #314 because some dependencies were deprecated. 
Some servers are returning an error when powering off, if they are already powered off, the ipmitool command returns an error message like this:
```
Set Chassis Power Control to Down/Off failed: Command not supported in present state
```
And this makes ansible play to fail.

Some nodes do not through an error, they simply show:
```
Chassis Power is off
```

Proposal
While we find an alternative to go back to ansible ipmi module, we want to include some extra task in the playbook to verify the power status of the node, and power off only those nodes that are powered on.

PR #370, opened 2024-07-05
ipmi poweroff only the powered on nodes - role [installer]

manurodriguez:
Fixes: #369

TestBos2: virt

### DIFF
--- a/roles/installer/tasks/59_power_off_cluster_servers.yml
+++ b/roles/installer/tasks/59_power_off_cluster_servers.yml
@@ -18,7 +18,7 @@
       loop: "{{ groups.masters + groups.workers | default([]) }}"
 
     - name: Check node power status via ipmi to filter off nodes
-      ansible.builtin.command: >
+      ansible.builtin.command: >-
         ipmitool -I lanplus -H {{ hostvars[item]['ipmi_address'] }}
         -L ADMINISTRATOR -U {{ hostvars[item]['ipmi_user'] }}
         -R7 -N 5 -P {{ hostvars[item]['ipmi_password'] }}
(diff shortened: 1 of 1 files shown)


roles/installer/tasks/59_power_off_cluster_servers.yml
```diff
@@ -31,7 +58,9 @@
       retries: 10
       delay: 5
       with_items: "{{ groups['poweroff_hosts'] }}"
-      when: groups['poweroff_hosts'] is defined
+      when:
+        - groups['poweroff_hosts'] is defined
+        - item in installer_powered_on_hosts | default([])
```
tkrishtop, line 63: does it make sense to use loop_var here as well?

manurodriguez: sure, I replaced all the `with_items`  with loop in this file.

roles/installer/tasks/59_power_off_cluster_servers.yml
```diff
@@ -15,10 +15,37 @@
         ( item not in groups['dell_hosts_redfish'] | default([]) )
           and
         ( item not in groups['hp_hosts_redfish'] | default([]) )
-      with_items:
+      loop:
         - "{{ groups.masters }}"
         - "{{ groups.workers | default([]) }}"
```
tonyskapunk, line 20: The change to loop is causing it to fail, this will get it to work
```suggestion
      loop: "{{ groups.masters + groups.workers | default([]) }}"
```

tonyskapunk (changes requested):

tonyskapunk (approved):
LGTM

PR #371, opened 2024-07-06
Use api_vips and ingress_vips

betoredhat:
api_vip and ingress_vips is deprecated, we need to move to an assisted version that supports ocp-4.16

Test-Hints: no-check

### DIFF
--- a/roles/create_cluster/tasks/main.yml
+++ b/roles/create_cluster/tasks/main.yml
@@ -33,8 +33,16 @@
 - name: Assign api_vip and ingress_vip of not doing SNO
   set_fact:
     create_body_extras: "{{ create_body_extras | combine({
-        'api_vip': api_vip,
-        'ingress_vip': ingress_vip,
+        'ingress_vips': [
+          {
+            'ip': ingress_vip,
+          }
+        ],
+        'api_vips': [
+          {
+            'ip': api_vip,
+          }
+        ],
       }) }}"
   when: not (single_node_openshift_enabled | bool)
 
--- a/roles/install_cluster/tasks/main.yml
+++ b/roles/install_cluster/tasks/main.yml
@@ -45,8 +45,16 @@
     set_fact:
       cluster_config_patch_request_body: {
         "vip_dhcp_allocation": "{{ vip_dhcp_allocation | bool }}",
-        "ingress_vip": "{{ ingress_vip }}",
-        "api_vip": "{{ api_vip }}",
+        "ingress_vips": [
+          {
+            "ip": "{{ ingress_vip }}",
+          }
+        ],
+        "api_vips": [
+          {
+            "ip": "{{ api_vip }}",
+          }
+        ],
       }
     when: not single_node_openshift_enabled
 


tonyskapunk:
Do you remember ir `api_vip` and `ingress_vip` got deprecated, and in what version?

I think we need to switch to add support to `api_vips`  and `ingress_vips`. And only allow `api_vip`, `ingress_vip` for those versions that still support it.

manurodriguez:
> Do you remember ir `api_vip` and `ingress_vip` got deprecated, and in what version?
> 
> I think we need to switch to add support to `api_vips` and `ingress_vips`. And only allow `api_vip`, `ingress_vip` for those versions that still support it.

It's supposed to be in 4.12, but I remember when I changed it in the baremetal_deploy repo last year it was still working, in theory anything before 4.12 should work the singular form: https://github.com/openshift-kni/baremetal-deploy/pull/953 

From the latest 4.12 Installer, I see it's still showing the deprecated message:
```
[dciteam@jumphost ~]$ /opt/cache/4.12.60/openshift-baremetal-install version
/opt/cache/4.12.60/openshift-baremetal-install 4.12.60
built from commit 596554b163e55e81f35954804a69a09c1734eebf
release image quay.io/openshift-release-dev/ocp-release@sha256:22174072f8aaef46f80bded25d5a2bdfa6fd8bb01fc5242a07e550f132cafc5a
release architecture amd64

[dciteam@jumphost ~]$ /opt/cache/4.12.60/openshift-baremetal-install explain installconfig.platform.baremetal | less
...
    apiVIP <string>                                
      Format: ip
      DeprecatedAPIVIP is the VIP to use for internal API communication Deprecated: Use APIVIPs                                                              
                                                                                                                                                              
    apiVIPs <[]string>
      Format: ip
      APIVIPs contains the VIP(s) to use for internal API communication. In dual stack clusters it contains an IPv4 and IPv6 address, otherwise only one VIP 
...
    ingressVIP <string>
      Format: ip                                                                                                                         
      DeprecatedIngressVIP is the VIP to use for ingress traffic Deprecated: Use IngressVIPs                                                                 
                             
    ingressVIPs <[]string>                                                                                                                                    
      Format: ip                              
      IngressVIPs contains the VIP(s) to use for ingress traffic. In dual stack clusters it contains an IPv4 and IPv6 address, otherwise only one VIP      
```

tonyskapunk:
> It's supposed to be in 4.12

Indeed, it was in 4.12 where it was deprecated[0]:

| Feature  | 4.10 | 4.11 | 4.12
| ------------ | ------ | ------ | -------
| `ingressVIP` and `apiVIP` settings in the install-config.yaml file for installer-provisioned infrastructure clusters | General Availability | General Availability | Deprecated

> but I remember when I changed it in the baremetal_deploy repo last year it was still working

Right, it is because is `Deprecated` and not yet `Removed`. Looked into the current latest release 4.16[1] and is still `Deprecated` but not `Removed` so we should be good for now in IPI

| Feature  | 4.14 | 4.15 | 4.16
| ------------ | ------ | ------ | -------
| `ingressVIP` and `apiVIP` settings in the install-config.yaml file for installer-provisioned infrastructure clusters | Deprecated | Deprecated | Deprecated

---

Back into assisted installer I found this also in the same release notes of 4.12, but different section[2] that the plural option was added:

> Assisted Installer supports installation of OpenShift Container Platform 4.12 and later versions with dual stack networking for the API VIP and Ingress VIP on bare metal only. This support introduces two new configuration settings: `api_vips` and `ingress_vips`, which can take a list of IP addresses. The legacy settings, `api_vip` and `ingress_vip` **must also be set in OpenShift Container Platform 4.12**; however, since they only take one IP address, you must set the IPv4 address when configuring dual stack networking for the API VIP and Ingress VIP with the legacy api_vip and ingress_vip configuration settings.

>  The API VIP address and the Ingress VIP address must be of the primary IP address family when using dual-stack networking. Currently, Red Hat does not support dual-stack VIPs or dual-stack networking with IPv6 as the primary IP address family. However, Red Hat does support dual-stack networking with IPv4 as the primary IP address family. Therefore, **you must place the IPv4 entries before the IPv6 entries**. See the Assisted Installer documentation for details.

I couldn't find any other place where the singular form was removed :crying_cat_face: , the only doc I found was a 2024 assisted installer doc[3] where only the plural form is mentioned.  I believe is time to make the switch and validate 4.12+ can use the latest version of the installer, as the expectation is that it must.


[0] https://docs.openshift.com/container-platform/4.12/release_notes/ocp-4-12-release-notes.html#ocp-4-12-deprecated-removed-features
[1] https://docs.openshift.com/container-platform/4.16/release_notes/ocp-4-16-release-notes.html#ocp-4-16-deprecated-removed-features_release-notes
[2] https://docs.openshift.com/container-platform/4.12/release_notes/ocp-4-12-release-notes.html#support-for-dual-stack-addressing-for-the-API-VIP-and-Ingress-VIP
[3] https://docs.redhat.com/en/documentation/assisted_installer_for_openshift_container_platform/2024/html-single/installing_openshift_container_platform_with_the_assisted_installer/index#vip-dhcp-allocation_network-configuration

betoredhat:
Re-opening this. I think the change for this should be based on the version of AI that supported apiVIPs and ingressVIPs, not when the parameters were added to OCP. 

Having this working is important to support dual stack in assisted.

betoredhat:
not needed anymore

PR #372, opened 2024-07-08
Small refactor

reihl:
Add dash to command so it'll be single line,
Changed with_item to new loop syntax

TestBos2: virt

### DIFF
--- a/roles/installer/tasks/59_power_off_cluster_servers.yml
+++ b/roles/installer/tasks/59_power_off_cluster_servers.yml
@@ -20,7 +20,7 @@
         - "{{ groups.workers | default([]) }}"
 
     - name: Power off hosts via IPMI
-      ansible.builtin.command: >
+      ansible.builtin.command: >-
         ipmitool -I lanplus -H {{ hostvars[item]['ipmi_address'] }}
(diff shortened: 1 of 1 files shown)


roles/installer/tasks/59_power_off_cluster_servers.yml
```diff
@@ -20,7 +20,7 @@
         - "{{ groups.workers | default([]) }}"
 
     - name: Power off hosts via IPMI
-      ansible.builtin.command: >
+      ansible.builtin.command: >-
```
tonyskapunk, line 23: In other jobs I don't see the newline `\n` added to the command, last entries I see in the command are:
```
"power", "off"]
```

I'm curious if you see it and if you have a job example. Nonetheless I see no issues with removing it from the end.

tonyskapunk (approved):

tonyskapunk (changes requested):
Oh, my bad, https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/370 is also touching exactly the same file and is doing a more thorough validation on the state of a host. There we can add your suggestions.

Let's not merge this change, nothing wrong, just taking advantage of the other change to include the removal of the newline

PR #373, opened 2024-07-08
Migration of nfv-example-cnf-deploy roles to redhatci.ocp.collection

capolrik:
##### SUMMARY

- Migrating the roles from https://github.com/rh-nfv-int/nfv-example-cnf-deploy
- Updating catalog_source role to include a new field required by example-cnf catalog, setting default value as empty to not affect to current use cases

##### ISSUE TYPE

- New or Enhanced Feature

##### Tests

- [x] example-cnf deployment in direct mode - https://www.distributed-ci.io/jobs/1b0bc57c-2acc-413e-9a67-40818fd83617/jobStates?sort=date
- [x] example-cnf deployment in LB mode - https://www.distributed-ci.io/jobs/40aa9647-57d5-4d5c-8adf-a85d2483cfd3/jobStates/?sort=date
- [x] example-cnf deployment in LB mode with preflight and certsuite execution - https://www.distributed-ci.io/jobs/56d33661-20cd-47e3-8d1c-a6876f02f4b9/jobStates?sort=date
- [x] example-cnf extra trex job + draining scenario - https://www.distributed-ci.io/jobs?limit=20&offset=0&sort=-created_at&where=pipeline_id:af8c00f4-7683-41c3-aa32-8bc661931e35,state:active
- [x] example-cnf upgrade validation scenario - https://www.distributed-ci.io/jobs?limit=20&offset=0&sort=-created_at&where=pipeline_id:d2757257-08c8-4167-982d-77a4f36ea5e2,state:active

Test-Hints: no-check

### DIFF
--- a/roles/example_cnf_deploy/README.md
+++ b/roles/example_cnf_deploy/README.md
@@ -1,0 +1,190 @@
+# example_cnf_deploy
+
+This role allows the deployment of example-cnf on OCP, a DPDK based application. See the main project [here](https://github.com/openshift-kni/example-cnf)
+
+## Prerequisites
+
+OpenShift 4 Cluster with at least 3 worker nodes should be deployed with additional operators:
+
+- SR-IOV operator should be deployed and necessary resources for `SriovNetworkNodePolicy` and `SriovNetwork` should be created with target namespace (default namespace used to deploy example-cnf is `example-cnf`).
+- Node Tuning Operator should be deployed and a `PerformanceProfile` should be created.
+
+Also, exported `kubeconfig` file is required to run the Ansible tasks on the target cluster.
+
+## Variables
+
+| Variable                              | Default                                                                                         | Required | Description                                                                                     |
+|---------------------------------------|-------------------------------------------------------------------------------------------------|----------|------------------------------------------------------------------------------------------------|
+| ecd_action                            | None                                                                                            | yes      | Defines which actions to perform. Possible values: ['catalog', 'deploy', 'validate', 'deploy_extra_trex', 'draining'] |
+| ecd_cnf_namespace                     | example-cnf                                                                                     | no       | Name of the namespace where example-cnf is deployed |
+| ecd_registry_url                      | quay.io                                                                                         | no       | Registry URL |
+| ecd_repo_name                         | rh-nfv-int                                                                                      | no       | Repository name |
+| ecd_catalog_name                      | nfv-example-cnf-catalog                                                                         | no       | Catalog name |
+| ecd_operator_version                  | ""                                                                                              | yes      | Version of the index to be used |
+| ecd_catalog_image                     | "{{ ecd_registry_url }}/{{ ecd_repo_name }}/{{ ecd_catalog_name }}:{{ ecd_operator_version }}"  | no       | Index image to be used |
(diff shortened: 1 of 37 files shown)


ramperher (commented):
Some general comments for the moment, since I know this is WIP and there's still some work to do:

- There are several references to `job_logs` variable. We should provide it as a var (`ecd_log_folder`), and define it when calling to the role.
- I suppose you have in mind to update all vars to follow the collections good practices, just as a reminder.
- Before, in nfv-example-cnf-deploy, we had different roles, and within each role, we defined multiple files. Here, we have cases like draining.yml or validate.yml, where we are gathering the content of all files of example-cnf-draining role in one simple play, and it's a bit hard to follow the workflow in this way, from my point of view. Shouldn't it better to create a folder for each action, and place there the same file structure we had in nfv-example-cnf-deploy, rather than having everything gathered in one single file? Maybe this implies to have a larger project structure, with more folders and files, but probably for troubleshooting, would be better at the end.
- Maybe, it could be good to take the chance to create a new Github team, called `example-cnf`, and assigned to this new role, so that this team needs to check and validate new changes for this role. Similarly to what happens with `crucible` or `ztp`, for example.

ramperher:
The change is ready for review, in terms of testing, all possible scenarios are passing. I just have pending to make the sanity check to pass and to refactor the README. The rest of the code is ready for review, cc @capolrik

ramperher:
Commits have been squashed and the README is now updated, this change is ready.

roles/example_cnf_deploy/README.md
```diff
@@ -0,0 +1,190 @@
...
+| ecd_lb_gen_port_mac_list              | ["40:04:0f:f1:89:01","40:04:0f:f1:89:02"]                                                       | no       | Static MAC addresses used by the LoadBalancer in its connection with TRex |
+| ecd_lb_cnf_port_mac_list              | ["60:04:0f:f1:89:01","60:04:0f:f1:89:02"]                                                       | no       | Static MAC addresses used by the LoadBalancer in its connection with CNFApp |
+| ecd_testpmd_app_mac_list              | ["80:04:0f:f1:89:01","80:04:0f:f1:89:02"]                                                       | no       | Static MAC addresses used by CNFApp |
+| ecd_cnf_app_networks                  | []                                                                                              | yes      | Connection between LB-CNF in LB mode (or TRex-CNF in direct mode) |
+| ecd_packet_generator_networks         | []                                                                                              | yes       | Connection between TRex-LB in LB mode (not used in direct mode). Required in LB mode |
```
manurodriguez, line 41:
```suggestion
| ecd_packet_generator_networks         | []                                                                                              | yes      | Connection between TRex-LB in LB mode (not used in direct mode). Required in LB mode |
```

manurodriguez (dismissed):
LGTM, only a couple of things in the README to update if possible
Also, I wonder if it's worth it to bump up the version in `ansible-collection-redhatci-ocp.spec` because of the new role addition.

ramperher:
> LGTM, only a couple of things in the README to update if possible Also, I wonder if it's worth it to bump up the version in `ansible-collection-redhatci-ocp.spec` because of the new role addition.

I didn't do it on a first instance because it didn't imply a change in our agents, but I'll do it anyway, thanks!

PR #374, opened 2024-07-08
Check if node exists without fail

reihl:
Check if node exists without fail

Test-Hints: no-check

Successful job: https://www.distributed-ci.io/jobs/251779e1-c1c1-42b8-acc2-7cf343ba774d/jobStates

### DIFF
--- a/roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
+++ b/roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
@@ -1,20 +1,29 @@
 ---
 - name: "Does node exist"
   community.libvirt.virt:
+    command: list_vms
+  register: _ool_vms_list
+
+- name: "Set node_not_exists"
+   ansible.builtin.set_fact:
+     ool_node_exists: "{{ host['name'] in _ool_vms_list }}"
+
+- name: "Get node status if exists"
+  community.libvirt.virt:
     name: "{{ host['name'] }}"
     command: status
-  ignore_errors: true
-  register: node_exists
+  register: _ool_node_status
+  when: ool_node_exists
 
 - name: "Destroy node"
   community.libvirt.virt:
     name: "{{ host['name'] }}"
     command: destroy
-  when: node_exists['status'] is defined and node_exists['status'] == 'running'
+  when: _ool_node_status is defined and _ool_node_status['status'] == 'running'
 
 - name: "Undefine node"
   become: true
   ansible.builtin.command:
     # community.libvirt.virt undefine can't specify --nvram
     cmd: "virsh undefine --remove-all-storage --nvram {{ host['name'] }}"
-  when: not node_exists.failed
+  when: ool_node_exists
--- a/roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
+++ b/roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
@@ -1,27 +1,27 @@
 ---
-- name: "Does node exist"
+- name: "List nodes for {{ host['name'] }}"
   community.libvirt.virt:
     command: list_vms
   register: _ool_vms_list
 
-- name: "Set node_not_exists"
+- name: "Set ool_node_exists"
   ansible.builtin.set_fact:
     ool_node_exists: "{{ host['name'] in _ool_vms_list.list_vms }}"
 
-- name: "Get node status if exists"
+- name: "Get status if exists - {{ host['name'] }}"
   community.libvirt.virt:
     name: "{{ host['name'] }}"
     command: status
   register: _ool_node_status
   when: ool_node_exists
 
-- name: "Destroy node"
+- name: "Destroy {{ host.name }}"
   community.libvirt.virt:
     name: "{{ host['name'] }}"
     command: destroy
   when: _ool_node_status['status'] is defined and _ool_node_status['status'] == 'running'
 
-- name: "Undefine node"
+- name: "Undefine {{ host.name }}"
   become: true
   community.libvirt.virt:
     name: "{{ host['name'] }}"
--- a/roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
+++ b/roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
@@ -5,8 +5,8 @@
   register: _ool_vms_list
 
 - name: "Set node_not_exists"
-   ansible.builtin.set_fact:
-     ool_node_exists: "{{ host['name'] in _ool_vms_list }}"
+  ansible.builtin.set_fact:
+    ool_node_exists: "{{ host['name'] in _ool_vms_list.list_vms }}"
 
 - name: "Get node status if exists"
   community.libvirt.virt:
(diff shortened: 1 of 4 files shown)


roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
```diff
@@ -1,20 +1,31 @@
 ---
 - name: "Does node exist"
+  community.libvirt.virt:
+    command: list_vms
+  register: _ool_vms_list
+
+- name: "Set node_not_exists"
+   ansible.builtin.set_fact:
```
manurodriguez, line 8: Last job failed because of an extra space of indentation here.
```suggestion
  ansible.builtin.set_fact:
```

manurodriguez (approved):

reihl:
retest

roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
```diff
@@ -1,20 +1,31 @@
 ---
-- name: "Does node exist"
+- name: "Does node exist - {{ host['name'] }}"
```
ramperher, line 2: I think this task can be generalized and executed from https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/ocp_on_libvirt/tasks/main.yml, probably as a first task that can be used for both install and cleanup hook_action, because the output is exactly the same on every iteration for both cases and the variable name is the same.
I would also change the task name to something more representative, such as `List VMs`

roles/ocp_on_libvirt/tasks/libvirt_host_up.yml
```diff
@@ -1,66 +1,69 @@
 ---
-- name: "Does node already exist"
+- name: "Does node already exist - {{ host['name'] }}"
```
ramperher, line 2: Same than what I've said in `libvirt_host_destroy.yml`

roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
```diff
@@ -1,20 +1,31 @@
...
+  community.libvirt.virt:
+    command: list_vms
+  register: _ool_vms_list
+
+- name: "Set node_not_exists"
```
ramperher, line 7: If keeping this task (i.e. not merging this one and the next one), should we rename it to `Set ool_node_exists`, instead?

roles/ocp_on_libvirt/tasks/libvirt_host_up.yml
```diff
@@ -1,66 +1,69 @@
...
   community.libvirt.virt:
     name: "{{ host['name'] }}"
-    command: status
-  ignore_errors: true
-  register: node_exists
+    command: list_vms
+  register: _ool_vms_list
+
+- name: "Set node_not_exists"
```
ramperher, line 8: `Set ool_node_not_exists`, instead?

roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
```diff
@@ -1,20 +1,31 @@
...
+  ansible.builtin.set_fact:
+    ool_node_exists: "{{ host['name'] in _ool_vms_list.list_vms }}"
+
+- name: "Get status if exists - {{ host['name'] }}"
```
ramperher, line 11: Maybe we can merge this task and the previous one in the following way (if we also use `_ool_node_status` variable in the last task of this file):

```
- name: "Get status if exists - {{ host['name'] }}"
  vars:
    ool_node_exists: "{{ host['name'] in _ool_vms_list.list_vms }}"
  community.libvirt.virt:
    name: "{{ host['name'] }}"
    command: status
  register: _ool_node_status
  when: ool_node_exists
```

roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
```diff
@@ -1,20 +1,31 @@
...
   become: true
-  ansible.builtin.command:
-    # community.libvirt.virt undefine can't specify --nvram
-    cmd: "virsh undefine --remove-all-storage --nvram {{ host['name'] }}"
-  when: not node_exists.failed
+  community.libvirt.virt:
+    name: "{{ host['name'] }}"
+    command: undefine
+    flags: nvram
+  when: ool_node_exists
```
ramperher, line 30: Should we use the same condition than in the previous task? Just to follow the same terminology: `when: _ool_node_status['status'] is defined and _ool_node_status['status'] == 'running'`

roles/ocp_on_libvirt/tasks/libvirt_host_up.yml
```diff
@@ -1,66 +1,69 @@
...
+  ansible.builtin.set_fact:
+    ool_node_not_exists: "{{ host['name'] not in _ool_vms_list.list_vms }}"
 
 - name: "Set qemu_path"
```
ramperher, line 12: `Set ool_qemu_path`?

roles/ocp_on_libvirt/tasks/libvirt_host_up2.yml
ramperher, line 1: I think this file is wrongly designed, in fact. We should check if the VM *exists* and use that condition to launch the tasks of this play, because the VM was already created and started previously in `libvirt_host_up.yml`. Here, we're using, in fact, a wrong condition because `node_exists.failed` (or `ool_node_not_exists`) is referring to the last iteration of `libvirt_host_up.yml`, and will not be updated for each VM we have in the scenario.
Probably we can take the chance and apply this update in this change, now that we're refactoring this code. What do you think?

ramperher (commented):
Some comments from my side

roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
reihl: You will get different list each iteration you will create new node or destroy one, but we can take this task outside

roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
reihl: No need cause the previous one should move the status from running to shutdown

roles/ocp_on_libvirt/tasks/libvirt_host_up2.yml
reihl: I think you're right and I'm working in my head on more refactor to this stage, no need to get the hosts from out side, we can do the create and destroy internally

roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
reihl: In the next PR I'll create loop on "hosts" with no need to get single node from outside

reihl:
@ramperher WDYT, Can we merge this changes and create a big refactor in new PR?

roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
ramperher: ack, let's do that, both options are fine for me

roles/ocp_on_libvirt/tasks/libvirt_host_destroy.yml
ramperher: ack

roles/ocp_on_libvirt/tasks/libvirt_host_up2.yml
ramperher: ack, let's do it in a separate change

ramperher:
> @ramperher WDYT, Can we merge this changes and create a big refactor in new PR?

That's fine for me, sure!

ramperher (approved):
LGTM

PR #379, opened 2024-07-11
Iterate nodes inside the task

reihl:
Iterate nodes inside the task

TestBos2: virt


roles/ocp_on_libvirt/tasks/libvirt_host_up.yml
```diff
@@ -1,74 +1,124 @@
...
 - name: "Create an additional disk"
-  ansible.builtin.command: "qemu-img create -f qcow2 {{ libvirt_image_path }}/{{ host['name'] }}-disk2.qcow2 {{ host['disk2_size'] }} "
+  ansible.builtin.command: "qemu-img create -f qcow2 {{ libvirt_image_path }}/{{ host['name'] }}-disk2.qcow2 {{ host['disk2_size'] }}"
   become: "{{ libvirt_become }}"
+  loop: "{{ hosts }}"
+  loop_control:
+    loop_var: host
   when:
-    - host['disk2_size'] is defined and ool_node_not_exists
+    - host['disk_size'] is defined and host['name'] in ool_nodes_list
```
thekad, line 44: the parameter `disk2_size` actually creates a 2nd disk attached to the VM, check `templates/libvirt_node.xml.j2` you would have to rework that (maybe make it an array? or just drop the second disk altogether)

roles/ocp_on_libvirt/tasks/libvirt_host_up.yml
```diff
@@ -1,74 +1,124 @@
...
 
 - name: "Resize OS disk"
-  ansible.builtin.command: "qemu-img resize {{ libvirt_image_path }}/{{ host['name'] }}.qcow2 +{{ host['disk_size'] }}"
   become: "{{ libvirt_become }}"
-  when: host['disk_size'] is defined and ool_node_not_exists
+  ansible.builtin.command: "qemu-img resize {{ libvirt_image_path }}/{{ host['name'] }}.qcow2 +{{ host['disk_size'] }}"
+  loop: "{{ hosts }}"
+  loop_control:
+    loop_var: host
+  when: host['disk_size'] is defined and host['name'] in ool_nodes_list
```
thekad, line 35: I feel that if `name` is going to be a required parameter there should be a validation at the top of the role, or force the end user to set it somehow e.g. make it a dictionary instead of a list

roles/ocp_on_libvirt/tasks/main.yml
```diff
@@ -13,14 +13,6 @@
         loop_var: network
     - name: Create hosts
       ansible.builtin.include_tasks: libvirt_host_up.yml
-      loop: "{{ hosts }}"
-      loop_control:
-        loop_var: host
-    - name: Create hosts 2
-      ansible.builtin.include_tasks: libvirt_host_up2.yml
```
thekad, line 20: this makes `tasks/libvirt_host_up2.yml` unnecessary, please delete

roles/ocp_on_libvirt/tasks/libvirt_host_up.yml
```diff
@@ -1,74 +1,124 @@
...
   ansible.builtin.command: sync
 
 - name: "Start VM"
   community.libvirt.virt:
-    name: "{{ host['name'] }}"
+    name: "{{ item }}"
+    state: "running"
+  loop: "{{ ool_nodes_list }}"
+
+- name: "Wait for VM to shut down"
```
thekad, line 101: it is not clear to me how the VM goes from running to shutdown without intervention, I assume this expects the installer to shut it down? clarify with a comment in the code

thekad (changes requested):
I am not really sure where this roles is used in the agents other than samples, as there's a second role `create_vms` that is a tad bit more flexible. If you are using this role then great, but I'd recommend investigating `create_vms` as an alternative cc @tonyskapunk

ramperher:
> I am not really sure where this roles is used in the agents other than samples, as there's a second role `create_vms` that is a tad bit more flexible. If you are using this role then great, but I'd recommend investigating `create_vms` as an alternative cc @tonyskapunk

I think that, as you said, this role is only used in the `ocp_on_libvirt` sample. In fact we use it for the `virt` pipeline we have defined in BOS2 setup. But could be interesting to transition towards `create_vms` at some point.

PR #381, opened 2024-07-12
Updating odf_setup role

semihkisa:
Updating odf_setup oc path and adding default value for local_storage_devices.

---

- [x] TestBos2: virt libvirt:ansible_extravars=enable_lso:true libvirt:ansible_extravars=enable_odf:true libvirt:ansible_extravars='{"dci_operators":[{"name":"odf-operator","catalog_source":"redhat-operators","namespace":"openshift-storage","operator_group_name":"sg","operator_group_spec":{"targetNamespaces":["openshift-storage"]}},{"name":"local-storage-operator","catalog_source":"redhat-operators","namespace":"openshift-local-storage","operator_group_name":"ols","operator_group_spec":{"targetNamespaces":["openshift-local-storage"]}}]}' -  https://www.distributed-ci.io/jobs/d5fa5521-4394-4f71-b94c-179885536326

---
Test-hints: no-check

### DIFF
--- a/roles/odf_setup/README.md
+++ b/roles/odf_setup/README.md
@@ -27,9 +27,11 @@ Role Variables
 | default_storageclass_annotation  | '{"storageclass.kubernetes.io/is-default-class": "true"}'  | String       | No          | Default storageclass annotation             |
 | external_ceph_data               |                               | JSON         | No          | A JSON payload generated from RHCS                                       |
 | ocs_install_type                 |                               | String       | Yes         | `internal` for LSO, `external` for Ceph/RHCS                             |
-| local_storage_devices            |                               | List         | No          | For LSO, a list of local devices that will be use as backend             |
+| local_storage_devices            |                               | List         | Yes          | For LSO, a list of local devices that will be use as backend             |
 | ocs_default_storage_class        | storagecluster-cephfs         | String       | No          | Default storage class name                                               |
 | gatherer_image                   | registry.access.redhat.com/ubi8/ubi | String | No          | Image for disk-gatherer deployment                                       |
+| odf_setup_oc_tool_path                   | '/usr/local/bin/oc` | String | No          | Path to the OpenShift Command Line Interface binary.
+
 
 Inventory Groups and Variables
 --------------
@@ -43,9 +45,8 @@ ocs_install_type=
 # with ceph-external-cluster-details-exporter.py script
 external_ceph_data='JSON_PAYLOAD'
 
-# (Optional) when enable_lso=true List of disk devices per node to use for LSO
-# If not specified, it will use all the local disks available
-# comma separated, all servers must have the same
+# When enable_lso=true, list of disk devices per node to use for LSO
+# Comma separated, all servers must have the same
 local_storage_devices=["/dev/sdX", "/dev/sdY", "/dev/sdZ"]
 
 # (Optional) Default storage class name
--- a/roles/odf_setup/defaults/main.yml
+++ b/roles/odf_setup/defaults/main.yml
@@ -30,3 +30,6 @@ default_storageclass_annotation: '{"storageclass.kubernetes.io/is-default-class"
 
 # OCS gatherer_image
 gatherer_image: registry.access.redhat.com/ubi8/ubi
+
+# OpenShift Command Line Interface path
+odf_setup_oc_tool_path: /usr/local/bin/oc
--- a/roles/odf_setup/tasks/validation.yml
+++ b/roles/odf_setup/tasks/validation.yml
@@ -52,4 +52,11 @@
           There are less than 3 nodes with the label
           cluster.ocs.openshift.io/openshift-storage
       when: nodes_info.resources | length < 3
+
+- name: Assert that local_storage_devices is defined and is a list with at least one item
+  ansible.builtin.assert:
+    that:
+      - local_storage_devices is defined
+      - local_storage_devices | length > 0
+    fail_msg: "local_storage_devices must be a list with at least one item"
 ...
--- a/roles/odf_setup/tasks/local-storage-operator.yml
+++ b/roles/odf_setup/tasks/local-storage-operator.yml
@@ -27,7 +27,7 @@
   ansible.builtin.shell:
     cmd: >
       sleep 90;
-      {{ oc_tool_path }} logs
+      {{ odf_setup_oc_tool_path }} logs
       --selector name=ocs-disk-gatherer
       --tail=-1
       --since=10m


roles/odf_setup/README.md
```diff
@@ -30,6 +30,8 @@ Role Variables
 | local_storage_devices            |                               | List         | No          | For LSO, a list of local devices that will be use as backend             |
 | ocs_default_storage_class        | storagecluster-cephfs         | String       | No          | Default storage class name                                               |
 | gatherer_image                   | registry.access.redhat.com/ubi8/ubi | String | No          | Image for disk-gatherer deployment                                       |
+| odf_oc_tool_path                   | `{{ oc_tool_path | default('/usr/local/bin/oc) }}` | String | No          | Path to the OpenShift Command Line Interface binary.
```
tonyskapunk, line 33: This role was imported when we did not follow a naming convention on role variables, but now we do validate that on changes, so let's rename the variable to `odf_setup_oc_tool_path` or `os_oc_tool_path` that should satisfy the naming convention

I suggest to make this default to `/usr/local/bin/oc`, the override will come through the variable itself (`odf_setup_oc_tool_path` or the other var name)

roles/odf_setup/templates/openshift-storage-cluster.yml.j2
```diff
@@ -16,7 +16,7 @@ spec:
   manageNodes: false
   monDataDirHostPath: /var/lib/rook
   storageDeviceSets:
-  - count: {{ local_storage_devices | length }}
+  - count: {{ local_storage_devices | default([1]) | length }}
```
tonyskapunk, line 19: let's set the default of `local_storage_devices` in the default/main.yml directory instead. 

Because of the naming convention we most likely need to update this variable to use the role prefix, that would be either `odf_setup_`  or `os_`  any of them are fine.

roles/odf_setup/defaults/main.yml
```diff
@@ -30,3 +30,6 @@ default_storageclass_annotation: '{"storageclass.kubernetes.io/is-default-class"
 
 # OCS gatherer_image
 gatherer_image: registry.access.redhat.com/ubi8/ubi
+
+# OpenShift Command Line Interface path
+odf_oc_tool_path: "{{ oc_tool_path | default('/usr/loca/bin/oc') }}"
```
tonyskapunk, line 35: Let's deprecate the use of `oc_tool_path` and instead use the role variable i.e. `<role_prefix>_oc_tool_path`

tonyskapunk (changes requested):
Thanks @semihkisa for your contribution in this repo, I have some suggestions that will help to follow the role var naming convention in the repo.

Thanks for signing your commit, it's advised to also upload your **public** key into your GH account so the verification can be completed.

roles/odf_setup/defaults/main.yml
thekad: the default path is also wrong s/loca/local/

roles/odf_setup/templates/openshift-storage-cluster.yml.j2
semihkisa: Well, if we set the default value in `default/main.yml` then it will also affect this part: https://github.com/semihkisa/ansible-collection-redhatci-ocp/blob/update_odf_role/roles/odf_setup/templates/ocs-disk-gatherer.yml.j2#L84

Which I am not sure if this is the intention. The reason why I assigned a default value there is just to have minimum count for `storageDeviceSets`. So I'd rather set this variable a required one and don't set a default device list. What do you think?

thekad: > Well, if we set the default value in `default/main.yml` then it will also affect this part: https://github.com/semihkisa/ansible-collection-redhatci-ocp/blob/update_odf_role/roles/odf_setup/templates/ocs-disk-gatherer.yml.j2#L84
> 
> Which I am not sure if this is the intention. The reason why I assigned a default value there is just to have minimum count for `storageDeviceSets`. So I'd rather set this variable a required one and don't set a default device list. What do you think?

if there is a required minimum that it absolutely needs to work, then maybe the variable should be reworked into something like `extra_storage_devices` which defaults to 0/none/empty. Then in the template you would add the required 1 + however many extra devices.

betoredhat:
recheck

manurodriguez:
> from change #381:
> 
> * FAILURE https://www.distributed-ci.io/jobs/43ddbeaa-3f5a-44d4-a7df-d4df32e0b662/jobStates

From the failed job, we need to define an extra variable: `libvirt:ansible_extravars=ocs_install_type:internal`

betoredhat:
recheck

betoredhat:
Let’s check if the role still works with the latest operators  versions. ODF had lot of changes in 4.16

betoredhat:
recheck

betoredhat:
recheck

roles/odf_setup/templates/openshift-storage-cluster.yml.j2
betoredhat: @semihkisa I think we should make the `local_storage_devices` required and assert that it is a list of at least 1 item. 

Sorry for the delay, I was checking how to validate this on our lab and the PTO. Cheers.

semihkisa: Np at all @betoredhat, I was also thinking that this var should be required so now I made the change to make it required and added a small step in the validation.

betoredhat:
retest

roles/odf_setup/tasks/validation.yml
```diff
@@ -52,4 +52,11 @@
           There are less than 3 nodes with the label
           cluster.ocs.openshift.io/openshift-storage
       when: nodes_info.resources | length < 3
+
+- name: Assert that local_storage_devices is defined and is a list with at least one item
+  assert:
```
betoredhat, line 57: Just use the fqcn for assert and we should be good with this change. Thanks!

ramperher:
> from change #381:
> 
> * SUCCESS https://www.distributed-ci.io/jobs/5af72326-e12b-4a8b-8e26-66d37be790a6/jobStates

Are we really testing the new code in the last job? Because I cannot see the new assert task listed there

betoredhat:
> > from change #381:
> > 
> > * SUCCESS https://www.distributed-ci.io/jobs/5af72326-e12b-4a8b-8e26-66d37be790a6/jobStates
> 
> Are we really testing the new code in the last job? Because I cannot see the new assert task listed there
> 
> Also, I don't see `local_storage_devices` defined in the `TestBos2` entry to test it properly. Is that intended?

Yes, there is something wrong with the outputs in the DCI GUI. Those are truncated, but if you check the ansible.log you will see the details. 

Regarding the local_storage_devices, in our case, those are defined in the inventory.

betoredhat:
recheck

betoredhat (approved):
lgtm, thanks!

tonyskapunk (approved):

betoredhat:
@semihkisa   thanks for your contribution!

PR #382, opened 2024-07-12
HCP - Use oc binary path from the role

betoredhat:
- Use the oc binary downloaded by role
- Fix typos
- Set time for an existing hcp cluster to be deleted

Test-Hints: no-check

### DIFF
--- a/roles/acm_hypershift/tasks/delete-cluster.yml
+++ b/roles/acm_hypershift/tasks/delete-cluster.yml
@@ -15,8 +15,23 @@
       --name={{ ah_cluster_generated_name }}
       --namespace={{ ah_clusters_ns }}
       --infra-id={{ ah_cluster_generated_name }}
-      --cluster-grace-period 0
   changed_when: false
   when:
-    - hosted_cluster.resources | length == 1
+    - hosted_cluster.resources | default([]) | length == 1
+
+- name: Confirm that HCP is deleted
+  community.kubernetes.k8s_info:
+    api_version: hypershift.openshift.io/v1beta1
+    kind: HostedCluster
+    namespace: "{{ ah_clusters_ns }}"
+    name: "{{ ah_cluster_generated_name }}"
+  retries: 10
+  delay: 6
+  register: _ah_hosted_cluster
+  until:
+    - _ah_hosted_cluster is defined
+    - _ah_hosted_cluster.resources is defined
+    - _ah_hosted_cluster.resources | length == 0
+  when:
+    - hosted_cluster.resources | default([]) | length == 1
 ...
--- a/roles/acm_hypershift/tasks/download-cli.yml
+++ b/roles/acm_hypershift/tasks/download-cli.yml
@@ -2,9 +2,9 @@
 - name: Download stable openshift client
   vars:
     ocp_client_url: >-
-      "https://mirror.openshift.com/pub/openshift-v4/\
-      x86_64/clients/ocp/stable/openshift-client-linux-amd64-rhel\
-      {{ ansible_distribution_major_version }}.tar.gz"
+      {{ "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/" +
+      "stable/openshift-client-linux-amd64-rhel" +
+      ansible_distribution_major_version + ".tar.gz" }}
   ansible.builtin.unarchive:
     src: "{{ ocp_client_url }}"
     dest: "{{ ah_tmp_dir }}"
--- a/roles/acm_hypershift/tasks/download-cli.yml
+++ b/roles/acm_hypershift/tasks/download-cli.yml
@@ -10,6 +10,9 @@
     dest: "{{ ah_tmp_dir }}"
     remote_src: true
     mode: "0755"
+    extra_opts:
+      - oc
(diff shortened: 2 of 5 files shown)


roles/acm_hypershift/tasks/download-cli.yml
tonyskapunk, line 1: I see we are still using stable 4.14 oc, is probably time to update it

betoredhat:
recheck

betoredhat:
retest

roles/acm_hypershift/tasks/download-cli.yml
```diff
@@ -1,9 +1,13 @@
 ---
 - name: Download stable openshift client
   vars:
-    ocp_client_url: "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable-4.14/openshift-client-linux.tar.gz"
+    ocp_client_url: >
+      https://mirror.openshift.com/pub/openshift-v4/
+      x86_64/clients/ocp
```
tonyskapunk, line 6: I like the split, and I think you could include the lines 9 and 10 here as well so you build the client url in a single place.

tonyskapunk:
recheck

roles/acm_hypershift/tasks/download-cli.yml
betoredhat: Thanks

tonyskapunk (commented):
LGTM, thanks!

tonyskapunk (approved):
LGTM, thanks!

roles/acm_hypershift/tasks/download-cli.yml
```diff
@@ -1,7 +1,10 @@
 ---
 - name: Download stable openshift client
   vars:
-    ocp_client_url: "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable-4.14/openshift-client-linux.tar.gz"
+    ocp_client_url: >-
+      {{ "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/" +
+      "stable/openshift-client-linux-amd64-rhel" +
+      ansible_distribution_major_version + ".tar.gz" }}
```
tonyskapunk, line 7: I think it only needed to remove the newline and remove the slash before the version i.e.
```suggestion
    ocp_client_url: >-
      https://mirror.openshift.com/pub/openshift-v4/
      x86_64/clients/ocp/stable/openshift-client-linux-amd64-rhel
      {{ ansible_distribution_major_version }}.tar.gz
```

betoredhat:
retest

roles/acm_hypershift/tasks/delete-cluster.yml
```diff
@@ -15,8 +15,23 @@
       --name={{ ah_cluster_generated_name }}
       --namespace={{ ah_clusters_ns }}
       --infra-id={{ ah_cluster_generated_name }}
-      --cluster-grace-period 0
   changed_when: false
   when:
-    - hosted_cluster.resources | length == 1
+    - hosted_cluster.resources | default([]) | length == 1
+
+- name: Confirm that HCP is deleted
```
ramperher, line 22: I haven't seen this task as tested in the last job, could we have the chance to test this `delete-cluster` play?

roles/acm_hypershift/tasks/download-cli.yml
```diff
@@ -50,4 +56,8 @@
 - name: Set hcp cli path
   ansible.builtin.set_fact:
     ah_hcp_cli_path: "{{ ah_tmp_dir }}/hcp"
+
+- name: Set oc cli path
```
ramperher, line 60: This variable is just defined here in `download-cli`, but this playbook is only called if `ah_download_cli` is true: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/acm_hypershift/tasks/main.yml#L12-L15. Its default value is true, but what happens if it is false? Then the usage of `ah_oc_cli_path` will fail outside this playbook because it's an undefined variable. Should we protect the code for that case?

betoredhat: The assertion during the validation playbook will fail if that is undefined or false. It is covered. The user has the option to pass a specific binary already present on the system. 

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/74bb35eb9f185cbb68f341b9e260de58b7b6da4f/roles/acm_hypershift/tasks/validations.yml#L10

roles/acm_hypershift/tasks/delete-cluster.yml
betoredhat: yeah, just need to run it with a fixed cluster_name. The last jobs were run using kvirt as the name for the hcp cluster, So it will be deleted.

roles/acm_hypershift/tasks/download-cli.yml
betoredhat: My bad, that variable was not asserted. Should be good now. Good catch!

thekad (approved):
LGTM not sure why the bot got confused and didn't report the latest SUCCESS

betoredhat:
retest

PR #386, opened 2024-07-18
Define k8s_best_practices_certsuite role

ramperher:
- Move cnf_cert role logic to new k8s_best_practices_certsuite, to follow the new naming convention of the cnf-certification-test project
- Added a fix in both roles, cnf_cert and k8s_best_practices_certsuite, for retrieving the debug-partner image, based on version.json file, when testing changes from stable releases. This is because, for example, debug-partner image for v5.2.2 is v5.2.1, so it's not true anymore that certsuite version could be used to retrieve the correct debug-partner image

Test-Hints: no-check

Testing:

- [x] certsuite v5.2.2 using legacy cnf_cert role in disconnected mode (to confirm stable versions are still working for legacy role) - https://www.distributed-ci.io/jobs/f34ecd32-f607-4cf2-8f5e-6c77b271e559/jobStates
- [x] certsuite HEAD version using new role in disconnected mode (this is used for testing PRs in cnf-certification-test repo) - https://www.distributed-ci.io/jobs/a3575bcb-4cbd-4f28-b208-1bff41a740bc/jobStates?sort=date
- [x] certsuite v5.2.2 using new role in disconnected mode - https://www.distributed-ci.io/jobs/7ed62949-9ea1-490f-8901-485f43c38bd8/jobStates?sort=date
- [x] certsuite v5.2.2 using new role in connected mode - https://www.distributed-ci.io/jobs/33254a5b-d9ad-40f2-b15f-0d014ffcda07/jobStates?sort=date
- [x] certsuite change (https://github.com/test-network-function/cnf-certification-test/pull/2264) using new role in disconnected mode, to confirm dci-pipeline update is working fine - https://www.distributed-ci.io/jobs/5670930f-1475-4718-8312-c43ce78d15c1/jobStates?sort=date
- [x] example-cnf using certsuite v5.2.2 and new role in disconnected mode - https://www.distributed-ci.io/jobs/71252887-50d2-4ea0-a5ae-d42e407b0184/jobStates/?sort=date

PR #389, opened 2024-07-19
Properly evaluate variables passed to other roles

thekad:
As demonstrated in #386 the ansible-lint custom rule was incorrectly requiring the variable prefixes to match the importing/source role, not the imported/destination role.

Test-Hints: no-check


thekad:
Hurrrrrr I updated the error message so of course that meant the diff was the whole thing. I do like the new message though, because is not only referring to variables within the same role but in imported roles too, WDYT?

tonyskapunk:
retest

tonyskapunk (approved):

PR #293, opened 2024-05-07
Replace openshift-baremetal-install with openshift-install in 4.16+

tonyskapunk:
The baremetal installer (openshift-baremetal-install) drops support to el8 starting on 4.16. Another installer (openshift-install) provides support to it for el8. Doing a drop-in replacement when OCP release is 4.16+

For disconnected environments, when mirror the ocp release is required, extract both installers to be ready during the install.

---

- [x] TestDallas: ocp-4.16-vanilla  :arrow_right:   https://www.distributed-ci.io/jobs/5f87f994-b35b-44aa-afc9-f40a7cdfb5b7
- [x] TestBos2: virt 4.16 :arrow_right: https://www.distributed-ci.io/jobs/4aaa654a-0899-492f-9de8-d5dde40e390f 
- [x] TestBos2: virt 4.15 :arrow_right:  https://www.distributed-ci.io/jobs/4ae86ad0-4970-4b77-bf53-3a69b63815b2
- [x] TestDallas: ocp-4.15-vanilla :arrow_right:   https://www.distributed-ci.io/jobs/7b278fb8-d00a-44b2-abba-6fcb89c19fdf/jobStates
- [x] TestDallas: ocp-4.17-vanilla :arrow_right:  https://www.distributed-ci.io/jobs/94457317-f301-4aeb-90eb-5f8850344792/jobStates
---

TestBos2: virt
Test-Hints: no-check

PR #390, opened 2024-07-22
Add template for PR

tonyskapunk:
Include common templates to ease the testing of the PR

---

Test-Hints: no-check

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,34 @@
+##### SUMMARY
+<!--- Describe the change, including rationale and design decisions -->
+
+<!--- HINT: Include "Fixes #nnn" if you are fixing an existing issue -->
+
+##### ISSUE TYPE
+
+<!--- Pick one below and delete the other: -->
+- Bug, Docs Fix or other nominal change
+- New or Enhanced Feature
+
+##### Tests
+
+<!--- Document the tests for this change, if any -->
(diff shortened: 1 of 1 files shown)


betoredhat (approved):
lgtm, thanks

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+- Bug, Docs Fix or other nominal change
+- New or Enhanced Feature
+
+##### Tests
+
+<!--- Document the tests for this change, if any -->
+<!--- See: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/CONTRIBUTING.md#ci-pipelines -->
+
+<!--- Examples:
+- [] TestDallas - <JobID>
```
ramperher, line 18: Question: wouldn't these lines be interpreted by our automation? Because, as far as I know, dci-pipeline reads the full commit message and looks for the keywords defined in auto.conf, so maybe this would cause that dci-pipeline attempts to launch an undersired job
so, maybe, it's better to avoid using TestXXX when reporting the test results in the checklist

tonyskapunk: > wouldn't these lines be interpreted by our automation?

Not at all, because we do regex to match the beginning of the line.

I've done it multiple times before, e.g. #324, #293, #284

tonyskapunk:
retest

thekad (approved):
LGTM

PR #394, opened 2024-07-30
[ocp_on_libvirt] reduce retries to power off VMs

manurodriguez:
##### SUMMARY

1500 retries with a delay of 10 is too much, 150 retries with a delay of 10 should be enough.
The current config takes up to 4 hours to fail because of the retries.

##### ISSUE TYPE

- Bug

##### Tests

TestBos2: virt


betoredhat (approved):
Time flies! 
:100:

PR #395, opened 2024-07-31
Prevent from creating client tarball if already exists

nsilla:
##### SUMMARY
Fixes: CILAB-1745: Failed at redhatci.ocp.mirror_ocp_release : Create a client tarball for el8 on 4.16

If two Ansible jobs attempt at running the task from redhatci/ocp/roles/mirror_ocp_release/tasks/unpack.yml from the same runner host a conflict will happen since the tar command monitors the target tarball file for external changes. In other words, the first job running the task will eventually detect that the second job overrode the file and fail.

Likewise, there's a conflict between the tasks "Extract installer command from release image" and "Write rhcos.json from command 4.8+" where the first one exports the openshift-install script and the second one executes it so, if one job tries to execute the script while a different one is exporting (and overwriting) it a failure is thrown with the script execution reporting the script is being altered.

With this change, Ansible uses filesystem locks to ensure only one job at a time can create the client tools tarball and extract the openshift installer, while any other job detecting the tasks above are already undergoing, will put on hold and rely on the artifacts generated by the first job.

##### ISSUE TYPE

##### Tests
To be tested with two parallel SNO deployments in a partner lab:

```
$ dci-pipeline-check https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/395 -p sno intel-nm-ocp sno-config &
$ dci-pipeline-check https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/395 -p sno-2 intel-nm-ocp sno-config
```

Following jobs illustrate the fixed used case:
- https://www.distributed-ci.io/jobs/28328ec2-9774-4f34-a22f-67e3bac1b2ae/jobStates?sort=date&task=0bc66810-6ac3-4228-a18b-f2433d21fcec
- https://www.distributed-ci.io/jobs/711ed29c-39c8-41de-9cb5-46543fa41ab8/jobStates?sort=date&task=7ec68d27-7dba-4b60-85d8-89afcfc9205c

TestBos2Sno: sno

### DIFF
--- a/roles/mirror_ocp_release/tasks/facts.yml
+++ b/roles/mirror_ocp_release/tasks/facts.yml
@@ -7,8 +7,9 @@
   when:
     - mor_version is version("4.8", ">=")
   ansible.builtin.shell: >
-    {{ mor_cache_dir }}/{{ mor_version }}/{{ mor_installer }} coreos print-stream-json >
-    {{ mor_cache_dir }}/{{ mor_version }}/rhcos.json
+    flock -x -w 30 "{{ mor_cache_dir }}/{{ mor_version }}/.installer.lock" -c
+    '{{ mor_cache_dir }}/{{ mor_version }}/{{ mor_installer }} coreos print-stream-json >
+    {{ mor_cache_dir }}/{{ mor_version }}/rhcos.json'
   changed_when: true
 
 # TODO: Remove this block when 4.7 is no longer supported
--- a/roles/mirror_ocp_release/tasks/installer.yml
+++ b/roles/mirror_ocp_release/tasks/installer.yml
@@ -10,12 +10,23 @@
 - name: "Extract installer command from release image"
   when:
     - mor_force | bool or not target.stat.exists
-  ansible.builtin.command: >
-    {{ mor_oc }} adm release extract
-    --registry-config={{ mor_auths_file }}
-    --command={{ mor_installer }}
-    --from {{ mor_pull_url }}
-    --to "{{ mor_cache_dir }}/{{ mor_version }}"
+  ansible.builtin.shell: |
+    flock -x -n -E 200 "{{ mor_cache_dir }}/{{ mor_version }}/.installer.lock" -c '
+    {{ mor_oc }} adm release extract \
+    --registry-config="{{ mor_auths_file }}" \
+    --command="{{ mor_installer }}" \
+    --from "{{ mor_pull_url }}" \
+    --to "{{ mor_cache_dir }}/{{ mor_version }}"'
+
+    RESULT="$?"
+
+    if [ "$RESULT" -eq 200 ]
+    then
+      echo "Waiting for lock release"
+      flock -x -w 30 "{{ mor_cache_dir }}/{{ mor_version }}/.installer.lock" -c 'echo "Lock release"'
+    else
+      exit $RESULT
+    fi
   register: extract_cmd
   until: extract_cmd is succeeded
   retries: 9
--- a/roles/mirror_ocp_release/tasks/unpack.yml
+++ b/roles/mirror_ocp_release/tasks/unpack.yml
@@ -50,31 +50,44 @@
       delay: 10
       until: _mor_result is not failed
 
-    - name: Extract el8 client for 4.16+
-      ansible.builtin.command: >
-        {{ _mor_tmp_dir.path }}/oc adm release extract
-        --registry-config={{ mor_auths_file }}
-        --command=oc.rhel8
-        --from {{ mor_pull_url }}
-        --to "{{ mor_cache_dir }}/{{ mor_version }}"
-      register: _mor_extract_cmd
-      until: _mor_extract_cmd is succeeded
-      retries: 9
-      delay: 10
-      changed_when: false
-
     # redhatci.ocp.installer 4.16+ for el8 expects a rhel8 tarball
     # required only for nightly builds
     # using community.general.archive generates the tarball but
     # throws an exception: load_file_common_arguments() unexpected keyword argument 'path'
     - name: Create a client tarball for el8 on 4.16+ # noqa command-instead-of-module
-      ansible.builtin.shell: >
-        tar czf
-        "{{ mor_cache_dir }}/{{ mor_version }}/openshift-client-linux-amd64-rhel8-{{ mor_version }}.tar.gz"
-        -C "{{ mor_cache_dir }}/{{ mor_version }}"
-        oc
-        kubectl
+      ansible.builtin.shell:
+        cmd: |
+          if [ ! -f "{{ mor_cache_dir }}/{{ mor_version }}/openshift-client-linux-amd64-rhel8-{{ mor_version }}.tar.gz" ]
+          then
+            flock -x -n -E 200 {{ mor_cache_dir }}/{{ mor_version }}/.tar.lock -c '
+            {{ _mor_tmp_dir.path }}/oc adm release extract \
+            --registry-config={{ mor_auths_file }} \
+            --command=oc.rhel8 \
+            --from {{ mor_pull_url }} \
+            --to "{{ mor_cache_dir }}/{{ mor_version }}" \
+            && \
+            tar czf \
+            "{{ mor_cache_dir }}/{{ mor_version }}/openshift-client-linux-amd64-rhel8-{{ mor_version }}.tar.gz" \
+            -C "{{ mor_cache_dir }}/{{ mor_version }}" \
+            oc \
+            kubectl
+            '
+
+            RESULT="$?"
+
+            if [ "$RESULT" -eq 200 ]
+            then
+              echo "Waiting for lock release"
+              flock -x -w 30 {{ mor_cache_dir }}/{{ mor_version }}/.tar.lock -c 'echo "Lock released"'
+            else
+              exit $RESULT
+            fi
+          fi
       changed_when: true
+      register: _mor_create_tarball
+      until: _mor_create_tarball is succeeded
+      retries: 9
+      delay: 10
 
     - name: Delete temporary directory for oc client
       ansible.builtin.file:


nsilla:
While testing the change by clearing the target version artifacts and running two parallel jobs I noticed both of them declared the tarball (openshift-client-linux-amd64-rhel8-4.16.5.tar.gz) already existed.

I relaunched the job and then I noticed the tarball is initially created by the task ["Extract artifacts from release image"](https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/fffb978a5635d0a42213e9aa4553be5e4b897e88/roles/mirror_ocp_release/tasks/download.yml#L11).

So the only reason I find to recreate the archive is to replace the kubectl link in the original file by a copy of the oc binary.

nsilla:
I repeated the test of the PR with a nightly build:

https://www.distributed-ci.io/jobs/36bba354-0263-4e95-8699-5251bdcacf71/jobStates
https://www.distributed-ci.io/jobs/c1aea5b3-600b-4d3b-bab9-cb983471c656/jobStates
And I see the same behavior as for the GA version. This is, the rhel8 tarball is initially created with extracting the artifacts from release image, so the "Create a client tarball for rel8 on 4.16" task is just skipped, and the jobs eventually succeed.

So I'm starting to suspect the "create tarball" task was a workaround to make sure the el8 openshift client was correctly built for such environment, but now this is already taken into account when building the release image, so the tasks may not be required anymore.

@tonyskapunk, what do you think?

tonyskapunk:
> So I'm starting to suspect the "create tarball" task was a workaround to make sure the el8 openshift client was correctly built for such environment, but now this is already taken into account when building the release image, so the tasks may not be required anymore.

Is not a workaround per se, but it is to have the same behavior that would happen with a non-nightly build (9048e432a4a066047b57f80c220ef7223e1e3489). And yes, this is only for nightly builds

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/9dcf5ce3e35ff323d59e47a87737003682c6ffe0/roles/mirror_ocp_release/tasks/unpack.yml#L66-L67

You're right, the module internally checks the file already exists and skips its action, despite the ansible task is `changed`.

---

Now, back to the origin of the problem. This change will not prevent it from happening because the problem is when the first tarball is created and another process is extracting the clients (oc, kubectl) that are used for the tarball. This race-condition is the problem.

A common solution for this type of issues is to use a lockfile, i.e. create the lockfile *before* extraction and remove it *after* successfully extracted. The creation of the tarball then should verify first the lockfile is not in place before creating the tarball. Feel free to give it a shot.

Let me know if I can be of any help. Thanks!

roles/mirror_ocp_release/tasks/unpack.yml
```diff
@@ -67,14 +67,41 @@
...
+                echo "Lock acquired"
+
+                tar czf \
+                "{{ mor_cache_dir }}/{{ mor_version }}/openshift-client-linux-amd64-rhel8-{{ mor_version }}.tar.gz" \
+                -C "{{ mor_cache_dir }}/{{ mor_version }}" \
+                oc \
+                kubectl
+              ) 200>{{ mor_cache_dir }}/{{ mor_version }}/tar.lock
+
+              rm -f {{ mor_cache_dir }}/{{ mor_version }}/tar.lock
```
tkrishtop, line 86: The removal of the lock file `rm -f {{ mor_cache_dir }}/{{ mor_version }}/tar.lock` after the tarball creation is unnecessary within the block because flock automatically releases the lock when the script exits.

tkrishtop (dismissed):
LGTM just one minor comment to improve

roles/mirror_ocp_release/tasks/unpack.yml
nsilla: Yup. I just remove it so we don't keep "garbage" files.

Do you think it's a better idea just to leave the file there?

tkrishtop: I believe it's automatically removed by flock

ramperher:
@nsilla , since this is a recurrent issue in our daily jobs in BOS2, I wanted to know what's the current status of the change and if it's ready for testing and review. Thanks!

roles/mirror_ocp_release/tasks/unpack.yml
nsilla: I run different tests but in all of them I see the lock is kept.
Also, I'm using it as a mean to have the second job blocked, and waiting for the first one to complete.
The second option would be to have a script looping and trying to get the lock again, but this way I can do it using a builtin module (file) instead of a script.

tkrishtop (approved):
LGTM, thank you @nsilla merging to check on tomorrow's daily

tonyskapunk (changes requested):
I'm sorry, it does not seem like this is working the way we are expecting, see my comments and let's talk if they are not clear.

tkrishtop:
> I'm sorry, it does not seem like this is working the way we are expecting, see my comments and let's talk if they are not clear.

no worries @tonyskapunk, thank you for checking, we could review again tomorrow

roles/mirror_ocp_release/tasks/facts.yml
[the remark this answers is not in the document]
nsilla: Hi Tony, I'm going to revert the changes related to the rhcos.json file extraction because taking a closer look to the failed task output I'd say the problem in this case is not caused by two jobs running the same task simultaneously, but rather with one job executing the openshift-install script while the other job is trying to overwrite the script.

I'll try to reproduce this other scenario, but in the meantime I'd rather have the current fix merged not to keep the daily jobs in BOS2 failing.

tkrishtop:
@nsilla there is still an issue with ansible-lint

> -ERROR: roles/mirror_ocp_release/tasks/unpack.yml:: yaml[trailing-spaces]: Trailing spaces

roles/mirror_ocp_release/tasks/facts.yml
```diff
@@ -7,8 +7,9 @@
   when:
     - mor_version is version("4.8", ">=")
   ansible.builtin.shell: >
-    {{ mor_cache_dir }}/{{ mor_version }}/{{ mor_installer }} coreos print-stream-json >
-    {{ mor_cache_dir }}/{{ mor_version }}/rhcos.json
+    flock -x -w 30 "{{ mor_cache_dir }}/{{ mor_version }}/installer.lock" -c
```
ramperher, line 10: So, are we finally cover the case of rhcos.json file here? From previous commands, I was supposing that this would not be covered

nsilla: yes, we'll cover it. I was able to reproduce the issue and identified the two tasks that are conflicting.
They are the "Extract installer command from release image" and "Write rhcos.json from command 4.8+" tasks.
I'll update the PR description as Tony requested in the first place.

roles/mirror_ocp_release/tasks/installer.yml
```diff
@@ -10,12 +10,19 @@
...
+    --registry-config="{{ mor_auths_file }}" \
+    --command="{{ mor_installer }}" \
+    --from "{{ mor_pull_url }}" \
+    --to "{{ mor_cache_dir }}/{{ mor_version }}"'
+
+    if [ "$?" -eq 200 ]
+    then
+      echo "Waiting for lock release"
+      flock -x -w 30 "{{ mor_cache_dir }}/{{ mor_version }}/installer.lock" -c 'echo "Lock release"'
+    fi
```
tonyskapunk, line 25: The new logic introduces a problem with the retries, these are never done because the result is always successful. When there is a failure in the task inside flock, e.g. a timeout, the task is not retried.

nsilla: Good point!
Let's see if this new version is better.

nsilla:
Although one of the jobs fails at a later task, these two jobs reproduce the concurrent task use case in the "Extract installer command from the release image" task:
- https://www.distributed-ci.io/jobs/711ed29c-39c8-41de-9cb5-46543fa41ab8/jobStates?sort=date&task=7b4b208e-c2b3-4484-bdc2-8e16f21e2028
- https://www.distributed-ci.io/jobs/28328ec2-9774-4f34-a22f-67e3bac1b2ae/jobStates?sort=date&task=0bc66810-6ac3-4228-a18b-f2433d21fcec

tonyskapunk (changes requested):
Thanks for the progress, there is a couple of changes that are still needed, and the rests are optional.

roles/mirror_ocp_release/tasks/facts.yml
```diff
@@ -7,8 +7,9 @@
   when:
     - mor_version is version("4.8", ">=")
   ansible.builtin.shell: >
-    {{ mor_cache_dir }}/{{ mor_version }}/{{ mor_installer }} coreos print-stream-json >
-    {{ mor_cache_dir }}/{{ mor_version }}/rhcos.json
+    flock -x -w 30 "{{ mor_cache_dir }}/{{ mor_version }}/.installer.lock" -c
```
nsilla, line 10: @tonyskapunk, I accepted your commit but then I realized this lock and the one for the installer must be the same one, because in the end there is this interdependency.

If the openshift-install binary is being extracted, we have to wait in order to use it to export the rhcos.json file, thus the use of the same lock.

tonyskapunk: I see, but how does this task identifies between waiting for another task running `mor_installer coreos print-stream.json` from the extraction of the installer?

I believe it should be separated, i.e. if a job has reached this task, then the `mor_installer ` must exist already, and anything attempting to re-write it should not thanks to the other file lock mechanism in installer.yml.

That's why I think this locking mechanism must be different and focused on avoiding multiple jobs writing the same rhcos.json., wydt?

nsilla: As we discussed online, the thing is that:

1) a job trying to extract the installer while there's another job already doing so will wait for the installer to be extracted and move on to the next task.
2) a job trying to generator the rhcos.json file while there's another job extracting the installer will wait for the installer to be extracted and then generate the rhcos.json file.
3) a job trying to extract the installer while there's another job generating the rhcos.json file (this means the installer is already extracted), will wait for the rhcos to be generated and then move on to the next task.
4) a job trying to generate the rhcos.json while other job is generating it, will wait for the first job to finish generating and then generate it itself. The resulting file will be the same one but this is acceptable.

So, if I'm not missing anything, we should be safe with this solution.

nsilla:
With the current version of this change, here we have an example of a job waiting for the other one to extract the installer:

[extracting the installer](https://www.distributed-ci.io/jobs/7853ac35-7f47-484a-aa14-0ae7feda94b6/jobStates?sort=date&task=05578f68-c658-438d-a75f-3da5acc16468)

[waiting](https://www.distributed-ci.io/jobs/c7258b52-b87f-4d3b-a4f2-73e4ea470f4e/jobStates?sort=date&task=8a1c7bcf-53fc-44f6-bf02-955660a5d42e)

tonyskapunk (approved):
Thanks for the work in here @nsilla

PR #408, opened 2024-08-16
Patch provisioning if available

betoredhat:
##### SUMMARY

- Allow ACM hubs setup on SNO instances
- Provisioning resource is not present in SNO deployments

##### ISSUE TYPE

- Fix

##### Tests
Test-Hint: no-check

### DIFF
--- a/roles/acm_setup/tasks/main.yml
+++ b/roles/acm_setup/tasks/main.yml
@@ -196,7 +196,7 @@
       spec:
         watchAllNamespaces: true
   when:
-    - _as_prov_conf.resources | default([0]) | length == 1
+    - _as_prov_conf.resources | default([]) | length == 1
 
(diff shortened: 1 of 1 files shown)


roles/acm_setup/tasks/main.yml
```diff
@@ -188,6 +195,8 @@
         name: provisioning-configuration
       spec:
         watchAllNamespaces: true
+  when:
+    - _as_prov_conf.resources | default([0]) | length == 1
```
tonyskapunk, line 199: This condition is met when there are no `Provisioning` resources, did you mean:
```suggestion
    - _as_prov_conf.resources | default([]) | length == 1
```

tonyskapunk (approved):
LGTM

PR #412, opened 2024-08-21
Supporting concurrent jobs in the SNO provision host

nsilla:
##### SUMMARY
Fixes CILAB-1776: Failed at redhatci.ocp.sno_installer : Load variables from rhcos.jso

> The rhcos.json file is first downloaded and placed in the /tmp directory, then processed, and finally removed.
> 
> When running two simultaneous jobs, it may happen that the first job removes the file just after the second one pulls it but before it processes it's contents, thus the error.

##### ISSUE TYPE

- Bug

##### Tests

TestBos2Sno: sno

### DIFF
--- a/roles/sno_installer/tasks/21_rhcos_image_paths.yml
+++ b/roles/sno_installer/tasks/21_rhcos_image_paths.yml
@@ -25,20 +25,14 @@
 
 - name: "Get the rhcos.json from cache server into temp file on control machine"
   ansible.builtin.fetch:
-    dest: /tmp/
+    dest: "{{ _si_tmp_dir.path }}/"
     flat: true
     src: "{{ cache_dir }}/{{ version }}/rhcos.json"
   delegate_to: "{{ cache_server }}"
 
 - name: "Load variables from rhcos.json"
   ansible.builtin.include_vars:
-    file: "/tmp/rhcos.json"
+    file: "{{ _si_tmp_dir.path }}/rhcos.json"
     name: rhcos_json_vars
   tags: rhcospath
-
-- name: "Remove the temporary copy of rhcos.json on control machine"
-  ansible.builtin.file:
-    path: "/tmp/rhcos.json"
-    state: absent
-  delegate_to: localhost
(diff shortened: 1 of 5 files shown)


roles/sno_installer/tasks/21_rhcos_image_paths.yml
```diff
@@ -25,20 +25,14 @@
...
+    dest: "{{ _si_tmp_dir.path }}"
     flat: true
     src: "{{ cache_dir }}/{{ version }}/rhcos.json"
   delegate_to: "{{ cache_server }}"
 
 - name: "Load variables from rhcos.json"
   ansible.builtin.include_vars:
-    file: "/tmp/rhcos.json"
+    file: "{{ _si_tmp_dir.path }}/rhcos.json"
     name: rhcos_json_vars
```
manurodriguez, line 36: Looks like the last job failed here because we are missing a `delegate_to`, I recognise there are a few tasks that are delegated, some to localhost and some to cache_server, which makes it confusing, so we probable need to review if all that back and forth is actually needed.
 
https://www.distributed-ci.io/jobs/ff0c6f78-7804-45ab-b814-a54793dd9405/jobStates?sort=date&task=06a5b2ad-f4b3-4992-9a92-e16540f74166

nsilla: I have a different understanding here.
The missing delegate_to is not the problem.
In the previous task, we use the fetch module to pull the rhcos.json file from the cache_server when the play target host is the provisioner.
By setting the delegate to to the cache_server we're not forcing the file to be copied to the cache_server but to be pulled from the cache_server.
The problem, I believe, is the fetch task is missing a trailing "/" in the destination.
I'll update the code accordingly.

manurodriguez: Thanks for catching that

manurodriguez (approved):
LGTM

nsilla:
retest

PR #419, opened 2024-09-03
Do not fail if skippable is true

betoredhat:
##### SUMMARY
It allows to skip operators that may not be present in the catalog, but it is acceptable not to try installing them.

##### ISSUE TYPE

- New or Enhanced Feature

##### Tests

- [x] TestBos2 - https://www.distributed-ci.io/jobs/6552213a-2739-4440-bd92-4a84184f7dce/jobStates

---

Test-Hint: no-check

### DIFF
--- a/roles/olm_operator/tasks/main.yml
+++ b/roles/olm_operator/tasks/main.yml
@@ -55,9 +55,24 @@
     - operator_packagemanifest is defined
     - operator_packagemanifest.resources is defined
     - operator_packagemanifest.resources | length
-  failed_when: operator_packagemanifest is skipped
+  failed_when: false
+  ignore_errors: true
+
+- name: Set package as not found
+  ansible.builtin.set_fact:
+    olm_operator_found: false
+  when: operator_packagemanifest is skipped
     or operator_packagemanifest.resources is undefined
     or operator_packagemanifest.resources | length != 1
+
+- name: Fail if not found and nor skippable
+  ansible.builtin.fail:
+    msg: >
+      The operator {{ operator }} is not found in the catalog
+      and skippable is {{ olm_operator_skippable }}.
+  when:
+    - not olm_operator_skippable
+    - not olm_operator_found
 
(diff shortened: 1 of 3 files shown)


roles/olm_operator/tasks/main.yml
```diff
@@ -55,10 +55,23 @@
...
     or operator_packagemanifest.resources is undefined
     or operator_packagemanifest.resources | length != 1
 
+- name: Fail if not found and nor skippable
+  ansible.builtin.fail:
+    msg: >
+      The operator {{ operator }} is not found in the catalog
+      and skippable is {{ olm_operator_skippable }}.
+  when:
+    - olm_operator_skippable is false
```
manurodriguez, line 72: Beto, here shouldn't be something like this:
```suggestion
    - not olm_operator_skippable | bool
```
I haven't used an expression like you did, I wonder if it would be treated as a string and not as a boolean?

roles/olm_operator/tasks/main.yml
```diff
@@ -55,10 +55,25 @@
     - operator_packagemanifest is defined
     - operator_packagemanifest.resources is defined
     - operator_packagemanifest.resources | length
-  failed_when: operator_packagemanifest is skipped
+  failed_when: false
+  ignore_errors: true
```
tkrishtop, line 59: do we really need failed_when and ignore_errors together? Maybe `ignore_errors` is enough?

tkrishtop (approved):
LGTM just one minor note

roles/olm_operator/tasks/main.yml
betoredhat: > do we really need failed_when and ignore_errors together? Maybe `ignore_errors` is enough?

Yeah, hehe I was attempting an ignore retry :) Thanks!

tkrishtop (approved):
LGTM

betoredhat:
recheck

ramperher (approved):
LGTM

PR #420, opened 2024-09-04
Added support for Git HTTP transport

nsilla:
##### SUMMARY
This change updates the gitops_configure_repo role to add support for Git HTTP transport.
This is needed because the role currently only supports SSH transport, but HTTP is easier to set up in repositories hosted in an OpenShift cluster, for instance, in disconnected networks

##### ISSUE TYPE

- Enhanced Feature

##### Tests

Test-Hints: no-check

### DIFF
--- a/roles/gitops_configure_repo/tasks/main.yaml
+++ b/roles/gitops_configure_repo/tasks/main.yaml
@@ -1,12 +1,19 @@
 ---
 
+- name: Identify the Git repo transport protocol
+  ansible.builtin.set_fact:
+    gcr_repo_protocol_http: "{{ gcr_ztp_gitops_repo | regex_search('https?://') | type_debug == 'str' }}"
+
 - name: Assert the required variables are defined
   ansible.builtin.assert:
     that:
-      - gcr_ssh_key_path is defined
-      - gcr_ssh_key_path | length > 0
       - gcr_ztp_gitops_repo is defined
       - gcr_ztp_gitops_repo | length > 0
+      - >
+          (gcr_repo_protocol_http and
+          gcr_repo_username is defined and gcr_repo_username | length > 0 and
+          gcr_repo_password is defined and gcr_repo_password | length > 0) or
+          (gcr_ssh_key_path is defined and gcr_ssh_key_path | length > 0)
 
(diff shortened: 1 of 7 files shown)


roles/gitops_configure_repo/tasks/main.yaml
```diff
@@ -1,12 +1,19 @@
...
     that:
-      - gcr_ssh_key_path is defined
-      - gcr_ssh_key_path | length > 0
       - gcr_ztp_gitops_repo is defined
       - gcr_ztp_gitops_repo | length > 0
+      - >
+          (_gcr_repo_protocol_http and
+          gcr_repo_username is defined and gcr_repo_username | length > 0 and
+          gcr_repo_password is defined and gcr_repo_password | length > 0) or
+          (gcr_ztp_gitops_repo gcr_ssh_key_path is defined and gcr_ssh_key_path | length > 0)
```
ramperher, line 16: I think you have here a `gcr_ztp_gitops_repo` that has to be removed, should be `(gcr_ssh_key_path is defined and gcr_ssh_key_path | length > 0)
`

roles/gitops_configure_repo/templates/http_private_repo.yaml.j2
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    argocd.argoproj.io/secret-type: repository
+  name: {{ gcr_private_repo_secret }}
+  namespace: openshift-gitops
+stringData:
+  project: default
+  type: git
+  url: {{ gcr_ztp_gitops_repo }}
+  username: {{ gcr_repo_username }}
+  password: {{ gcr_repo_password }}
```
ramperher, line 14: Missed line break

roles/gitops_configure_repo/templates/http_private_repo.yaml.j2
ramperher, line 1: I suppose you're still working on integrating this to deploy it in the role.

nsilla: Yes, this is still a work in progress.
I still need to figure out some issues regarding the disconnected environments.

ramperher (approved):
LGTM

PR #434, opened 2024-09-11
Extracting RHEL8 OC client even if related tar.gz file already exists

nsilla:
##### SUMMARY
Fixes CILAB-445

The mirror_ocp_release role runs some tasks over tar.gz archives that we susceptible of fail when multiple jobs were operating over the same files simultaneously.

To overcome this, we introduced some file locks to secure the pieces of code that could cause this conflicts.

Along with this, we introduced an initial check so, if the resulting tar.gz file already existed the code wouldn't run.

This introduced a bug since, as part of the generation of the openshift-client tar.gz file for RHEL8, the RHEL8 OC binary was extracted in the cache directory to replace the existing RHEL9 binary.

If this is not done, running the RHEL9 binary in a later task will fail due to missing glibc libraries.

- Bug

- [x] TestDallas: ocp-4.17-vanilla -  https://www.distributed-ci.io/jobs/acbfe98e-1664-4486-b014-f650626d6310/jobStates

---

Test-Hints: no-check

### DIFF
--- a/roles/mirror_ocp_release/tasks/unpack.yml
+++ b/roles/mirror_ocp_release/tasks/unpack.yml
@@ -57,31 +57,28 @@
     - name: Create a client tarball for el8 on 4.16+ # noqa command-instead-of-module
       ansible.builtin.shell:
         cmd: |
-          if [ ! -f "{{ mor_cache_dir }}/{{ mor_version }}/openshift-client-linux-amd64-rhel8-{{ mor_version }}.tar.gz" ]
+          flock -x -n -E 200 {{ mor_cache_dir }}/{{ mor_version }}/.tar.lock -c '
+          {{ _mor_tmp_dir.path }}/oc adm release extract \
+          --registry-config={{ mor_auths_file }} \
+          --command=oc.rhel8 \
+          --from {{ mor_pull_url }} \
+          --to "{{ mor_cache_dir }}/{{ mor_version }}" \
+          && \
+          tar czf \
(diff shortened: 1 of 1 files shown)


ramperher:
Change should be good, let's wait for the debug job.

ramperher (approved):
I can see the file is now generated fine, even though the binaries are present in the cache folder:

```
-rwxr-xr-x.  1 dciteam dciteam  65560322 Sep 11 05:53 openshift-client-linux-amd64-rhel8-4.17.0-0.nightly-2024-09-09-120947.tar.gz
```

And now the oc binary is working fine: https://www.distributed-ci.io/jobs/acbfe98e-1664-4486-b014-f650626d6310/jobStates?sort=date&task=6aff7919-1789-4966-831d-1debfc93029c

CC @tonyskapunk to confirm everything is fine, but from my side, LGTM

tonyskapunk (approved):
The way I see this fix is that it will always re-generate the tarball but it does not prevent the extraction of the clients the very first time it runs with the cache enabled.

I created this PR https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/435 to test this. My understanding looking at the [first job that failed]( https://www.distributed-ci.io/jobs/a582aada-7379-48b2-8f32-a486d2427606/jobStates?sort=date&task=8248764c-a13b-47c7-ad94-332e5e1441a3) is that the extraction of the client happened twice and it should only occur once.

It is still unclear to me why the tarball for rhel8 already existed but the `oc` was not the one extracted in the same task, i.e. `oc-rhel8`

All in all, the change LGTM

PR #438, opened 2024-09-18
Remove certsuite backwards compatibility for versions below v5.2.3

ramperher:
##### SUMMARY

Currently, we support backwards compatibility until v5.1.3.

We'd like to remove support for versions below v5.2.3, which is the version where certsuite name started to be used instead of cnf-certification-test.

##### ISSUE TYPE

- Enhanced Feature

##### Tests

- [x] TestDallasWorkload - https://www.distributed-ci.io/jobs/bab3e32e-8590-43e6-ba54-c8076610a60c/jobStates

Test-Hints: no-check


manurodriguez (approved):
LGTM

PR #442, opened 2024-09-23
sno_installer: allow for external cache server even in connected environments

nsilla:
##### SUMMARY
The SNO Installer role delegates the OCP installation artifacts to the host set in the cache_server variable.

The role either sets this variable to the provision host, if the environment is connected, or to a host identified as the registry_host in the inventory in case of disconnected networks.

This change allows to provide the role with an explicitly identified cache server, even in connected environments, in order to save resources by sharing the cache server between different deployment types.

##### ISSUE TYPE

- Enhanced feature

##### Tests

TestBos2Sno: sno

### DIFF
--- a/roles/sno_installer/vars/main.yml
+++ b/roles/sno_installer/vars/main.yml
@@ -4,3 +4,5 @@ pullsecret_file: "{{ si_cache_dir }}/pull-secret.txt"
 default_libvirt_pool_dir: "/var/lib/libvirt/images"
 force_mirroring: false
 ocp_binary_path: "/usr/local/bin"
+
+snp_cache_dir: "{{ si_cache_dir }}"
--- a/roles/sno_installer/vars/main.yml
+++ b/roles/sno_installer/vars/main.yml
@@ -1,7 +1,6 @@
 ---
 # vars file for sno_installer
-cache_dir: "/opt/cache"
-pullsecret_file: "{{ cache_dir }}/pull-secret.txt"
+pullsecret_file: "{{ si_cache_dir }}/pull-secret.txt"
 default_libvirt_pool_dir: "/var/lib/libvirt/images"
 force_mirroring: false
 ocp_binary_path: "/usr/local/bin"
--- a/roles/sno_node_prep/tasks/10_validation.yml
+++ b/roles/sno_node_prep/tasks/10_validation.yml
@@ -161,7 +161,7 @@
 
 - name: "Ensure cache_dir is created"
   ansible.builtin.file:
-    path: "{{ cache_dir }}"
+    path: "{{ snp_cache_dir }}"
     state: directory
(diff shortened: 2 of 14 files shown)


manurodriguez (approved):
LGTM

roles/sno_installer/vars/main.yml
```diff
@@ -1,7 +1,8 @@
 ---
 # vars file for sno_installer
-cache_dir: "/opt/cache"
-pullsecret_file: "{{ cache_dir }}/pull-secret.txt"
+pullsecret_file: "{{ si_cache_dir }}/pull-secret.txt"
 default_libvirt_pool_dir: "/var/lib/libvirt/images"
 force_mirroring: false
 ocp_binary_path: "/usr/local/bin"
+
+snp_cache_dir: "{{ si_cache_dir }}"
```
ramperher, line 8: Is this variable correctly defined? Because the role is `sno_installer`, but prefix is `snp` (`sno_node_prep`?), so looks like used in a different role

roles/sno_node_prep/tasks/10_validation.yml
```diff
@@ -161,7 +161,7 @@
 
 - name: "Ensure cache_dir is created"
   ansible.builtin.file:
-    path: "{{ cache_dir }}"
+    path: "{{ snp_cache_dir }}"
```
ramperher, line 164: as I said, def value of this variable is defined in a different role, they should be independent

roles/sno_node_prep/vars/main.yml
```diff
@@ -25,4 +25,4 @@ snobm_pkg_list:
   - shim-x64
   - dnsmasq
 
-pullsecret_file: "{{ cache_dir }}/pull-secret.txt"
+pullsecret_file: "{{ snp_cache_dir }}/pull-secret.txt"
```
ramperher, line 28: same

ramperher (commented):
Linting needs to be fixed. Apart from this, some comments from my side.

roles/sno_installer/vars/main.yml
nsilla: Yes, the name was intended.

This is something I noticed when adding the "si" prefix to the variables I'm adding or modifying.

I was wondering why some variables were defined in the "vars" directory instead of the "defaults" and the problem is that some variables defined in the sno_installer role are used by sno_node_prep, but for the later to have access to the variables the former must define them in the "vars" directory.

So "si_cache_dir" is a default for the sno_installer role, and "snp_cache_dir" is an input variable for the sno_node_prep role.

Now that I have a working version of the PR, I'll see to simplify and optimize it, so it may be I'll remove the variable after all.

pierreblanc (approved):
Lgtm

PR #444, opened 2024-09-25
Create admin user in oc_setup role

tonyskapunk:
##### SUMMARY

The previous admin user in the oc_setup rule was created with a cluster-admin role. Because there is an admin role, we are now creating three users: admin, cluster_admin, nonadmin.

nonadmin remains without change, this user has no role associated, so nothing can be done with it, other than authenticate.

cluster_admin, is a new user and is now assigned the cluster-admin role, previoulsy this was the role assigned to the admin user.

admin, this user now has the admin role, it has less priviledges as before when it had cluster-admin role.

##### ISSUE TYPE

- New or Enhanced Feature

##### Tests

- [x] TestBos2Sno - https://www.distributed-ci.io/jobs/9f175cc8-ae49-4b57-a5e7-b31058b482a8

---

Test-Hints: no-check

### DIFF
--- a/ansible-collection-redhatci-ocp.spec
+++ b/ansible-collection-redhatci-ocp.spec
@@ -3,7 +3,7 @@
 %global forgeurl https://github.com/%{org}/%{repo}
 
 Name:           %{repo}
-Version:        0.20.EPOCH
+Version:        0.21.EPOCH
 Release:        VERS%{?dist}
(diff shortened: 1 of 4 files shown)


betoredhat (approved):
LGTM, thanks

betoredhat:
I confirmed that the credentials work. I think having this validated in Connected is ok.

tonyskapunk:
retest

PR #446, opened 2024-09-26
expand oc_setup role to set MC with SSH pub keys

manurodriguez:
##### SUMMARY
This will allow to specify multiple SSH key paths
to be applied at day2 and SSH can be performed
from multiple sources.

##### ISSUE TYPE

- Enhanced Feature in an existing role

##### Tests

- [] TestDallas - <JobID>
- [] TestBos2: - <JobID>

TestBos2: virt

### DIFF
--- a/roles/oc_setup/tasks/main.yml
+++ b/roles/oc_setup/tasks/main.yml
@@ -105,4 +105,52 @@
       subjects:
         - kind: User
           name: "admin"
+
+- name: Add SSH keys to cluster nodes at day2
+  when:
+    - oc_ssh_extra_keys_paths is defined
+    - oc_ssh_extra_keys_paths | length > 0
+  block:
+    - name: Check if SSH keys exist
+      ansible.builtin.stat:
+        path: "{{ item }}"
+      loop: "{{ oc_ssh_extra_keys_paths }}"
+      register: _os_ssh_path
+
+    - name: "Fail if any ssh path does not exist"
+      ansible.builtin.fail:
+        msg: "SSH public key {{ key.item }} file does not exist"
+      when: not key.stat.exists
+      loop: "{{ _os_ssh_path.results }}"
+      loop_control:
+        loop_var: key
+        label: "{{ key.item }}"
+
+    - name: Set SSH Keys manifests in master group
+      vars:
+        os_role: master
+      community.kubernetes.k8s:
+        state: present
+        definition: "{{ lookup('template', '99_ssh_mc.yml.j2') }}"
+
+    - name: Set SSH Keys manifests in worker group
+      vars:
+        os_role: worker
+      community.kubernetes.k8s:
+        state: present
+        definition: "{{ lookup('template', '99_ssh_mc.yml.j2') }}"
+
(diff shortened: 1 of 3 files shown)


roles/oc_setup/templates/99_ssh_mc.yml.j2
```diff
@@ -0,0 +1,22 @@
...
+  config:
+    ignition:
+      version: 3.2.0
+    passwd:
+      users:
+      - name: core
+        sshAuthorizedKeys:
+  {% for ssh_key_path in oc_ssh_extra_keys_paths %}
+      - '{{ lookup('file', ssh_key_path ) }}'
+  {% endfor %}
```
tonyskapunk, line 17: Is the `~` expanded as expected? iirc, with ansible is not possible, I think it's needed to be replaced with the `HOME` env var, e.g.

```suggestion
  {% for ssh_key_path in oc_ssh_extra_keys_paths %}
      - '{{ lookup('file', ssh_key_path | regex_replace('~', lookup('env', 'HOME'))) }}'
  {% endfor %}

File: roles/oc_setup/tasks/main.yml [hunk 2]
```diff
@@ -105,4 +105,52 @@
...
           name: "admin"
+
+- name: Add SSH keys to cluster nodes at day2
+  when:
+    - oc_ssh_extra_keys_paths is defined
+    - oc_ssh_extra_keys_paths | length > 0
+  block:
+    - name: Check if SSH keys exist
+      ansible.builtin.stat:
+        path: "{{ item }}"
```
tonyskapunk, line 116: See my comment regarding tilde in the template

roles/oc_setup/tasks/main.yml
```diff
@@ -105,4 +105,52 @@
...
+    - name: "Fail if any ssh path does not exist"
+      ansible.builtin.fail:
+        msg: "SSH public key {{ key.item }} file does not exist"
+      when: not key.stat.exists
+      loop: "{{ _os_ssh_path.results }}"
+      loop_control:
+        loop_var: key
+        label: "{{ key.item }}"
+
+    - name: Set SSH Keys manifests in master group
```
betoredhat, line 129: Maybe we can check the available machine pools and loop through them if we have more than masters and workers.

```
    - name: Retrieve MachineConfigPool resources
      community.kubernetes.k8s_info:
        api_version: machineconfiguration.openshift.io/v1
        kind: MachineConfigPool
      register: mcp_info

    - name: Extract MachineConfigPool Names
      set_fact:
        mcp_names: "{{ mcp_info.resources | map(attribute='metadata.name') | list }}"
```

manurodriguez:
@betoredhat @tonyskapunk appreciate the feedback, all good points, I took some of that into account in a new PR https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/448 to see if we can do this at day1 after our discussion. We'll see

manurodriguez:
closing, we preferred the option to inject SSH keys at day1 #448 in IPI jobs, we might do the day in the future if is needed

PR #452, opened 2024-10-02
Improve Subs/CSV validation

betoredhat:
##### SUMMARY

- installPlanApproval for the initial subscription must be Automatic to complete the installation. 
- Use installedCSV to validate the subscription as it represents the current version installed. currentCSV represents the version available for upgrading.

##### ISSUE TYPE

##### Tests

- [x] TestBos2 - https://www.distributed-ci.io/jobs/d334c5cf-05f1-441b-9925-ec4a2738c625/jobStates

### DIFF
--- a/roles/olm_operator/tasks/main.yml
+++ b/roles/olm_operator/tasks/main.yml
@@ -130,7 +130,7 @@
             namespace: "{{ namespace }}"
           spec:
             channel: "{{ desired_channel }}"
-            installPlanApproval: "{{ starting_csv | default('') | length | ternary('Manual', 'Automatic') }}"
+            installPlanApproval: "Automatic"
             config:
               resources: {}
             name: "{{ operator }}"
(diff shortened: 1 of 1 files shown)


tonyskapunk (approved):
LGTM!! :100:

nsilla:
I'd wish I had seen this PR before it got approved but I see it's causing some jobs to fail.

https://www.distributed-ci.io/jobs/1fd1de75-5ecf-4a0e-aa7f-73549c216797/jobStates?sort=date&task=dd157386-954d-41eb-b1a5-4be22f6b420c

The problem is hardcoding the approval method to "Automatic" is not compatible with using the starting_csv parameter.

With the automatic approval method, the OLM will install the starting_csv and then keep upgrading the CSVs until it reaches the currentCSV.

To start with, this renders the starting_csv parameter without effect, but, furthermore, when the olm_operator role waits for the subscription to be ready, the condition is that the installedCSV has the same value as the operator_csv variable, which is set to the starting_csv if set. As you can see in the linked job, with the automatic approval method these values are bound not to match and thus the job will fail.

betoredhat:
ouch, no worries... time travel is now possible. The following revert change will do the trick: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/454. Will wait for you next time. :)

PR #464, opened 2024-10-11
Fail preflight release action if the retrieved release is null

tkrishtop:
##### SUMMARY

This PR is to force preflight GH action to fail when the retrieved release is `null` because of the GH API glitch.

##### ISSUE TYPE

Current behavior:
- [GH action](https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/11288743625/job/31397141913) has retrieved the release being null because of an API glitch
```
0s
Run preflight_latest=$(curl -s https://api.github.com/repos/redhat-openshift-ecosystem/openshift-preflight/releases/latest | jq -r .tag_name)
  preflight_latest=$(curl -s https://api.github.com/repos/redhat-openshift-ecosystem/openshift-preflight/releases/latest | jq -r .tag_name)
  echo "Preflight latest release: $preflight_latest"
  echo "preflight_latest=$preflight_latest" >> $GITHUB_OUTPUT
  shell: /usr/bin/bash -e {0}
Preflight latest release: null
```
while the real latest release is not null but 1.10.1

```
$ curl -s https://api.github.com/repos/redhat-openshift-ecosystem/openshift-preflight/releases/latest | jq -r .tag_name
1.10.1
```

- GH action opens weird PR [Move to Preflight release null](https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/463).

The idea for the fix is to fail pipeline asap instead of opening the PR.

Test-Hint: no-check

PR #465, opened 2024-10-11
Use GHA to get latest release

tonyskapunk:
##### SUMMARY

Use a GHA to rely on the latest version of a repo.
Should prevent issues like #463 

##### ISSUE TYPE

- Bug

##### Tests

Test-Hints: no-check

### DIFF
--- a/.github/workflows/preflight-release.yml
+++ b/.github/workflows/preflight-release.yml
@@ -12,24 +12,25 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Check for new release
-        id: check_release
-        run: |
-          preflight_latest=$(curl -s https://api.github.com/repos/redhat-openshift-ecosystem/openshift-preflight/releases/latest | jq -r .tag_name)
-          echo "Preflight latest release: $preflight_latest"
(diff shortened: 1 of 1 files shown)


tkrishtop:
> @tkrishtop My bad, didn't realize about #464 was opened for this. I'll close this.
> 
> Feel free to reopen if you think this is valuable.

Hi @tonyskapunk I like your solution better because you use a ready GH module instead of my homemade patch. Let's reopen your PR and close mine. 

Does `pozetroninc/github-action-get-latest-release@v0.8.0` ensure the return to be non-null already?  Or should we add an explicit check like in my PR? 
```
if [ "$preflight_latest" == "null" ]; then
  echo "Failed to retrieve the latest release. Exiting."
  exit 1
fi
```

tonyskapunk:
> Does `pozetroninc/github-action-get-latest-release@v0.8.0` ensure the return to be non-null already? Or should we add an explicit check like in my PR?

I don't have a definitive answer for this, but I believe the step will fail, so no PR will be created.

tkrishtop (approved):
lgtm

thekad (approved):
yep this LGTM

PR #472, opened 2024-10-17
Modify lookup file usage to enable remote file retrieval

pierreblanc:
##### SUMMARY
This commit changes the way the lookup file is handled in the DCI OpenShift Agent. Previously, the lookup file could only access local files, which posed a limitation for containerized deployments.

With this update, the lookup file can now retrieve content from a remote system, specifically allowing for the cache files to reside on the jumphost.

##### ISSUE TYPE

- New or Enhanced Feature

##### Tests

- [x] TestBos2

---

TestBos2: virt


betoredhat (approved):
LGTM

PR #473, opened 2024-10-18
Added task to wait for GitOps plugin activation

nsilla:
##### SUMMARY
In order to run ZTP spoke cluster deployments, the OpenShift GitOps operator must be complemented with the ztp site generator plugin.

The plugin is enabled through the argocd CR in the hub cluster.

Upon updating the CR, the openshift-gitops-repo-server deployment is updated with the mounted volume containing the new kustomize configuration.

This task involves pulling two images to run as initContainers to the repo server pod.

The problem is the pulling of the images may take too much time in disconnected networks, causing the job to time out while waiting for the deployment to start.

This PR adds a tasks that waits for the repo server container to be updated before the role creates the GitOps applications, so once the application is created the job may move directly to the deployment stage.

##### ISSUE TYPE

- Bug

##### Tests

- [] TestBos2 

TestBos2: bevo/acm-hub bevo/ztp-spoke-4.15
Build-Depends: 32490

### DIFF
--- a/roles/configure_ztp_gitops_apps/tasks/main.yaml
+++ b/roles/configure_ztp_gitops_apps/tasks/main.yaml
@@ -208,6 +208,22 @@
             .dockerconfigjson: "{{ _czga_encoded_pull_secret['content'] }}"
       no_log: true
 
+    - name: Wait for the GitOps ZTP plugin activation
+      community.kubernetes.k8s_info:
+        api_version: v1
+        kind: Pod
+        namespace: openshift-gitops
+        label_selectors:
+          - app.kubernetes.io/name = openshift-gitops-repo-server
+      register: _czga_repo_server_pod
+      retries: 39
+      delay: 14
+      until:
+        - _czga_repo_server_pod.resources is defined
+        - _czga_repo_server_pod.resources | length == 1
+        - _czga_repo_server_pod.resources[0] | community.general.json_query("status.conditions[?type=='Ready'].status") == ["True"]
+        - '"kustomize" in _czga_repo_server_pod.resources[0] | community.general.json_query("spec.volumes[].name")'
+
     - name: Run the policies and cluster apps
       ansible.builtin.shell: |
         {{ czga_oc_tool_path }} --kubeconfig="{{ temp_dir.path }}/kubeconfig" apply -k "{{ temp_dir.path }}/ztp/argocd/deployment"


ramperher:
I think you need to take a look at the condition you're defining in the new task because it's not working fine:

```
fatal: [localhost -> jumpbox]: FAILED! => {"msg": "The conditional check 'repo_server_pod.resources[-1] | community.general.json_query(\"status.conditions[?type=='Ready'].status\") == [\"True\"]' failed. The error was: Error in jmespath.search in json_query filter plugin:
list object has no element -1"}
```

roles/configure_ztp_gitops_apps/tasks/main.yaml
```diff
@@ -208,6 +208,22 @@
             .dockerconfigjson: "{{ _czga_encoded_pull_secret['content'] }}"
       no_log: true
 
+    - name: Wait for the GitOps ZTP plugin activation
+      community.kubernetes.k8s_info:
+        api_version: v0
```
ramperher, line 213: v1 instead of v0?

roles/configure_ztp_gitops_apps/tasks/main.yaml
```diff
@@ -208,6 +208,22 @@
...
       no_log: true
 
+    - name: Wait for the GitOps ZTP plugin activation
+      community.kubernetes.k8s_info:
+        api_version: v1
+        kind: Pod
+        namespace: openshift-gitops
+        label_selectors:
+          - app.kubernetes.io/name = openshift-gitops-repo-server
+      register: repo_server_pod
```
ramperher, line 218: this var needs to be renamed to fit in ansible lint check: `-ERROR: roles/configure_ztp_gitops_apps/tasks/main.yaml:: redhat-ci[no-role-prefix]: Variable names should use a prefix related to the role. ({'_configure_ztp_gitops_apps', '_czga'} are possible) (register: repo_server_pod)
`

ramperher (changes requested):
Only one minor issue to fix, rest looks good.

roles/configure_ztp_gitops_apps/tasks/main.yaml
ramperher: this is done

roles/configure_ztp_gitops_apps/tasks/main.yaml
```diff
@@ -208,6 +208,22 @@
...
       no_log: true
 
+    - name: Wait for the GitOps ZTP plugin activation
+      community.kubernetes.k8s_info:
+        api_version: v1
+        kind: Pod
+        namespace: openshift-gitops
+        label_selectors:
+          - app.kubernetes.io/name = openshift-gitops-repo-server
+      register: czga_repo_server_pod
```
ramperher, line 218: I think it needs to be `_czga_repo_server_pod`

ramperher (approved):
The task is passing: https://www.distributed-ci.io/jobs/ce72c227-6267-4471-8806-8a05ef8d7bfe/jobStates?sort=date&task=2361b5c2-6021-4e2c-9b2c-a906743f63c0, should be good to merge

PR #477, opened 2024-10-22
sriov_config | Add plugins field in SriovNetwork

ramperher:
##### SUMMARY

Allow the user to provide [plugins](https://docs.openshift.com/container-platform/4.17/networking/configure-syscontrols-interface-tuning-cni.html) field within the SriovNetwork resource, so that this can be propagated to the NetworkAttachmentDefinition resource.

##### ISSUE TYPE

- Enhanced Feature

##### TESTS

- [] example-cnf test

Testing will be made in a separate PR

### DIFF
--- a/roles/sriov_config/README.md
+++ b/roles/sriov_config/README.md
@@ -73,6 +73,7 @@ The `sriov_config_file` **must** meet the following requirements
 | meta_plugins      | No       | String  | MetaPluginsConfig configuration to be used in order to chain metaplugins to the sriov interface returned by the operator.
 | min_tx_rate       | No       | Int     | Minimum tx rate, in Mbps, for the VF, min_tx_rate should be <= max_tx_rate.
 | network_namespace | No       | String  | Namespace of the NetworkAttachmentDefinition custom resource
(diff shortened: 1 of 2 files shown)


roles/sriov_config/README.md
```diff
@@ -73,6 +73,7 @@ The `sriov_config_file` **must** meet the following requirements
 | meta_plugins      | No       | String  | MetaPluginsConfig configuration to be used in order to chain metaplugins to the sriov interface returned by the operator.
 | min_tx_rate       | No       | Int     | Minimum tx rate, in Mbps, for the VF, min_tx_rate should be <= max_tx_rate.
 | network_namespace | No       | String  | Namespace of the NetworkAttachmentDefinition custom resource
+| plugins           | No       | String  | Plugins to be configured for this network.
```
tonyskapunk, line 76: `plugins` is not a field supported by [SRIOV Network.](https://github.com/openshift/sriov-network-operator/blob/master/manifests/stable/sriovnetwork.openshift.io_sriovnetworks.yaml) Are you sure this is where you want to add that setting?

PR #480, opened 2024-10-28
fbc catalog - Use default channel

betoredhat:
##### SUMMARY
Use the default bundle channel

##### ISSUE TYPE

- Bug

##### Tests

- [] TestDallas - <JobID>

TestDallas: ocp-4.16-vanilla

### DIFF
--- a/roles/fbc_catalog/tasks/add-bundle.yml
+++ b/roles/fbc_catalog/tasks/add-bundle.yml
@@ -36,10 +36,10 @@
 
 - name: "Set default channel"
   vars:
-    query_channel: "Labels.\"operators.operatorframework.io.bundle.channels.v1\""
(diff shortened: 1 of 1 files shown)


betoredhat:
Currently, the fbc_catalog is only used when testing upstream catalogs, which are no longer active. This change has been validated in the pre-ga automation.

ramperher (commented):
Do you have a job where this is validated in pre-ga automation? Because in the last passed job, I cannot see this modified task being executed.

betoredhat:
> Do you have a job where this is validated in pre-ga automation? Because in the last passed job, I cannot see this modified task being executed.

Thanks, Ramon. Yes, I have one. Unfortunately, it is an internal site that not all the team can access.

tonyskapunk (approved):
LGTM

PR #497, opened 2024-11-28
Restarting the SNO image cache service only if settings change

nsilla:
##### SUMMARY

Fixes: CILAB-1916

Every time the SNO installer role is run it sets up and restarts the image cache service.

Running this task is redundant and may result in a race condition due to concurrency of jobs.

This change verifies the status of the image cache container, service unit and service and only restarts the service if a change happened.

##### ISSUE TYPE

- Bug

##### Tests

- [x] PartnerLab: https://www.distributed-ci.io/jobs/aa07d335-6333-42cc-8d4f-b5cfb400e207/jobStates

---

Test-Hint: no-check

### DIFF
--- a/roles/sno_installer/tasks/22_rhcos_image_cache.yml
+++ b/roles/sno_installer/tasks/22_rhcos_image_cache.yml
@@ -20,11 +20,23 @@
     - ['httpd_sys_content_t']
   tags: cache
 
+- name: "Get status of the RHCOS image cache container"
+  containers.podman.podman_container:
+    name: sno_image_cache
+    image: "{{ webserver_caching_image }}"
+    state: started
+    publish:
+      - "{{ webserver_caching_port_container }}:8080"
+    volumes:
+      - "{{ si_cache_dir }}:/var/www/html"
+  check_mode: true
+  register: _si_image_cache_status
+
 # Leaving SELinux details alone and not using ":z"
 # for the si_cache_dir due to issues attempting
 # to revert context. Leaving behavior as it was previously
 # to avoid breaking other user environments.
-- name: "Start RHCOS image cache container"
+- name: "Create RHCOS image cache container"
   containers.podman.podman_container:
     name: sno_image_cache
     image: "{{ webserver_caching_image }}"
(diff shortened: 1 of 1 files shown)


betoredhat:
Thanks, Nacho. Can we test this?

nsilla:
Beto, this changes seems to only be relevant to a partner environment which is not controlled by the autotest feature so I'm running the verification job manually there.

ramperher:
There are some issues with ansible lint that need to be checked:

```
roles/sno_installer/tasks/22_rhcos_image_cache.yml:32: yaml[truthy]: Truthy value should be one of
roles/sno_installer/tasks/22_rhcos_image_cache.yml:39: literal-compare: Don't compare to literal True/False.
roles/sno_installer/tasks/22_rhcos_image_cache.yml:39: no-handler: Tasks that run when changed should likely be handlers.
roles/sno_installer/tasks/22_rhcos_image_cache.yml:59: literal-compare: Don't compare to literal True/False.
roles/sno_installer/tasks/22_rhcos_image_cache.yml:116: yaml[truthy]: Truthy value should be one of
roles/sno_installer/tasks/22_rhcos_image_cache.yml:119: literal-compare: Don't compare to literal True/False.
```

roles/sno_installer/tasks/22_rhcos_image_cache.yml
```diff
@@ -91,13 +108,27 @@
...
+        state: started
+        scope: user
+      environment:
+        DBUS_SESSION_BUS_ADDRESS: "{{ ansible_env.DBUS_SESSION_BUS_ADDRESS | default('unix:path=/run/user/' + ansible_effective_user_id | string + '/bus') }}"
+      check_mode: true
+      register: _si_cache_service_status
+
     - name: "Start container-cache.service"
       ansible.builtin.systemd:
         name: container-cache.service
```
pierreblanc, line 123: stupid question, just put the state to "started" is not doing the same thing ?

pierreblanc (commented):
stupid question, just put the state to "started" is not doing the same thing ?

ramperher (approved):
LGTM

roles/sno_installer/tasks/22_rhcos_image_cache.yml
nsilla: yes and no.
If you set it as "started" and the service is already running, the service will continue running with the current settings, regardless of whether we changed the configuration of the container (for instance, the path to the mounted cache directory) or the settings in the service unit. Thus the need to restart the service.

PR #498, opened 2024-12-03
example-cnf | Remove references to load balancer

ramperher:
##### SUMMARY

example-cnf's testpmd-lb-operator is deprecated and we're removing it from the repo, so removing all code related to deploying it.

##### ISSUE TYPE

- Enhanced Feature

##### Tests

TBD

Test-Hints: no-check


ramperher:
This is ready for review, the change is working in both connected and disconnected environments, issues in previous jobs are related to certsuite because of not using HEAD version and tests_to_verify complained, but with regards to example-cnf installation, this works fine.

ramperher:
recheck

manurodriguez (approved):
Thanks, LGTM

PR #499, opened 2024-12-05
Execute dci component task in localhost

pierreblanc:
##### SUMMARY

When DCI is executed on podman, dci tasks must be executed in the container(localhost) instead of the jumphost.

##### ISSUE TYPE

- Enhanced Feature

Test-Hints: no-check


ramperher (approved):
LGTM

PR #503, opened 2024-12-10
Retry multibench workload after not-enough-CPU failure

tkrishtop:
This is a WA to retry the multibench job. 
It otherwise fails due to CPU over-consumption on server worker-3, preventing multibench pods from being ready and multibench tests to start.

Here is a [green job](https://www.distributed-ci.io/jobs/178ea82a-1f01-4c2e-930f-e1439ec05cbd/jobStates) demonstrating workaround to work.

Test-hints: no-check


manurodriguez (approved):
LGTM, about the CPU load on the first run maybe we can see if possible to reduce the number of pods, if 70 is too much.

PR #508, opened 2024-12-16
Preflight: Update the role to be able to run with dci container

pierreblanc:
##### SUMMARY

With dci in container, the source is localhost which is not the jumphost. 
- We must do these copies from the jumphost with 'remote_src'.
- Remove 'with_fileglob' which is executed in the control server. 

##### ISSUE TYPE
- Bug (with dci-podman)


ramperher:
@pierreblanc , can we specify in the PR title that we're targeting preflight role in this change?

tkrishtop (approved):
Thank you @pierreblanc LGTM

PR #509, opened 2024-12-17
Update repostory images

betoredhat:
##### SUMMARY

Update in-use role image's repositories to point to dockerhub alternatives

##### ISSUE TYPE
- nominal change

##### Tests

- [ ] TestDallasWorkload: preflight-green - <JobURL>

---

 TestDallasWorkload: preflight-green


roles/preflight/files/scorecard-old-images.yml
tonyskapunk, line 1: I don't think we are mirroring these images anymore from the DCI app-agent, maybe we can remove them?
@tkrishtop wdyt?

roles/operator_sdk/README.md
tonyskapunk, line 1: I believe we are not using operator_sdk role, maybe is time to deprecate it? @tkrishtop thoughts?

pierreblanc (approved):
LGTM

roles/operator_sdk/README.md
betoredhat: Will deprecate in a separate change if needed.

manurodriguez (approved):
Thanks, LGTM

PR #511, opened 2024-12-18
Wait for file retrival, use random file names

betoredhat:
##### SUMMARY
This used to fail the first time the http_store pod was created, as there is a short time between the pod creation and attempting to get the file.

Also in parallel executions, another deployment could modify the content of the test file, so creating random file names,

##### ISSUE TYPE

- nominal change

##### Tests

- [ ] TestBos2: abi - https://www.distributed-ci.io/jobs/cfe2ce6c-3900-4ece-8710-dc13d5aeef73/jobStates

---

### DIFF
--- a/roles/validate_http_store/tasks/main.yml
+++ b/roles/validate_http_store/tasks/main.yml
@@ -15,8 +15,11 @@
   uri:
     url: "{{ http_host }}/{{ test_file_name }}"
     return_content: true
-  register: response
+  register: _vhs_response
   delegate_to: bastion
+  retries: 10
+  delay: 3
+  until: _vhs_response.status == 200
 
 - name: Check content matches
   assert:
--- a/roles/validate_http_store/tasks/main.yml
+++ b/roles/validate_http_store/tasks/main.yml
@@ -16,16 +16,16 @@
   uri:
     url: "{{ http_host }}/{{ test_file_name }}"
     return_content: true
-  register: _vhs_response
+  register: response
   delegate_to: bastion
   retries: 12
   delay: 5
-  until: _vhs_response.status == 200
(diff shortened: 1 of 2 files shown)


betoredhat:
restest

betoredhat:
recheck

pierreblanc (approved):
LGTM

roles/validate_http_store/tasks/main.yml
```diff
@@ -1,32 +1,40 @@
+---
+- name: Create a random name for the test file
+  ansible.builtin.set_fact:
+    vhs_test_file_name: "{{ lookup('password', '/dev/null length=12 chars=ascii_letters,digits') }}"
```
tonyskapunk, line 4: I like the idea of using a random name, I believe you could keep it in the defaults and that way the var will not persist after the role.

betoredhat: I tried that, but when it was defined as the default, it generated a new random string every time the variable was referenced.

betoredhat:
recheck

roles/validate_http_store/tasks/main.yml
tkrishtop: if it's ok to use only digits for the filename, there is an option to [generate idempotent random number](https://kottapar.medium.com/idempotent-random-port-in-ansible-82f536c88f74) with the seed.
```
# Generate an idempotent random port number for the dev_user in the range 1024–32767
dev_port: “{{ 32767 | random(1024, seed=dev_user) }}”
```

ramperher (approved):
LGTM

nocturnalastro (approved):

PR #515, opened 2025-01-02
Wait twice longer for crucible ns to be destroyed

tkrishtop:
##### SUMMARY

- Trying to wait twice longer when destroying crucible namespace to [allow for the remaining pod to be deleted](https://www.distributed-ci.io/jobs/475e63a3-e915-45c8-9307-6a9e5066e9df/jobStates?sort=date&task=14be164d-2ce6-4891-8ed4-30d46e110708).
- [failed_when](https://www.distributed-ci.io/jobs/475e63a3-e915-45c8-9307-6a9e5066e9df/jobStates?sort=date&task=a2f220af-4b15-4ddf-a4e9-004d7621a7dc) is evaluated even if glfn_dir is not defined causing the task to fail. Ensured failed_when is only evaluated when glfn_dir is defined.

##### Tests

Test-Hint: no-check

### DIFF
--- a/roles/get_logs_from_namespace/tasks/main.yml
+++ b/roles/get_logs_from_namespace/tasks/main.yml
@@ -10,7 +10,9 @@
     path: "{{ glfn_dir }}"
   register: dir_stat
   when: glfn_dir is defined
-  failed_when: not dir_stat.stat.exists
+  failed_when:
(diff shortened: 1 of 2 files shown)


ramperher (approved):
LGTM

PR #524, opened 2025-01-11
Deprecate assisted on-prem related roles

betoredhat:
##### SUMMARY

Deprecate the following roles, those have been confirmed not been related to ABI and not used by DCI automation.

create_cluster
generate_discovery_iso
get_image_hash
install_cluster
monitor_cluster
monitor_host
patch_host_config
post_install
setup_assisted_installer
patch_cluster
create_day2_cluster
add_day2_node

##### ISSUE TYPE
- New or Enhanced Feature

##### Tests

Test-Hint: no-check

### DIFF
--- a/roles/create_day2_cluster/tasks/main.yml
+++ b/roles/create_day2_cluster/tasks/main.yml
@@ -1,3 +1,12 @@
+---
+- name: Deprecation warning
+  ansible.builtin.debug:
+    msg:
+      - "[DEPRECATION WARNING]: This role will be removed from the collection in the future."
+      - "Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg."
+  when:
+    - lookup('ansible.builtin.config', 'DEPRECATION_WARNINGS')
+
 - name: Create day 2 cluster
   uri:
     url: "{{ ASSISTED_INSTALLER_BASE_URL }}/clusters/import"
--- a/roles/add_day2_node/README.md
+++ b/roles/add_day2_node/README.md
@@ -1,3 +1,5 @@
 # add_day2_node
 
+⚠️ **This Ansible role is deprecated and will be removed in the future.**
+
 Adds as nodes to a pre-existing cluster using a pre-existing on-prem assisted installer instance.
--- a/roles/add_day2_node/tasks/main.yml
+++ b/roles/add_day2_node/tasks/main.yml
@@ -1,3 +1,12 @@
+---
+- name: Deprecation warning
+  ansible.builtin.debug:
+    msg:
+      - "[DEPRECATION WARNING]: This role will be removed from the collection in the future."
+      - "Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg."
+  when:
+    - lookup('ansible.builtin.config', 'DEPRECATION_WARNINGS')
+
 - name: "Install node: {{inventory_hostname }}"
   delegate_to: bastion
   block:
--- a/roles/create_cluster/README.md
+++ b/roles/create_cluster/README.md
@@ -1,3 +1,5 @@
 # create_cluster
 
+⚠️ **This Ansible role is deprecated and will be removed in the future.**
+
 Creates a cluster definition in an on-prem assisted installer instance.
--- a/roles/create_cluster/tasks/main.yml
+++ b/roles/create_cluster/tasks/main.yml
@@ -1,6 +1,14 @@
 ---
 # tasks file for create_cluster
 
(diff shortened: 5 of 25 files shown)


nocturnalastro:
Most I think are fine to be removed if we don't want on-prem assisted installer capabilities in the collection. There are just a few I wanted to query. 

`extract_openshift_installer` is used (in crucible at least) to extract the openshift installer from the release image perhaps DCI is doing this in a way? If so it is that in a standalone role in the collection? 

`post_install` should be moving the artefacts such as the kubeconfig and kubeadmin password to a standard location even in the ABI flow so perhaps it needs to be split up? If DCI is doing that a different way then perhaps its just not needed.

`mount_discovery_iso_for_pxe` I don't think this is specific to on-prem its just a method for mounting an iso. Is there other roles standalone for this in the collection that provides this functionality? 

`get_image_hash` I think there are some assumptions in the ABI roles that assumes this has been done however it might matter if that assumption is broken I don't remember the roles in enough details, if you've ran the ABI flow without this role then I'll assume its fine :)

Any of the roles removed I will probably move back into crucible to keep the functionality for a bit longer (until something breaks as it is no longer being maintained) - as far as I know there are still things that ABI doesn't do that on-prem assisted install can.

ramperher (commented):
I think the PR title should be change to "Deprecate assisted on-prem related roles", rather than "Remove..."

roles/main.yml
```diff
@@ -0,0 +1,3 @@
+deprecated:
```
nocturnalastro, line 1: I think this is needs deleting as its in the roles dir

nocturnalastro (changes requested):

roles/create_day2_cluster/tasks/main.yml
```diff
@@ -1,3 +1,11 @@
+---
+- name: Deprecation warning
```
nocturnalastro, line 2: Does Ansible do some magic to detect this as a deprecation warning? Or should this have a when clause?

betoredhat:
> Most I think are fine to be deprecated if we don't want on-prem assisted installer capabilities in the collection. There are just a few I wanted to query.
> 
> `extract_openshift_installer` is used (in crucible at least) to extract the openshift installer from the release image perhaps DCI is doing this in a way? If so it is that in a standalone role in the collection?
> 

Yes, this is not tied to AI. So will keep it in the collection

> `post_install` should be moving the artefacts such as the kubeconfig and kubeadmin password to a standard location even in the ABI flow so perhaps it needs to be split up? If DCI is doing that a different way then perhaps its just not needed.
> 

We already have some tasks in DCI for this. And this is mostly related to assisted on prem.

> `mount_discovery_iso_for_pxe` I don't think this is specific to on-prem its just a method for mounting an iso. Is there other roles standalone for this in the collection that provides this functionality?
> 

Keeping it as a standalone role. Thanks!

> `get_image_hash` I think there are some assumptions in the ABI roles that assumes this has been done however it might matter if that assumption is broken I don't remember the roles in enough details, if you've ran the ABI flow without this role then I'll assume its fine :)

I took a look to the use of this in our integration and it was only used for on-prem deployments, to get the tags, and digest of the images to pull on the provisioner node. We have validated that it is not required for ABI, as the installer binary already have the information/artifacts required for the deployment.

This is the play that we are using to deploy with ABI using DCI https://github.com/redhat-cip/dci-openshift-agent/blob/master/plays/abi-install.yml.

> 
> When the roles get removed I will probably move back into crucible to keep the functionality for a bit longer (until something breaks as it is no longer being maintained) - as far as I know there are still things that ABI doesn't do that on-prem assisted install can. :+1:

roles/create_day2_cluster/tasks/main.yml
betoredhat: I was wrongly assuming that the magic is triggered by setting the deprecated attribute in the metadata. I found that we need to check thes status of ansible.builtin.configm, so I updated the notice task accordingly

roles/extract_openshift_installer/tasks/main.yml
```diff
@@ -1,11 +1,4 @@
 ---
-- name: "Create {{ extract_dest_path }}"
```
ramperher, line 2: Is this change expected?

betoredhat: Fixed

tonyskapunk (approved):
LGTM, squash your commits and :rocket: !

betoredhat:
Thanks

PR #525, opened 2025-01-11
Add resource checks

betoredhat:
##### SUMMARY

Add conditions to check that the resource attribute exists

##### ISSUE TYPE

##### Tests

- [ ] TestBos2: virt - <JobURL>

TestBos2: virt

### DIFF
--- a/roles/nfs_external_storage/tasks/main.yml
+++ b/roles/nfs_external_storage/tasks/main.yml
@@ -215,7 +215,11 @@
   register: nfs_pod
   retries: 15
   delay: 10
-  until: nfs_pod.resources[0].status.phase == 'Running'
+  until:
+    - nfs_pod is defined
+    - nfs_pod.resources is defined
+    - nfs_pod.resources | length > 0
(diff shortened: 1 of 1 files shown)


manurodriguez (approved):
LGTM

roles/nfs_external_storage/tasks/main.yml
```diff
@@ -215,7 +215,11 @@
   register: nfs_pod
   retries: 15
   delay: 10
-  until: nfs_pod.resources[0].status.phase == 'Running'
+  until:
+    - nfs_pod is defined
```
tonyskapunk, line 219: nfs_pod should always be defined as is the register for the task, It should not be required.

roles/nfs_external_storage/tasks/main.yml
```diff
@@ -243,7 +247,9 @@
     name: test-nfs-claim
   register: test_pvc
   until:
+    - test_pvc is defined
```
tonyskapunk, line 250: ditto

tonyskapunk (approved):
LGTM

PR #530, opened 2025-01-14
multibench: Create TMPDIR directory if it's defined

pierreblanc:
##### SUMMARY

By default, tmp dir is created in the $HOME/tmp directory, 
https://github.com/redhat-cip/dci-openshift-app-agent/blob/master/plays/check_kubeconfig.yml#L32-L36

In the multibench use case it also needs to be created in the multibench host.

##### ISSUE TYPE

- Bug

### DIFF
--- a/roles/multibench_run/tasks/main.yml
+++ b/roles/multibench_run/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
-- name: "Ensure tmp directory existes"
+- name: "Ensure tmp directory exists"
   ansible.builtin.file:
     state: directory
     path: "{{ ansible_env.TMPDIR }}"
--- a/roles/multibench_run/tasks/main.yml
+++ b/roles/multibench_run/tasks/main.yml
@@ -1,4 +1,11 @@
 ---
+- name: "Ensure tmp directory existes"
+  ansible.builtin.file:
+    state: directory
+    path: "{{ ansible_env.TMPDIR }}"
+    mode: "0755"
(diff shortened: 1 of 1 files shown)


roles/multibench_run/tasks/main.yml
```diff
@@ -1,4 +1,11 @@
 ---
+- name: "Ensure tmp directory existes"
```
tonyskapunk, line 2:
```suggestion
- name: "Ensure tmp directory exists"
```

roles/multibench_run/tasks/main.yml
```diff
@@ -1,4 +1,11 @@
...
+  ansible.builtin.file:
+    state: directory
+    path: "{{ ansible_env.TMPDIR }}"
+    mode: "0755"
+  when: ansible_env.TMPDIR is defined
```
tkrishtop, line 7: will it create the folders in a recursive way? like `mkdir -p`? the problem is that there is no user dciteam in multibench vm, so as a workaround I've done `mkdir -p /home/dciteam/tmp`

tonyskapunk: This is a good observation, if this is having a problem in a non-localhost then it sounds like the ansible_env.TMPDIR should be only available in `localhost`, :thinking: I wonder if that requires a change in the app-agent instead?

pierreblanc:
I closed this one, and created a new patch in the appropriate repository.

PR #533, opened 2025-01-16
generate_agent_iso | Explicitly set PATH env var when generating the agent ISO

ramperher:
##### SUMMARY

We've seen cases like [this one](https://www.distributed-ci.io/jobs/33c420f6-7535-498b-80e5-4cde52841837/jobStates?sort=date&task=261cd5a0-62ba-4bea-a91e-0fd109164904), where the agent ISO generation is not working because `oc` binary cannot be found in `PATH`, but in fact it is. This could be because `PATH` env var seen by Ansible has a different value than running it locally.

Let's explicitly set PATH env var in this case.

Also, refactoring the whole role to follow best practices from redhatci.ocp collection.

##### ISSUE TYPE

- Bug

##### Tests

- [x] TestBos2: abi - https://www.distributed-ci.io/jobs/2970906d-7043-477f-86ab-0d678d1aee14/jobStates
- [x] deployment where we found the issue and now it's working - https://www.distributed-ci.io/jobs/2fa10b28-4881-4ab4-b4c0-0fd5a174f310/jobStates?sort=date&task=d244c1b7-c67e-4b48-b88e-6c384a0604c1

Test-Hints: no-check

build-depends: 32860

### DIFF
--- a/roles/generate_agent_iso/tasks/main.yml
+++ b/roles/generate_agent_iso/tasks/main.yml
@@ -1,14 +1,29 @@
+---
+- name: Assert the required variables are defined
+  ansible.builtin.assert:
+    that:
+      - gai_cluster_name is defined
+      - gai_cluster_name | length > 0
+      - gai_repo_root_path is defined
+      - gai_repo_root_path | length > 0
+      - gai_pull_secret is defined
+      - gai_pull_secret | length > 0
+      - gai_agent_based_installer_path is defined
+      - gai_agent_based_installer_path | length > 0
+      - gai_discovery_iso_name is defined
+      - gai_discovery_iso_name | length > 0
+
 - name: Create podman auth dir
   ansible.builtin.file:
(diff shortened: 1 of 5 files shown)


ramperher:
recheck

nocturnalastro (approved):

roles/generate_agent_iso/tasks/main.yml
```diff
@@ -1,14 +1,29 @@
+---
```
betoredhat, line 1: I think that with this we will break current role's users :(

ramperher:
recheck

ramperher:
I've bumped the version of the collections for this update, since it also requires some changes on DCI, that's why it needs a new review, cc @nocturnalastro

betoredhat:
recheck

betoredhat (approved):
lgtm

nocturnalastro (approved):

PR #537, opened 2025-01-21
Add tags to eject media tasks

betoredhat:
##### SUMMARY

Add tags to media eject tasks. 

##### ISSUE TYPE

- New or Enhanced Feature

##### Tests

- [ ] TestDallas: ocp-4.16-acm-hub - <JobURL>

Test-Hint: no-check


betoredhat:
Will bump the collections version once it is validated.

ramperher (commented):
I guess the change needs a version bump in the collections

roles/vendors/dell/tasks/iso_redfish.yml
```diff
@@ -29,7 +30,8 @@
         username: "{{ bmc_user }}"
         password: "{{ bmc_password }}"
 
-- block: # noqa unnamed-task
+- name: Shutdown
```
manurodriguez, line 33:
```suggestion
- name: Set one time boot to Virtual Media
```

roles/destroy_vms/tasks/main.yml
```diff
@@ -3,8 +3,8 @@
     cluster_vm_names: "{{ (cluster_vm_names | default([])) + [item.name] }}"
   loop: "{{ kvm_nodes }}"
 
-- debug: # noqa unnamed-task
-    var: cluster_vm_names
+- name: Debug
+  ansible.builtin.debug:
```
nocturnalastro, line 7: Var got dropped here

roles/vendors/supermicro/tasks/eject.yml
```diff
@@ -0,0 +1,43 @@
+---
```
nocturnalastro, line 1: I don't think these should be extracted into the same file. There should at least be a flag to select which tasks of which version of the API gets run.

ansible-collection-redhatci-ocp.spec
```diff
@@ -28,7 +28,7 @@ Requires: python3-jmespath
 Requires: python3-netaddr
 Requires: python3-pyyaml
 Requires: skopeo
-Conflicts: dci-openshift-agent < 1.1.0
+Conflicts: dci-openshift-agent < 1.6.0
```
tonyskapunk, line 31: This is not needed because older versions of dci-openshift-agent are not affected with this change. The requirement is satisfied in the dci-openshift-agent by using >= 1.1.0 (although there is never a 1.1.0 version).

roles/vendors/supermicro/tasks/eject.yml
betoredhat: @nocturnalastro I added a conditional to execute one or another if the API is available. Please take a look and let me know.

betoredhat:
@nocturnalastro this should be ready to review again. Thanks in advance

tonyskapunk (approved):
LGTM, please squash your commits before merging. Thanks! :smile_cat:

nocturnalastro (approved):
I like having eject in its own file :)

betoredhat:
recheck

PR #540, opened 2025-01-23
Use Hydra CNF submission provided by k8s_best_practices_certsuite

tkrishtop:
##### SUMMARY

Use new feature from certsuite included in v5.4.2 to be able to submit cert result to an existing CNF project using Hydra API.

##### ISSUE TYPE

- New feature

##### Tests

- [x] Successful submission on a dummy cert project: https://www.distributed-ci.io/jobs/77fd4bd2-5d4e-49a4-9c8f-6b152a8d6cab/jobStates. Details here: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/540#issuecomment-2621040773

Test-Hints: no-check

PR #544, opened 2025-01-29
Certsuite | Provide Hydra API parameters as flags rather than in the config file

ramperher:
##### SUMMARY

This continues https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/540

Its purpose is mainly to avoid publishing the API key, which should be a private value. If using flags in the certsuite invocation, since it's running under a `no_log: true` clause, the API key will not be printed in that case

This requires https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2739 and https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2741 to work, so consequently, this feature is only enabled when using certsuite HEAD, else we'll keep using the config file.

##### ISSUE TYPE

- Enhanced feature

##### Tests

- [x] Certsuite submission where API key is not revealed: https://www.distributed-ci.io/jobs/87330a72-b570-4c28-8a68-7e1f33dacec7/jobStates

Test-Hints: no-check

### DIFF
--- a/roles/k8s_best_practices_certsuite/tasks/tests.yml
+++ b/roles/k8s_best_practices_certsuite/tasks/tests.yml
@@ -58,6 +58,10 @@
           --preflight-dockerconfig=/usr/certsuite/config/config.json \
           --allow-preflight-insecure={{ kbpc_allow_preflight_insecure }} \
           {% endif %}
+          {% if kbpc_version == 'HEAD' and kbpc_cwe_apikey is defined and kbpc_cwe_project_id is defined %}
+          --connect-api-key={{ kbpc_cwe_apikey }} \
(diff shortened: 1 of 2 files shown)


ramperher:
> from change [dci-labs/dallas-pipelines#1345](https://github.com/dci-labs/dallas-pipelines/pull/1345):
> 
> * SUCCESS https://www.distributed-ci.io/jobs/87330a72-b570-4c28-8a68-7e1f33dacec7/jobStates

Worked here:

- Certsuite execution worked and parameters were received as flags
- Parameters were not printed in any log file, so they're safe
- Submission was made correctly

This is ready for review.

manurodriguez (approved):
LGTM

PR #554, opened 2025-02-04
Certsuite v5.4.4

rajchiluveru:
##### SUMMARY

 Bump to certsuite version to v5.4.4

##### ISSUE TYPE

Version bump

##### Tests

- [x] Certsuite submission : https://www.distributed-ci.io/jobs/da5d4e6a-a0c2-4ec0-a59e-609a7d3707f6/jobStates

Test-Hints: no-check

### DIFF
--- a/roles/k8s_best_practices_certsuite/defaults/main.yml
+++ b/roles/k8s_best_practices_certsuite/defaults/main.yml
@@ -1,6 +1,6 @@
 ---
 kbpc_check_commit_sha: false
-kbpc_version: v5.4.2
+kbpc_version: v5.4.4
 kbpc_repo_org_name: redhat-best-practices-for-k8s
 kbpc_project_name: certsuite
 kbpc_repository: "https://github.com/{{ kbpc_repo_org_name }}/{{ kbpc_project_name }}"
--- a/roles/k8s_best_practices_certsuite/tasks/tests.yml
+++ b/roles/k8s_best_practices_certsuite/tasks/tests.yml
@@ -58,7 +58,7 @@
           --preflight-dockerconfig=/usr/certsuite/config/config.json \
           --allow-preflight-insecure={{ kbpc_allow_preflight_insecure }} \
           {% endif %}
-          {% if kbpc_version == 'HEAD' and kbpc_cwe_apikey is defined and kbpc_cwe_project_id is defined %}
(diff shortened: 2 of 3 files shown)


ramperher (commented):
Hi @rajchiluveru , thanks for your change. There are a couple of modifications that are also required for this version bump:

- PR title should include the name of the role you're modifying
- PR description must refer to v5.4.4, and not v5.4.2
- [This condition](https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/k8s_best_practices_certsuite/templates/certsuite_config.yml.j2#L56) in certsuite_config.yml.j2 must be changed from: `kbpc_version is version('v5.4.2', '>=')` to `kbpc_version is version('v5.4.2', '==')`, because we only want to include the credentials and the projectId in this file if we're using v5.4.2. If using higher versions or HEAD, these variables have to be empty because we're going to provide them through flag variables to the certsuite execution.
- We also need to update the [conditions to provide flags](https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/k8s_best_practices_certsuite/tasks/tests.yml#L61) to certsuite. From: `{% if kbpc_version == 'HEAD' and kbpc_cwe_apikey is defined and kbpc_cwe_project_id is defined %}` to `{% if (kbpc_version is version('v5.4.3', '>=') or kbpc_version == 'HEAD') and kbpc_cwe_apikey is defined and kbpc_cwe_project_id is defined %}`, because v5.4.3 (and new higher versions that will come) requires Connect API feature to be enabled by using flags in our side.

ramperher:
Okey, so, the change looks good now. In your PR, you need to do several things now, to be able to merge:

- Add the last passed job to the PR description, so that it is there for future references. Typically, when doing tests, we create checklists and tick in the box when doing the corresponding test, e.g. https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/544
- Update your PR description and remove the TestDallas entry, and change it to Test-Hints: no-check. Since the change is validated, this means that no more DCI jobs will be launched from this PR, since it's stable
- You have now 3 commits in your PR, you need to squash all them to only have 1 commit

ramperher (approved):
LGTM

ramperher:
@rajchiluveru , you need to sign your commits with a GPG key to be able to merge the change.

PR #563, opened 2025-02-11
set cwe project id from CNF Project

ansvu:
##### ISSUE TYPE

- New or Enhanced Feature

##### Tests

---

### DIFF
--- a/roles/k8s_best_practices_certsuite/tasks/tests.yml
+++ b/roles/k8s_best_practices_certsuite/tasks/tests.yml
@@ -11,6 +11,13 @@
     state: directory
     mode: '0755'
 
+- name: Set KBPC CWE Project ID from CNF Certification Project
+  ansible.builtin.set_fact:
+    kbpc_cwe_project_id: "{{ cert_project_id }}"
+  when:
+    - cnf_to_certify is defined
+    - cert_project_id | length > 0
+
 - name: Tasks related to Certsuite submission through CWE Hydra API
   when: kbpc_version is version('v5.4.2', '>=') or kbpc_version == 'HEAD'
   block:


roles/k8s_best_practices_certsuite/tasks/tests.yml
```diff
@@ -11,6 +11,13 @@
     state: directory
     mode: '0755'
 
+- name: Set KBPC CWE Project ID from CNF Certification Project
```
ramperher, line 14: @ansvu, this needs to be done, probably, on `dci-openshift-app-agent` code. Here, we're mixing variables from a different role (openshift_cnf), and role variables should be decoupled.
I recommend you to do the following, instead:
- Create a change on Gerrit for dci-openshift-app-agent.
- You would need to move this [task](https://softwarefactory-project.io/r/plugins/gitiles/dci-openshift-app-agent/+/refs/heads/master/plays/tests.yml#60), `CNF Project Creation-Only tests for Vendor Validation`, [before](https://softwarefactory-project.io/r/plugins/gitiles/dci-openshift-app-agent/+/refs/heads/master/plays/tests.yml#21) the call to certsuite, else this will not work.
- Between the call to openshift_cnf and the call to certsuite role, you can include the task you're defining here. Please add a when clause to that task, to run this only if `kbpc_version is version('v5.4.2', '>=') or kbpc_version == 'HEAD'`

ramperher (changes requested):

roles/k8s_best_practices_certsuite/tasks/tests.yml
ramperher: Okey, I've seen you already did the change to run the CNF project creation before certsuite on https://softwarefactory-project.io/r/c/dci-openshift-app-agent/+/33006. So, let's use that change to include the task from this PR on that change.

ansvu: @ramperher so I will ignore your first comment? Next I will prepare dci-pipeline template to test (need @manurodriguez help to create this template on your jumphost). I need to use my CWE api token and our test-cnf. If you have better idea to test from your end, that would be great. Thanks for your comment.

ramperher: @ansvu , not at all, you need to move the task you're defining here to your change in gerrit, and close this PR. We need to automate all this in dci-openshift-app-agent

ansvu: @ramperher, Thanks for your brilliant comment! I tested locally and it seems to work fine. I will close this PR and move the changes to gerrit repo only and close this PR. Thanks again!

PR #577, opened 2025-02-18
monitor_abi | Check if certificate issue appears while monitoring installation

ramperher:
##### SUMMARY

Right now, we're hitting daily ABI issues like [this one](https://www.distributed-ci.io/jobs/6bf7bef5-ea29-4933-a1d8-69b0896d7f81/jobStates?sort=date&task=eb684fd4-b630-4a4c-a2b2-500e5c462ca7), where we're repeating the ABI monitoring check because we're not meeting the conditions to stop checking, but we should stop because we have a recurrent certificate issue, but each retry lasts 1-2h, so the job lasts >10h

Adding this certificate issue to the conditions to stop waiting. Checked manually, it works.

##### ISSUE TYPE

Bug

##### Tests

- [x] TestBos2: abi - https://www.distributed-ci.io/jobs/f5200ad3-a184-48f7-8932-2a22387d95c6/jobStates

Test-Hints: no-check

### DIFF
--- a/roles/monitor_agent_based_installer/tasks/main.yml
+++ b/roles/monitor_agent_based_installer/tasks/main.yml
@@ -23,7 +23,10 @@
       until:
         "'Attempted to gather ClusterOperator status after wait failure: Listing ClusterOperator objects' not in _mabi_install_output.stderr
         or
-        'No events logged from the Agent Rest API' in _mabi_install_output.stderr"
+        'No events logged from the Agent Rest API' in _mabi_install_output.stderr
+        or
+        'x509: certificate has expired or is not yet valid' in _mabi_install_output.stderr"
+
   rescue:
     # Using master-0 IP address to reach the bootstrap VM
     # Placing the logs in repo_root_path


roles/monitor_agent_based_installer/tasks/main.yml
```diff
@@ -23,7 +23,10 @@
       until:
         "'Attempted to gather ClusterOperator status after wait failure: Listing ClusterOperator objects' not in _mabi_install_output.stderr
         or
-        'No events logged from the Agent Rest API' in _mabi_install_output.stderr"
+        'No events logged from the Agent Rest API' in _mabi_install_output.stderr
+        or
+        'x509: certificate has expired or is not yet valid' in _mabi_install_output.stderr"
+
```
manurodriguez, line 29: Thanks @ramperher I'm running some tests, and comparing the lines you added with the log from the last failed job, the initial message `Attempted to gather ClusterOperator status after wait failure: Listing ClusterOperator objects` it's not in the log in the bootstrap run, it appears on the first retry, and it always shows on every retry, that's why it keeps falling in the loop.

`No events logged from the Agent Rest API` is not in the log at all in this job, and the message `x509: certificate has expired or is not yet valid` is on the same line as the first message in the condition, so this means it will also stay on the loop till the 10 retries.

Let me compare with a working job, to see what message we should expect to see, maybe we should add that as a condition

manurodriguez: after looking at the events from a [working job](https://www.distributed-ci.io/jobs/08f3aba7-a189-4d26-a06f-fb312889dfe0/jobStates?sort=date&task=0bd7b4f2-f8cc-40f9-be25-4e45f3885383), I was wondering if we better stay in the loop if the message `Bootstrap is complete` and `Cluster is installed` are not displayed in the logs, and only iterate 3 times, and give a 5-10 mins between the retry in case it's the API that is not responding. wdyt?

If still the job fails I believe it's acceptable, we can concentrate in finding the root cause, I reserved the cluster from the daily job that failed to dig more into the rendez-vous node logs.

manurodriguez: I got some interesting findings about today's issue, please see [CILAB-1975](https://issues.redhat.com/browse/CILAB-1975?focusedId=26622378&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26622378) and https://github.com/dci-labs/bos2-ci-config/pull/283

ramperher: @manurodriguez , got your message, thanks also for the findings, I think they're really interesting and helpful. With all this, let's recap on the status of this task:

- The retries on this task were originally included because of the issues we found in the ABI baremetal deployment on BOS2. It's in fact reported on [OCPBUGS-49157](https://issues.redhat.com/browse/OCPBUGS-49157).
- Then, we configured also the ABI pipelines launched on virt queue to run this task with retries, just to see if that helps - but it was not really making any effect.
- As you observed, it looks like the issues we had in vpool jobs were because of having a bootstrapVM in place before running the ABI job.

So, if we now manage to have ABI jobs deployed on vpool without interferences that come from bootstrapVM from IPI jobs, **maybe it's better to change these pipelines to avoid retrying on ABI installation**, so that this task would not be executed, and we would leave the retries only for the BOS2 baremetal scenario.

In that case, as it's implemented right now, we would just check the install-complete stage once; if it fails, the job fails. And in that particular task is where I'd add what you're proposing, in case we need that:  _I was wondering if we better stay in the loop if the message Bootstrap is complete and Cluster is installed are not displayed in the logs, and only iterate 3 times, and give a 5-10 mins between the retry in case it's the API that is not responding._ But maybe we don't really need it: it's highly likely that, if the installation fails once, it will fail again in the next retries.

Hence, **I'd say not modify the current task with the checks we're doing**, as long as we only use it for the BOS2 baremetal case. Checks work fine, I tried them manually, and:

- If `Attempted to gather ClusterOperator status after wait failure: Listing ClusterOperator objects` appears in the logs, we need to retry. Typically, if the check works, this will not appear, and also, if the check fails because of a different reason, this message will not appear and the task will fail, BUT...
   - If any of these messages: `No events logged from the Agent Rest API` or `x509: certificate has expired or is not yet valid`, appear in the logs (because we do the opposite check), then the job will fail.
- If none of these conditions are met, then retry.

Probably we may discuss if we really need to add these 2 extra checks, since we define them as a result of failures in the ABI jobs scheduled on vpool and not on baremetal, but they're not harmful to have them. And also, just commenting again about the 10 retries: in the original issue we found here, this check task lasted few seconds, so it may happen that 3 retries were not enough. That's why I'd keep the 10 retries for the moment.

Hope it's clearer now.

manurodriguez: Thanks for the detailed explanation Ramon, yes let's remove the WA for the virt pipelines once we implement the abi queue, and keep the retries as they are now for the baremetal pipeline. I'm also debating whether to have a dedicated pool or keep vpool as always, but clean the prov vm in ABI jobs, I'll test that.

nocturnalastro (approved):

ramperher:
thanks @nocturnalastro , as usual! :)

PR #584, opened 2025-02-25
Do not delete log dir

betoredhat:
##### SUMMARY

Do not delete the log directory passed to the role. No need to create a new log dir if the first run failed as the output is overridden.

##### ISSUE TYPE

### DIFF
--- a/roles/multibench_run/tasks/main.yml
+++ b/roles/multibench_run/tasks/main.yml
@@ -79,9 +79,11 @@
       register: _multibench_run_results
       changed_when: _multibench_run_results.rc == 0
 
-- name: Retrieve summary results from latest run if available
+- name: Retrieve summary results from Multibench VM
   vars:
-    multibench_run_filepath: "/var/lib/crucible/run/latest/run/result-summary.txt"
+    multibench_run_regex_file: >-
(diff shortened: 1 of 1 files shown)


tkrishtop (commented):
Hi @betoredhat I believe it will not work - in case of retry, we'll end up with two result files in the output folder, both conforming to timestamp-ignoring regex
```
_multibench_run_results.stdout | regex_search('iperf-and-uperf--(.+?)run/result-summary')
```

betoredhat:
> Hi @betoredhat I believe it will not work as expected - in case of retry, we'll end up with two result files in the output folder, both conforming to timestamp-ignoring regex
> 
> ```
> _multibench_run_results.stdout | regex_search('iperf-and-uperf--(.+?)run/result-summary')
> ```

Thanks for the comments @tkrishtop and the info about the parallel testing. I was wondering why we were not using the  `latest` symlink to get the latest run results.

As we are getting the path to each output file from the rc.output I think we should not worry about multiple files in the output dir as we are saving the one from the last execution.  And removing the output dir is not needed as is it not receiving anything until we grab the file from the VM.

The latest 2 jobs failed, in the second one I killed the crucible pods to force the rescue tasks.

Cheers

tkrishtop (approved):
thank you @betoredhat!

PR #585, opened 2025-02-25
Delete FBC tmp dir

betoredhat:
##### SUMMARY

Delete tmp dir once it is not needed

##### ISSUE TYPE

- Nominal change

##### Tests

- [ ] TestDallasWorkload: preflight-green - <JobURL>

TestDallasWorkload: preflight-green


betoredhat:
Already in process in another change

PR #590, opened 2025-02-27
Fix ansible lint command-instead-of-shell

tonyskapunk:
##### SUMMARY

Replace shell by command where a shell is not required. Fixes 10 command-instead-of-shell profile:basic tags:command-shell,idiom

##### ISSUE TYPE

- Nominal change

##### Tests

- [x] TestBos2: virt - https://www.distributed-ci.io/jobs/53d8e8cb-e6a0-4405-a086-ac3ceeb8e58b
- [x] TestBos2Sno: abi-sno - https://www.distributed-ci.io/jobs/4da5334d-3907-48e3-8b84-8674be3e3763

---

Test-hints: no-check

### DIFF
--- a/roles/apply_nmstate/tasks/main.yml
+++ b/roles/apply_nmstate/tasks/main.yml
@@ -19,10 +19,11 @@
         file: check_if_vm_host_is_bastion.yml
 
     - name: Apply nmstate
-      ansible.builtin.shell:
+      ansible.builtin.command:
         cmd: "nmstatectl apply --no-commit --timeout 120 {{ vm_nmstate_config_path }}"
       async: 60
       poll: 5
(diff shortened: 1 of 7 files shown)


betoredhat (approved):
lgtm

thekad (approved):
a few of these are short enough that you don't need the `cmd` piece and can just put it on the main call, but lgtm nonetheless

nocturnalastro (approved):

tonyskapunk:
> a few of these are short enough that you don't need the `cmd` piece and can just put it on the main call, but lgtm nonetheless

Agreed, I took the approach of consistency in this case instead of less typing, :see_no_evil:

PR #599, opened 2025-03-07
setup_gitea mirror mismatched repo branches

nsilla:
##### SUMMARY

Currently, the setup_gitea role only allows mirroring the repository from and to the same branch name.

This change allows mirroring between different branch names, as the role was intended from the beginning.

##### ISSUE TYPE

- Bug fix

##### Tests

- [x] TestDallasSno: ocp-4.18-hub-abi-sno - https://www.distributed-ci.io/jobs/33b6816b-f09c-49ca-b05e-2c3005581b55
- [x]  TestDallasSno: ocp-4.18-hub-abi-sno sno-abi:ansible_extravars=sg_repo_mirror_branch:ztp-flat-dirs sno-abi:ansible_extravars=sg_repo_branch:non-main - https://www.distributed-ci.io/jobs/d584cc1e-6994-4323-a4f8-d5006af7b9ce

---

Test-hints: no-check

### DIFF
--- a/roles/setup_gitea/tasks/install.yml
+++ b/roles/setup_gitea/tasks/install.yml
@@ -161,8 +161,7 @@
     - name: Mirror the repository # noqa command-instead-of-module
       ansible.builtin.shell:
         cmd: |
-          /usr/bin/git remote add mirror http://{{ sg_username }}:{{ sg_password }}@{{ sg_gitea_domain }}/{{ sg_username }}/{{ sg_repository }}.git
-          /usr/bin/git push mirror {{ sg_repo_mirror_branch }}
+          /usr/bin/git push http://{{ sg_username }}:{{ sg_password }}@{{ sg_gitea_domain }}/{{ sg_username }}/{{ sg_repository }}.git {{ sg_repo_mirror_branch }}:{{ sg_repo_branch }}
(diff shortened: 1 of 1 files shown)


nsilla:
retest

nsilla:
recheck

tonyskapunk (commented):
Sorry for the late chime in. Why do we need to be explicit about kubeconfig in this role?
We discussed in issue #296 about pros/cons of kubeconfig and my understanding is that at the time we agreed to rely on the env vars for the kubeconfig.

nsilla:
@tonyskapunk, you were right, as usual.
I created this change because when running the command from dci-pipeline the KUBECONFIG en var was missing.
I see now that it's possible to use the ansible_envvars property in the pipeline to define environment variables for the jobs.
I tested that option and it worked so I'm removing the sg_kubeconfig part of this change.
I'm keeping the change open though, since I had to add another fix for the setup_gitea role.
When mirroing the initial repository, the design was to be able to use different from and to branch names, but the implementation didn't allow for that.
The current version of this PR fixed that.

tonyskapunk:
recheck

tonyskapunk:
recheck

tonyskapunk:
image

tonyskapunk (approved):

PR #600, opened 2025-03-07
Catch when no IPMI hosts exist

thekad:
##### SUMMARY

If no IPMI connection check succeeded, the registered fact is empty, default to {} in that case.
Also a few more linting fixes.

##### ISSUE TYPE

- Bug
- 
##### Tests

Test-Hints: no-check


thekad:
@manurodriguez I see mentions to conserver in the bos2 configuration, but I am not sure where/how it is used, please advice

manurodriguez (approved):
LGTM

thekad:
retest

PR #602, opened 2025-03-08
Don't ignore the failure during bootstrap

tonyskapunk:
##### SUMMARY

Boostrap timeouts after 60 min[1], if the boostrapiong has not passed after this time, don't let the installation continue, is more likely to fail during installation but take more time to fail as the install timeouts after 90 min[2].

[1] https://github.com/openshift/installer/blob/ebe8abe386338ea666827189e5d82b55019de86c/pkg/agent/waitfor.go#L15
[2] https://github.com/openshift/installer/blob/ebe8abe386338ea666827189e5d82b55019de86c/pkg/agent/waitfor.go#L57

##### ISSUE TYPE

- Nominal change

##### Tests

- [ ] TestBos2Sno: abi-sno - 

---

TestBos2Sno: abi-sno

### DIFF
--- a/roles/monitor_agent_based_installer/tasks/main.yml
+++ b/roles/monitor_agent_based_installer/tasks/main.yml
@@ -2,7 +2,6 @@
   ansible.builtin.command:
     cmd: "{{ agent_based_installer_path }} --log-level=debug agent wait-for bootstrap-complete"
     chdir: "{{ manifests_dir }}"
-  failed_when: false
   changed_when: false
 
 - name: Check installation and gather jobs if it fails


ramperher:
I'm not against this change. Just want a double-check from @betoredhat before accepting it, because I know there are cases where bootstrap failed and then the install-complete passed, but as it's described in the PR, if the purpose of the change is to save time in case bootstrap fails because there are more chances to have an install-complete failure rather than an install-complete working execution, then I'm fine with it

betoredhat:
I agree with failing if the bootstrap is incomplete.

betoredhat (approved):

PR #606, opened 2025-03-11
setup_gitea | Mirror the correct branch of the repository

ramperher:
##### SUMMARY

Right now, we're assuming that we're always pushing the `main` branch to the mirror repository. However, there are some variables that control the branch to use. For the mirroring, it's used `sg_repo_mirror_branch`, so that variable should be used when pushing the changes to the mirrored repo.

##### ISSUE TYPE

- Bug

##### Tests

- [x] ACM hub + ZTP spoke deployment using a PR in the gitops repo - https://www.distributed-ci.io/jobs?limit=20&offset=0&sort=-created_at&where=pipeline_id:4d30330a-fe00-43da-9035-d702e76a2886,state:active

Test-Hints: no-check

### DIFF
--- a/roles/setup_gitea/tasks/install.yml
+++ b/roles/setup_gitea/tasks/install.yml
@@ -162,7 +162,7 @@
       ansible.builtin.shell:
         cmd: |
           /usr/bin/git remote add mirror http://{{ sg_username }}:{{ sg_password }}@{{ sg_gitea_domain }}/{{ sg_username }}/{{ sg_repository }}.git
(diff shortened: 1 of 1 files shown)


betoredhat:
Thanks @ramperher  we saw this behavior during teardown testing the teardwon was not deleting the applications because the job branch was different from what was tested. Great that you found the root cause. Cheeers

betoredhat (approved):
lgtm! :1st_place_medal:

PR #624, opened 2025-03-31
Delete cnf-suite namespace during tear down

rajchiluveru:
##### SUMMARY
This request is to perform cnf-suite namespace deletion during tear-down execution.
New task is added for namespace deletion.

##### ISSUE TYPE
Nominal change

##### Tests
Test-Hints: no-check

### DIFF
--- a/roles/k8s_best_practices_certsuite/tasks/teardown.yml
+++ b/roles/k8s_best_practices_certsuite/tasks/teardown.yml
@@ -25,33 +25,13 @@
 - name: Clean certsuite resources
   when: kbpc_postrun_delete_resources|bool
   block:
-    - name: Define probe daemonset namespace
-      ansible.builtin.set_fact:
-        kbpc_probe_ds_ns: certsuite
-
-    - name: Ensure probe DaemonSet is absent
-      vars:
-        kbpc_probe_ds_name: certsuite-debug
-      kubernetes.core.k8s:
-        api_version: v1
-        kind: DaemonSet
-        name: "{{ kbpc_probe_ds_name }}"
-        namespace: "{{ kbpc_probe_ds_ns }}"
-        state: absent
-
-    - name: Ensure tnf namespace is absent
+    - name: Delete cnf-suite Namespace
       kubernetes.core.k8s:
         api_version: v1
         kind: Namespace
-        name: tnf
+        name: cnf-suite
         state: absent
-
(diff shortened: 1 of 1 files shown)


roles/k8s_best_practices_certsuite/tasks/teardown.yml
```diff
@@ -21,6 +21,15 @@
     state: absent
   ignore_errors: true
 
+# Remove namespace to delete certsuite-probe pods
```
tonyskapunk, line 24: Seems to me like this task should be added in the next block named "Clean certsuite resources" as this seems to be one of those resources used by the tool.

rajchiluveru: Thanks @tonyskapunk  Agreed. Updated as suggested.

ramperher (commented):
I think these tasks can be safely removed: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/k8s_best_practices_certsuite/tasks/teardown.yml#L28-L54, they're not doing anything right now, I could see they would probably be old resources that are no longer deployed by certsuite, everything looks like running now in cnf-suite namespace.

rajchiluveru:
@ramperher It would be good to have a jira task since this is to remove tasks permanently .Hence I created this jira task https://issues.redhat.com/browse/CILAB-2102 to track the changes in certsuite role to remove those tasks that are no longer needed.

rajchiluveru:
recheck

roles/k8s_best_practices_certsuite/tasks/teardown.yml
```diff
@@ -22,36 +22,17 @@
   ignore_errors: true
 
 # Deleting resources created by the certsuite to support the execution
+# Deleted tasks that are no longer needed,refer CILAB-2102
```
tonyskapunk, line 25: Please don't include documentation that is not publicly available, from the previous line I think this line could be omitted.

rajchiluveru:
recheck

ramperher:
I think the change can be safely merged, the errors that are appearing in the debug jobs are not related to certsuite.

rajchiluveru:
Merging this change as suggested. Thanks

rajchiluveru:
nocheck

tonyskapunk (approved):
Thanks!

tonyskapunk:
retest

PR #626, opened 2025-04-01
mirror_ocp_release: fixes for concurrent jobs

nsilla:
##### SUMMARY

Fixes CILAB-2034: when multiple jobs run to mirror the same version on the same host, the status of the mirroring directory may be unstable depending on the order the tasks are run on each job.

This change implements temporary directories per role call so the artifacts are only moved to the final location at the end of the execution.

##### ISSUE TYPE

- Bug

##### Tests

- [x] TestBos2Sno: sno - https://www.distributed-ci.io/jobs/95df9c8c-513b-446b-b864-23374790bc2a/jobStates

---

TestBos2Sno: sno sno:components=ocp=4.18.5,

## Summary by CodeRabbit

- **Refactor**
  - Improved reliability and maintainability of artifact extraction and file operations by using a temporary workspace and consolidating file copy and permission steps.
  - Unified and clarified variable naming and data handling for release and image metadata across tasks.
  - Enhanced task consistency by standardizing variable names and result registration.
  - Streamlined command formatting and improved conditional logic for registry operations.

- **Bug Fixes**
  - Corrected variable references for container image sources to ensure accurate data usage.

### DIFF
--- a/roles/mirror_ocp_release/tasks/artifacts.yml
+++ b/roles/mirror_ocp_release/tasks/artifacts.yml
@@ -7,38 +7,51 @@
   when:
     - not mor_force
 
+- name: Create temporary working directory
+  ansible.builtin.tempfile:
+    state: directory
+    prefix: mor-
+  register: _mor_tmp
+  when: mor_force or not _mor_target.stat.exists
+
 - name: "Extract the OCP installer and metadata"
   when:
     - mor_force or not _mor_target.stat.exists
   block:
-    - name: "Extract installer and metadata from release image"
-      ansible.builtin.shell: >
-        flock -x {{ mor_cache_dir }}/{{ mor_version }}/release_extract.lock -c '
-        set -e;
+    - name: "Extract installer from release image"
+      ansible.builtin.command: >
         {{ mor_oc }} adm release extract
         --registry-config {{ mor_auths_file }}
         --command={{ mor_installer }}
         --from {{ mor_pull_url }}
-        --to "{{ mor_cache_dir }}/{{ mor_version }}";
-        {{ mor_oc }} adm release extract
-        --registry-config {{ mor_auths_file }}
-        --tools
-        --from {{ mor_pull_url }}
-        --to "{{ mor_cache_dir }}/{{ mor_version }}"'
+        --to "{{ _mor_tmp.path }}"
       register: _mor_extract_res
       retries: 9
       delay: 10
       until: _mor_extract_res is not failed
-      changed_when: false
+      changed_when: _mor_extract_res.rc == 0
+
+    - name: "Extract metadata from release image"
+      ansible.builtin.command: >
+        {{ mor_oc }} adm release extract
+        --registry-config {{ mor_auths_file }}
+        --tools
+        --from {{ mor_pull_url }}
+        --to "{{ _mor_tmp.path }}"
+      register: _mor_extract_res
+      retries: 9
+      delay: 10
+      until: _mor_extract_res is not failed
+      changed_when: _mor_extract_res.rc == 0
 
     - name: "Extract rhcos.json if version >= 4.8"
       when:
         - mor_version is version("4.8", ">=")
       ansible.builtin.shell: >
-        flock -x "{{ mor_cache_dir }}/{{ mor_version }}/release_extract.lock" -c '{
-        "{{ mor_cache_dir }}/{{ mor_version }}/{{ mor_installer }}" coreos print-stream-json >
-        "{{ mor_cache_dir }}/{{ mor_version }}/rhcos.json";
-        }'
+        "{{ _mor_tmp.path }}/{{ mor_installer }}" coreos print-stream-json >
+        "{{ _mor_tmp.path }}/rhcos.json"
+      args:
+        creates: "{{ _mor_tmp.path }}/rhcos.json"
 
     - name: "Download rhcos.json (< 4.8)"
       when:
@@ -47,7 +60,7 @@
         - name: "Get Git SHA from installer"
           ansible.builtin.shell: >
             set -e -o pipefail;
-            {{ mor_cache_dir }}/{{ mor_version }}/openshift-baremetal-install version |
+            {{ _mor_tmp.path }}/openshift-baremetal-install version |
             grep "^built from" |
             awk '{ print $NF }'
           register: _mor_commit_id
@@ -57,7 +70,7 @@
           ansible.builtin.include_tasks: fetch.yml
           vars:
             mor_uri: "https://raw.githubusercontent.com/openshift/installer/{{ _mor_commit_id.stdout }}/data/data/rhcos.json"
-            mor_dir: "{{ mor_cache_dir }}/{{ mor_version }}"
+            mor_dir: "{{ _mor_tmp.path }}"
 
     - name: "Figure out status of SELinux"
       ansible.builtin.command: /usr/sbin/selinuxenabled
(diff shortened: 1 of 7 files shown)


roles/mirror_ocp_release/tasks/artifacts.yml
```diff
@@ -11,20 +11,26 @@
   when:
     - mor_force or not _mor_target.stat.exists
   block:
-    - name: "Extract installer and metadata from release image"
-      ansible.builtin.shell: >
-        flock -x {{ mor_cache_dir }}/{{ mor_version }}/release_extract.lock -c '
```
nsilla, line 16: With this new approach, use of filesystem locks is not needed anymore.

roles/mirror_ocp_release/tasks/artifacts.yml
```diff
@@ -11,20 +11,26 @@
...
-        set -e;
+    - name: "Extract metadata from release image"
+      ansible.builtin.command: >
         {{ mor_oc }} adm release extract
         --registry-config {{ mor_auths_file }}
         --command={{ mor_installer }}
         --from {{ mor_pull_url }}
-        --to "{{ mor_cache_dir }}/{{ mor_version }}";
+        --to "{{ _mor_tmp.path }}";
```
nsilla, line 20: Artifacts are first extracted into the temporary directory for the job.

roles/mirror_ocp_release/tasks/fetch.yml
```diff
@@ -1,12 +1,4 @@
 ---
-- name: "Check if target file exists"
-  ansible.builtin.stat:
-    path: "{{ mor_dir }}/{{ mor_uri | basename }}"
-    get_checksum: false
-  register: target
-  when:
-    - not mor_force
```
nsilla, line 8: Since we're extracting the artifacts on a temporary directory, the file won't exist in advance.

roles/mirror_ocp_release/tasks/main.yml
```diff
@@ -78,4 +84,9 @@
   ansible.builtin.include_tasks: registry.yml
   when:
     - mor_mirror_container_images | bool
+
+- name: Copy artifacts to release directory
+  ansible.builtin.copy:
+    src: "{{ _mor_tmp.path }}/"
+    dest: "{{ mor_cache_dir }}/{{ mor_version }}"
```
nsilla, line 91: At the end of the job execution the artifacts are copied to the job's target directory. For this, we use a regular task at the end of the role instead of handlers, so we don't wait for the end of the play to copy the artifacts.

roles/mirror_ocp_release/tasks/main.yml
```diff
@@ -50,6 +50,13 @@
   ansible.builtin.debug:
     var: mor_version
 
+- name: Create temporary working directory
+  ansible.builtin.tempfile:
+    state: directory
+    prefix: mor-
+  register: _mor_tmp
+  notify: "Remove temporary directory"
```
nsilla, line 58: We use a handler to remove the temporary directory so we make sure it's run at the end of the play.

The goal is to have it running on successful and failed jobs. For this to work, the play calling this role must activate the flag "force_handlers: true", otherwise the handler will only be run on success.

We chose this approach to a block with an "always" section, so we don't need to add extra tasks to track the failed step properly.

roles/mirror_ocp_release/tasks/fetch.yml
```diff
@@ -35,4 +27,5 @@
   ansible.builtin.command: /usr/sbin/restorecon -R "{{ mor_dir }}/{{ mor_uri | basename }}"
   become: true
   when: _mor_selinux.rc == 0
+  # we may need to run this task over the target directory rather than mor_dir (= _mor_tmp.path)
```
nsilla, line 30: In the original scenario, the artifacts are extracted into an httpd served directory, so restoring the contexts is needed for the files to be properly served. Restoring the contexts on the temporary directory may not have any effect.

nsilla: Actually, the next tasks in the artifacts.yml file after including the fetch.yml set new the new context container_file_t on the extracted artifacts. This I assume is done from here so it'll be applied to both, OCP versions greater or equal than 4.8 and lower than 4.8. But that means the tasks in fetch.yml are not needed.

roles/mirror_ocp_release/tasks/fetch.yml
```diff
@@ -24,15 +16,4 @@
...
-- name: "Figure out status of SELinux"
-  ansible.builtin.command: /usr/sbin/selinuxenabled
-  ignore_errors: true
-  register: _mor_selinux
-  changed_when: false
-
-- name: "Apply new SELinux file context to file"
-  ansible.builtin.command: /usr/sbin/restorecon -R "{{ mor_dir }}/{{ mor_uri | basename }}"
-  become: true
-  when: _mor_selinux.rc == 0
```
nsilla, line 37: Restoring the selinux context does not make sense when extracting the artifacts on a temporary directory.
Also, the first tasks in artifacts.yml after including fetch.yml override the context and set it to container_file_t, which should be valid even after moving the artifacts to the target directory served from the cache container.
A different discussion is whether these tasks should be run before or after copying the artifacts to the target directory.

roles/mirror_ocp_release/tasks/artifacts.yml
```diff
@@ -90,9 +104,14 @@
       delay: 5
       until: _mor_install_mode is not failed
 
+    - name: Copy artifacts to release directory
+      ansible.builtin.copy:
+        src: "{{ _mor_tmp.path }}/"
+        dest: "{{ mor_cache_dir }}/{{ mor_version }}/"
```
betoredhat, line 110: Please apply the SELinux context to the file in the release directory.

betoredhat:
Hey @nsilla, I have some concerns that this change might undermine the caching mechanism, leading to unnecessary downloads of releases/ISOs on every job. I understand the complexity here, and I also recognize that some installers already perform similar tasks during installation.

Just my 2 cents- let's hear what the rest of the team thinks.

roles/mirror_ocp_release/tasks/artifacts.yml
nsilla: Done.

roles/mirror_ocp_release/tasks/artifacts.yml
tonyskapunk: Have you considered scenarios where two jobs run concurrently, both extracting to a temporary location and writing to `mor_cache_dir`? How do you prevent conflicts in such cases? Implementing a mechanism like a lock might help avoid these issues.

nsilla: that's a fair point. I was accepting the risk of facing such scenarios provided moving files around the filesystem is must faster than running the operations directly on the mor_cache_dir.

But at this point it's right what we could just limit the protected zone to the task were we copy the files to the cache directory once they have been processed in the temporary directory.

nsilla: here comes another thought. The way this implementation works, the problematic task would be when we run the copy module to move the files from the temporary directory to the cache directory. Alternatively or in combination of the lock usage, we can add the parameter "force: false" to the module call, so ansible won't replace a file that already exists, even if the contents are different.
The question here is whether it'd be safe to assume the artifact won't change between jobs deploying the same OCP release.
My guess is the files won't change if the jobs are running concurrently, but would those artifacts change between jobs running with, say, days of difference?

PR #640, opened 2025-04-10
http_store - Allow pods not manged by systemd and without root

betoredhat:
##### SUMMARY

* Allow to create pods not managed by systemd, useful if you to run multiple instances at the same that and managed manually
* Do not run the pod/containers as root
* Use FQCNs  

##### ISSUE TYPE

- Nominal change

##### Tests
- [x] TestBos2: abi - https://www.distributed-ci.io/jobs/66b8540f-4249-416a-8a0b-a8f7e594187e/jobStates

Test-Hint: no-check

## Summary by CodeRabbit

- **New Features**
	- Added a configuration option that lets you choose between automatic service management and manual container control for the HTTP service.

- **Documentation**
	- Updated setup guides to include details on the new configuration option and its default behavior.

- **Refactor**
	- Improved privilege escalation handling and task organization for better deployment control and clarity.

### DIFF
--- a/roles/setup_http_store/README.md
+++ b/roles/setup_http_store/README.md
@@ -11,6 +11,7 @@ Sets up a web host which can be used to distribute ISO's for `boot_iso` role
 | http_dir                  | no       | /opt/http_store                                    |                                                                                                                                      |
 | http_data_dir             | no       | "{{ http_dir }}/data"                              |                                                                                                                                      |
 | container_image           | no       | registry.centos.org/centos/httpd-24-centos7:latest | If you change this to anything other than the same image on a different host you may need to change then environment vars in the task |
+| http_store_ephemeral      | no       | false                                              | If set to true, the HTTP pod will be managed as a systemd service. If false, the pod runs directly using Podman and requires manual management |
 
 ## Dependencies
 
--- a/roles/setup_http_store/README.md
+++ b/roles/setup_http_store/README.md
@@ -11,7 +11,7 @@ Sets up a web host which can be used to distribute ISO's for `boot_iso` role
 | http_dir                  | no       | /opt/http_store                                    |                                                                                                                                      |
 | http_data_dir             | no       | "{{ http_dir }}/data"                              |                                                                                                                                      |
 | container_image           | no       | registry.centos.org/centos/httpd-24-centos7:latest | If you change this to anything other than the same image on a different host you may need to change then environment vars in the task |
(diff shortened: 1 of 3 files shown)


ramperher (commented):
LGTM, waiting for tests in Dallas before approving.
@betoredhat , do we need to test the case where `http_store_ephemeral` is true? right now I've seen that we've only tested the default case where it's set to false. I believe that, what we have right now, is the systemd service, but don't know if we want to check this here too

roles/setup_http_store/README.md
```diff
@@ -11,6 +11,7 @@ Sets up a web host which can be used to distribute ISO's for `boot_iso` role
 | http_dir                  | no       | /opt/http_store                                    |                                                                                                                                      |
 | http_data_dir             | no       | "{{ http_dir }}/data"                              |                                                                                                                                      |
 | container_image           | no       | registry.centos.org/centos/httpd-24-centos7:latest | If you change this to anything other than the same image on a different host you may need to change then environment vars in the task |
+| http_store_ephemeral      | no       | false                                              | If set to true, the HTTP pod will be managed as a systemd service. If false, the pod runs directly using Podman and requires manual management |
```
nocturnalastro, line 14: Shouldn't the default be true for backwards compatibility?

nocturnalastro, line 14: I think description is the wrong way round.

nocturnalastro (changes requested):

roles/setup_http_store/README.md
betoredhat: > I think description is the wrong way round.

Thanks, I updated it. Please take a look.

betoredhat:
> LGTM, waiting for tests in Dallas before approving. @betoredhat , do we need to test the case where `http_store_ephemeral` is true? right now I've seen that we've only tested the default case where it's set to false. I believe that, what we have right now, is the systemd service, but don't know if we want to check this here too

nocturnalastro (approved):

betoredhat:
> > LGTM, waiting for tests in Dallas before approving. @betoredhat , do we need to test the case where `http_store_ephemeral` is true? right now I've seen that we've only tested the default case where it's set to false. I believe that, what we have right now, is the systemd service, but don't know if we want to check this here too


@ramperher, I will make sure to override the var in the agent.  And update our inventories to remove the port

tonyskapunk (approved):

tonyskapunk (approved):

ramperher (approved):
LGTM

betoredhat:
recheck

PR #655, opened 2025-04-24
updated preflight loglevel to default when do submission and when test to trace

ansvu:
##### ISSUE TYPE

- New or Enhanced Feature

##### Tests

TestDallasWorkload: preflight - https://www.distributed-ci.io/jobs/6f599640-2069-4ae7-b026-220e939889ab/jobStates

## Summary by CodeRabbit

- **Refactor**
  - Updated preflight check container tasks to conditionally enable trace-level logging based on the presence of specific identifiers, reducing unnecessary verbose logs when relevant IDs are available.

test-hints: no-check

### DIFF
--- a/roles/preflight/tasks/test_preflight_container_parallel.yml
+++ b/roles/preflight/tasks/test_preflight_container_parallel.yml
@@ -12,7 +12,9 @@
       ansible.builtin.command: >
         podman run --rm --privileged
         -e PFLT_JUNIT=true
+        {% if (item.0.pyxis_container_identifier | default('') | length) == 0 %}
         -e PFLT_LOGLEVEL=trace
+        {% endif %}
         -e PFLT_ARTIFACTS=/artifacts
         -e PFLT_LOGFILE=/artifacts/preflight.log
         {% if partner_creds | length %}
--- a/roles/preflight/tasks/test_preflight_container_parallel.yml
+++ b/roles/preflight/tasks/test_preflight_container_parallel.yml
@@ -12,7 +12,8 @@
       ansible.builtin.command: >
         podman run --rm --privileged
         -e PFLT_JUNIT=true
-        {% if (item.0.pyxis_container_identifier | default('') | length) == 0 %}
+        {% if (item.0.pyxis_container_identifier | default('') | length) == 0
+           and (item.2.pyxis_container_identifier | default('') | length) == 0 %}
         -e PFLT_LOGLEVEL=trace
         {% endif %}
         -e PFLT_ARTIFACTS=/artifacts
--- a/roles/preflight/tasks/test_preflight_check_container_one_image.yml
+++ b/roles/preflight/tasks/test_preflight_check_container_one_image.yml
@@ -73,7 +73,9 @@
             --rm
             --privileged
             -e PFLT_JUNIT=true
+            {% if (cert_project_id | default('') | length) == 0 %}
             -e PFLT_LOGLEVEL=trace
+            {% endif %}
             -e PFLT_ARTIFACTS=/artifacts
             -e PFLT_LOGFILE=/artifacts/preflight.log
             {% if partner_creds | length %}


roles/preflight/tasks/test_preflight_container_parallel.yml
```diff
@@ -12,7 +12,10 @@
       ansible.builtin.command: >
         podman run --rm --privileged
         -e PFLT_JUNIT=true
+        {% if (item.0.pyxis_container_identifier | default('') | length) == 0
```
ramperher, line 15: Why `item.0` and `item.2`? What do they mean? Why not `item.1`?

roles/preflight/tasks/test_preflight_container_parallel.yml
ramperher, line 1: Do you have a test to confirm this works fine?

ramperher:
recheck

roles/preflight/tasks/test_preflight_container_parallel.yml
ansvu: When item.0 when test with recertification e.g. pyxis_container_identifier: 1111111111, item.2 is used when creating container projects + parallel. For item.0 Log, I can provide later, and item.2 logs, it needs to create projects, I can try to test it as well. Thanks for your comments.

ansvu: Recertification with parallel with submission using loglevel default e.g. info [Recertify-parallel](https://www.distributed-ci.io/jobs/c8c1f7fe-b766-47ca-ab54-b34cd389bbb5/files)

roles/preflight/tasks/test_preflight_container_parallel.yml
manurodriguez: I see in this job the preflight log is `info` : https://www.distributed-ci.io/jobs/c8c1f7fe-b766-47ca-ab54-b34cd389bbb5/jobStates?sort=date
Do you have a test where we see the log is in `trace` level`?

roles/preflight/tasks/test_preflight_container_parallel.yml
manurodriguez: Andrew, is still not very clear to me, the test you ran is an example of `re-certification + parallel` right? does that mean containers projects are not created? and as my previous question, I see the preflight logs with debug level `info`, so my question is when they will have `trace` debug level?

ansvu: When partner is doing recertification or e2e(create-project,update,attach PL), both of these use-cases are required preflight to submit the results to the backend. Since DCI preflight(sequential or parallel mode) are hard-coded LOGLEVEL=trace, so when partner do preflight test with or without submission, it always included trace logs that stored to preflight.log, logsize increased due to trace logvel 1MB+ which caused issue to preflight when it tried to submit the artifacts files to pyxis(backend), anything over 1MB, it will throw an error and saying `2025/04/22 15:46:33 could not submit to pyxis: could not create artifact: preflight.log: status code: 413: body: <html>`

So to avoid this issue, only use LOGLEVEL=trace when preflight check the container image without submission e.g. cert_project_id is not defined or empty, otherwise (submission) will use default LOGLEVEL e.g. info only so it has less log on preflight.log.

roles/preflight/tasks/test_preflight_container_parallel.yml
ansvu: Above JOB is preflight scan and submission with parallel(Recertification), so preflight.log should show only log=info.

This [ DCI JOB preflight test](https://www.distributed-ci.io/jobs/548022ec-1335-4954-ba1b-e9193b2813ba/jobStates?sort=date&task=f7cb39f1-7b1b-4067-af9a-95e8507e1243) without submission and parallel(sequential mode).  Preflight.log show log=trace

Waiting for jumphost available to test preflight with parallel but without submission(log=trace)

manurodriguez: ack ,thanks

roles/preflight/tasks/test_preflight_container_parallel.yml
manurodriguez: Thanks for the clarification, would it be possible you include a brief explanation like that somewhere in the readme or in comments in the playbook? the rest looks good to me

roles/preflight/tasks/test_preflight_container_parallel.yml
ansvu: Here is [DCI job with preflight test](https://www.distributed-ci.io/jobs/7809a67a-d243-45ec-bab3-9bc5aadbf70e/files) running in parallel without submission so expected log=trace

roles/preflight/tasks/test_preflight_container_parallel.yml
manurodriguez: I synced with Andrew about this PR, he is in a hurry for a partner to have this fix, tests looks good, so we are going to merge and then he will start a new PR to add documentation about this change.

manurodriguez (approved):
:+1:

PR #664, opened 2025-05-05
Preflight: Create certification project if necessary

pierreblanc:
##### SUMMARY
Presently the create_certification_project role try to create a project even when the project already exists.

This patch avoid this bad behavior

##### ISSUE TYPE

##### Tests
- [x] TestDallasWorkload: preflight-green : https://www.distributed-ci.io/jobs/de08d5bb-c6b7-4979-9046-19812b70aeaa/jobStates

### DIFF
--- a/roles/create_certification_project/tasks/main.yml
+++ b/roles/create_certification_project/tasks/main.yml
@@ -19,6 +19,7 @@
 
 - name: Create cert project
   ansible.builtin.include_tasks: create_project.yml
+  when: cert_project_id is undefined
 
 - name: Update the project with company info when cert_settings is defined
   ansible.builtin.include_tasks: update_project.yml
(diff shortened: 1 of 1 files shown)


roles/create_certification_project/tasks/main.yml
```diff
@@ -19,6 +19,7 @@
 
 - name: Create cert project
   ansible.builtin.include_tasks: create_project.yml
+  when: cert_project_id is undefined
```
tkrishtop, line 22: I believe it's set to empty by default: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/preflight/tasks/test_preflight_check_container_one_image.yml#L103-L106 
In the past, I did not find how to unset ansible var to undefined, so I used an empty string instead :/

pierreblanc: I understand, I am updating that

tkrishtop (approved):
@ansvu has tested the fix, the job is [green](https://www.distributed-ci.io/jobs/bf492f8c-5508-4b80-8fac-c664b051a4c7/jobStates?sort=date&task=24b84651-fc79-4b6d-ae40-a82534ddd3db).
Thank you @pierreblanc !

PR #669, opened 2025-05-07
Search sites and policy

betoredhat:
##### SUMMARY

Policies and Sites directories could be inside subdirectories. Using search to match the ArgoCD application that will be deleted

##### ISSUE TYPE

- Bug

##### Tests
- [x] Tested: ocp-4.18-spoke-ztp-sno openshift-ztp-sno:ansible_extravars=dci_teardown_on_success:true openshift-ztp-sno:components=ocp=4.18.0-0.nightly-2025-05-07-172834 https://www.distributed-ci.io/jobs/cfa8cf1c-bcd0-4e48-a5d5-cf598c557fd1/jobStates

Test-Hint: no-check

### DIFF
--- a/roles/acm_spoke_mgmt/tasks/delete-ztp-by-ref.yaml
+++ b/roles/acm_spoke_mgmt/tasks/delete-ztp-by-ref.yaml
@@ -22,7 +22,7 @@
                 | selectattr('spec.source.targetRevision', 'equalto', asm_target_revision)
                 | map(attribute='metadata.name') | list }}"
     app_policies: "{{ _asm_all_apps.resources | selectattr('spec.source.repoURL', 'equalto', asm_source_repo)
-                | selectattr('spec.source.path', 'search', '(^|/)policies($)')
+                | selectattr('spec.source.path', 'search', '(^|/)policies$')
                 | selectattr('spec.source.targetRevision', 'equalto', asm_target_revision)
                 | map(attribute='metadata.name') | list }}"
     apps_list: "{{ app_sites + app_policies | default([]) }}"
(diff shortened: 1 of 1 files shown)


ramperher (dismissed):
LGTM, I saw our current deployment works, but don't know if you want also to test the case where we have subdirectories.

betoredhat:
> LGTM, I saw our current deployment works, but don't know if you want also to test the case where we have subdirectories.

Thanks, I updated the regex to match apps where the path ends with [sites|policies]

Testing now

roles/acm_spoke_mgmt/tasks/delete-ztp-by-ref.yaml
```diff
@@ -18,11 +18,11 @@
 - name: Delete ArgoCD Applications
   vars:
     app_sites: "{{ _asm_all_apps.resources | selectattr('spec.source.repoURL', 'equalto', asm_source_repo)
-                | selectattr('spec.source.path', 'equalto', 'sites')
+                | selectattr('spec.source.path', 'search', '(^|/)sites($)')
```
tonyskapunk, line 21: I believe sites must not have anything after, so its regex should be: `sites$`, right? using grouping makes it optional so it could match something like: `sites/something_else` or `sitesSomethingElse`
```suggestion
                | selectattr('spec.source.path', 'search', '(^|/)sites$')
```

roles/acm_spoke_mgmt/tasks/delete-ztp-by-ref.yaml
```diff
@@ -18,11 +18,11 @@
...
                 | selectattr('spec.source.targetRevision', 'equalto', asm_target_revision)
                 | map(attribute='metadata.name') | list }}"
     app_policies: "{{ _asm_all_apps.resources | selectattr('spec.source.repoURL', 'equalto', asm_source_repo)
-                | selectattr('spec.source.path', 'equalto', 'policies')
+                | selectattr('spec.source.path', 'search', '(^|/)policies($)')
```
tonyskapunk, line 25: Likewise
```suggestion
                | selectattr('spec.source.path', 'search', '(^|/)policies$')
```

tonyskapunk (approved):

betoredhat:
recheck

PR #671, opened 2025-05-08
Allow set a random tag on the target

betoredhat:
##### SUMMARY

Allow to generation of a random tag for the mirrored image on the target. This allows preserving previous images that may be re-tagged with the same ID at the source

##### ISSUE TYPE

- nominal change


ramperher (approved):
LGTM

PR #672, opened 2025-05-12
Add more space for imageservice PV

betoredhat:
##### SUMMARY

The RHCOS images used to generate the discovery ISOs require more space for the assisted-image service PV.

##### ISSUE TYPE

- nominal change

##### Tests

- [ ] TestBos2: acm-hub

TestBos2: acm-hub

### DIFF
--- a/roles/acm_setup/defaults/main.yml
+++ b/roles/acm_setup/defaults/main.yml
@@ -8,6 +8,6 @@ hub_hugepages_type: hugepages-2Mi
 hub_hugepages_size: 1024Mi
 hub_db_volume_size: 40Gi
 hub_fs_volume_size: 50Gi
-hub_img_volume_size: 80Gi
+hub_img_volume_size: 200Gi
 hub_vm_external_network: true
 ...


manurodriguez (approved):
I guess this is thin space right? in BOS2 we have only 150GB disks, and even before we had exceeded this.

manurodriguez:
> > I guess this is thin space right? in BOS2 we have only 150GB disks, and even before we had exceeded this.
> 
> This space is taken from the storage class. In the current NFS server, we have 1.5TB available.

Thanks for the clarification, I've was working on ODF and I though the storage came from ODF disks :facepalm:

betoredhat:
> > > I guess this is thin space right? in BOS2 we have only 150GB disks, and even before we had exceeded this.
> > 
> > 
> > This space is taken from the storage class. In the current NFS server, we have 1.5TB available.
> 
> Thanks for the clarification, I've was working on ODF and I though the storage came from ODF disks 🤦

@manurodriguez that is a good point. The role will use the default SC, and that could be from ODF. Let me make a change to increase the disk size in IPI, for ABI, we still do not have extra disk support for ODF. 

Thanks

betoredhat:
No required

PR #678, opened 2025-05-19
[ci] check that version is managed correctly

fredericlepied:
- check the minor is incremented when a new role is added
- check the major is incremented when a role is removed

##### ISSUE TYPE

- Enhanced Feature for CI

##### Tests

---

### DIFF
--- a/hack/check_version.sh
+++ b/hack/check_version.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+#
+# Copyright (C) 2025 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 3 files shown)


fredericlepied:
Here is an example when a role is removed but the major version isn't bumped: https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/15107151678/job/42458158392?pr=678

hack/check_version.sh
```diff
@@ -0,0 +1,95 @@
...
+
+branch=$(git rev-parse --abbrev-ref HEAD)
+
+if [ -z "$branch" ]; then
+    error "Unable to determine the current branch."
+fi
+
+# Get potentially removed roles
+REMOVED=$(git diff --diff-filter=D --name-only origin/main...$branch | grep -E '^roles/[^/]+/' | cut -d/ -f2 | uniq)
+ADDED=$(git diff --diff-filter=A --name-only origin/main...$branch | grep -E '^roles/[^/]+/' | cut -d/ -f2 | uniq)
```
tonyskapunk, line 51: The grep is removing directories that could be part of a sub-role, as is the case for roles/acm/utils or roles/vendors/*

fredericlepied: Good catch. I have improved the code to take this into account. You can see the result when removing acm.utils here: https://github.com/redhatci/ansible-collection-redhatci-ocp/actions/runs/15131158226/job/42532362878?pr=678

PR #685, opened 2025-05-21
[dnm] tmp WA to disable OCP old checks

manurodriguez:
##### SUMMARY

[DNM] Temporal Work-around to make a job pass this step

Test-Hint: no-check


tonyskapunk:
We are done with our testing, closing.

PR #692, opened 2025-05-29
Update eject blocking media using key

rajchiluveru:
##### SUMMARY

When blocking_media exists, eject using device key instead of image_url

##### ISSUE TYPE
Bug

### DIFF
--- a/roles/vendors/hpe/tasks/eject.yml
+++ b/roles/vendors/hpe/tasks/eject.yml
@@ -23,7 +23,8 @@
 
     - name: Get blocking virtual_media
       ansible.builtin.set_fact:
-        blocking_virtual_media: "{{ result.redfish_facts.virtual_media.entries
+        blocking_virtual_media: "{{ result.redfish_facts.virtual_media
+            | dict2items
             | flatten(levels=2)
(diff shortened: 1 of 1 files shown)


manurodriguez:
> This job https://www.distributed-ci.io/jobs/9cbd1c95-9b5b-467b-ba67-a6659cd7a767/jobStates is successful . But it did not cover the code changes. Hence it is required to mount a device and verify this job again.

@rajchiluveru I provide some ideas in [CILAB-2170](https://issues.redhat.com/browse/CILAB-2170)

roles/vendors/hpe/tasks/iso.yml
```diff
@@ -7,9 +7,6 @@
     username: "{{ bmc_user }}"
     password: "{{ bmc_password }}"
```
tonyskapunk, line 9: Do we really want to remove the eject in here? Wondering if we want to query (if possible) to figure out what URL to request to eject first, wdyt?

betoredhat (dismissed):
lgtm

tonyskapunk (approved):
LGTM, Thanks!

betoredhat (approved):

PR #700, opened 2025-06-09
Retry eject and increase timeout

betoredhat:
##### SUMMARY

Retry media eject and increase the timeout threshold.

Keep only what is present in the rescue as it can detect the current media and eject it. 

##### ISSUE TYPE

Nominal Change

##### Tests

- [x]  Dell: https://www.distributed-ci.io/jobs/f2b178c8-a365-4f6b-907b-e24e5bce16b3/jobStates
- [x]  HPE: https://www.distributed-ci.io/jobs/eef4af95-ccb9-4a23-9e04-a5bcdf152dee/jobStates

Test-Hint: no-check

### DIFF
--- a/roles/vendors/dell/tasks/eject.yml
+++ b/roles/vendors/dell/tasks/eject.yml
@@ -12,8 +12,8 @@
   delay: 10
   until: result is succeeded
 
-- name: Get blocking virtual_media
-  ansible.builtin.set_fact:
+- name: Attempting to eject blocking media (if any)
+  vars:
     blocking_virtual_media: "{{ result.redfish_facts.virtual_media.entries
         | flatten(levels=2)
         | selectattr('ConnectedVia', 'defined') | list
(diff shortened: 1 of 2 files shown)


manurodriguez (dismissed):
LGTM

roles/vendors/dell/tasks/eject.yml
```diff
@@ -1,58 +1,54 @@
...
-                && Image != null
-                && (
-                  contains(MediaTypes, `CD`)
-                  || contains(MediaTypes, `DVD`)
-                  || contains(MediaTypes, `VCD-DVD`)
-                )
-              )]'
-            ) | from_yaml
-          }}"
+- name: Get blocking virtual_media
```
tonyskapunk, line 15: Because its only used once, this could be defined in the vars of the task using it, that way we avoid defining variables longer than needed.

tonyskapunk (commented):
The generalization makes it easier to use, as it does not matter what is mounted, it will always eject it, regardless.

Waiting for some tests to validate.

manurodriguez (approved):
LGTM

betoredhat:
recheck

betoredhat:
recheck

nocturnalastro (approved):
LGTM

PR #714, opened 2025-06-24
new role get_hub_kubeconfig

fredericlepied:
##### ISSUE TYPE

- New role get_hub_kubeconfig

##### Tests

---

### DIFF
--- a/roles/get_hub_kubeconfig/tasks/main.yaml
+++ b/roles/get_hub_kubeconfig/tasks/main.yaml
@@ -1,0 +1,60 @@
+---
+
+- name: "Validate required variables"
+  ansible.builtin.assert:
+    that:
+      - ghk_hub_kubeconfig_path is defined
+      - ghk_hub_kubeconfig_path | length > 0
+
+- name: Get base64-encoded kubeconfig from secret
+  ansible.builtin.command: >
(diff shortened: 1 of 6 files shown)


roles/get_hub_kubeconfig/tasks/main.yaml
```diff
@@ -0,0 +1,60 @@
...
+    oc get secret {{ ghk_secret_name }} -n {{ ghk_namespace }} -o jsonpath="{.data.kubeconfig}"
+  register: _ghk_kubeconfig_b64
+
+- name: Decode kubeconfig to extract details
+  ansible.builtin.set_fact:
+    ghk_hub_raw_kubeconfig: "{{ _ghk_kubeconfig_b64.stdout | b64decode }}"
+
+- name: Extract API server URL
+  ansible.builtin.set_fact:
+    ghk_hub_api_server: "{{ ghk_hub_raw_kubeconfig | regex_search('server: (https://\\S+)', '\\1') }}"
```
betoredhat, line 20: I tested this on a spoke, and unfortunately, the Server API does not refer to the public API of the hub, so it's not  going to work:
    server: https://kubernetes.default.svc:443
 
I do not see use cases where a full running managed cluster needs to make calls to the hub cluster.

PR #722, opened 2025-07-01
PowerForceOff for Dell servers

betoredhat:
##### SUMMARY

Set PowerForceOff actions for Dell servers. Graceful restart delays on nodes may cause issues on the Redevouz node. This will speed up the nodes' preparation for boostrapping.

##### ISSUE TYPE

- Nominal Change

##### Tests

- [ ]  ocp-baremetal-hcp-hub - <JobURL>

TestBos2MnoBaremetal: ocp-baremetal-hcp-hub


betoredhat:
recheck

betoredhat:
recheck

manurodriguez:
recheck

betoredhat:
> I got the impression the nodes are not booting from virtual media, but rather from the previous install in the disk because of the TLS error, we should consider to add `dci_erase_bootloader_on_disk: true` in the [cluster1 inventory](https://github.com/dci-labs/bos2-ci-config/blob/main/files/mno_baremetal/abi/cluster1) to delete the boot loader and if nodes try to boot from disk will fail and might go to virtual media at some point.

yeah @manurodriguez, thanks. I need that. I think it will also be good to not have to wait for a graceful restart during a deploy (we do not do that for HPE); have all servers start at the same time, and some minutes apart.

PR #724, opened 2025-07-07
Set only one default storageclass

betoredhat:
##### SUMMARY

The role sets 2 SCs as the default. There should not be more than one default StorageClass. Keeping only cephrbd as the default.

##### ISSUE TYPE

- Nominal change

##### Tests
- [ ] TestBos2: abi

TestBos2: abi


manurodriguez (approved):
LGTM

PR #758, opened 2025-07-31
Fix sno_node_prep overrides disconnected_registry_mirrors_file

nsilla:
##### SUMMARY

The disconnected_registry_mirrors_file variable references a yaml file with objects needed to allow for the different installers to connect to local registries mirroring the public images to allow for OCP deployment in restricted networks.

The sno_node_prep role sets this variable as a fact with the value taken from the registry_host hostvars. This usually points to the base trust-bundle file kept somewhere in the Ansible host filesystem.

However, the dci-openshift-agent also overrides the variable to a working copy it creates where it adds some changes needed to deploy the target OCP release. This means the task from the sno_node_prep role undoes what the d-o-a did in advance, causing the deployment to eventually fail.

This change removes the role task reverting the changes in the disconnected_registry_mirrors_file variable so an SNO OCP cluster may be deployed in disconnected networks.

##### ISSUE TYPE

- Bug

##### Tests

- [ ] TestDallas: ocp-4.18-vanilla - <JobURL>

---

TestDallas: ocp-4.18-vanilla
Depends-on: 34223


betoredhat:
recheck

nsilla:
retest

roles/sno_node_prep/tasks/15_validation_disconnected_registry.yml
```diff
@@ -7,14 +7,6 @@
   tags:
     - use_registry
 
-- name: Set fact disconnected_registry_mirrors_file
```
tonyskapunk, line 10: This breaks anyone defining in the inventory the `disconnected_registry_mirrors_file` under registry_host, doesn't it?

Also breaks if someone is not explicitly defining `disconnected_registry_mirrors_file` when calling this role it will break when doing a disconnected environment, right?

PR #776, opened 2025-08-27
sno_node_prep: take cache server as an input var

nsilla:
##### SUMMARY

Fixes CILAB-2258

The SNO installer role takes the input variable si_cache_server to indicate the server hosting the cache storage where the pull secret file to extract the installation artifacts is available.

The SNO node prep role is responsible for placing the pull-secret file in the cache storage but it does not allow to set a cache server different than the playbook target host.

This changes adds the variable snp_cache_server with a similar behavior to the si_cache_server variable from the sno_installer role.

##### ISSUE TYPE

- Bug fix

##### Tests

- [ ] TestBos2Sno: sno - <JobURL>
- [ ] TestBos2SnoBaremetal: sno - <JobURL>

---

TestBos2Sno: sno
TestBos2SnoBaremetal: sno


roles/sno_installer/defaults/main.yml
```diff
@@ -9,7 +9,12 @@ url_passed: false
 tftp_dir: "/var/lib/tftpboot"
 dnsmasq_enabled: true
 
-si_cache_server: "{{ cache_provisioner | ternary(groups['provisioner'][0], groups['registry_host'][0]) }}"
+si_cache_server: >-
+  {{ cache_provisioner | default(false)
```
tonyskapunk, line 13: This is changing the `cache_provisioner` default value. Because is set in `sno_node_prep`. This variable is not documented in this role, but is expected to come from that role. Why is this needed?

tonyskapunk: > the SNO roles work this way to reproduce the UPI installation workflow.

I believe you meant IPI.

> We added the si_cache_server logic when we faced an environment where the provision host had access to the internet but didn't have enough disk space to trust it with the storing of the artifact cache.

Sure, but this change is setting a default value and my comment is about that, why is it needed to set a default value?

> The two roles (sno_node_prep and sno_installer) are too coupled

Completely agree and I'm not meaning to fix that in here, with that in mind, why is it now that setting a default value of `cache_provisioner` is needed, when that is already set in `sno_node_prep`?

nsilla: I'm not sure I'm following.

You ask why the need to add the "| default(false)" to the cache_provisioner reference when setting the si_cache_server, right?

There is no actual need to do so. As you mention, the cache_server var will be set to either true or false in the sno_node_prep role.

In the end, since both roles are run sequentially, it's safe to assume the cache_provisioner variable will exist by the time this role is called.

I'll undo the change.

tonyskapunk (approved):
LGTM

PR #777, opened 2025-08-28
Fix name to reflect the right version covered

tonyskapunk:
##### SUMMARY

Name change to reflect the right action when a version of a DRAC is higher than 13G.

##### ISSUE TYPE

- Nominal change

##### Tests

Test-Hint: no-check

### DIFF
--- a/roles/vendors/dell/tasks/iso.yml
+++ b/roles/vendors/dell/tasks/iso.yml
@@ -22,6 +22,6 @@
   ansible.builtin.include_tasks: ./iso_idrac.yml
   when: drac_version.stdout | int <= 13
 
-- name: Using iDRAC ISO method for 13G and below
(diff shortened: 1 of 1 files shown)


roles/vendors/dell/tasks/iso.yml
```diff
@@ -22,6 +22,6 @@
   ansible.builtin.include_tasks: ./iso_idrac.yml
   when: drac_version.stdout | int <= 13
 
-- name: Using iDRAC ISO method for 13G and below
+- name: Using iDRAC ISO method for 14G and above
```
tonyskapunk, line 25:
```suggestion
- name: Using Redfish ISO method for 14G and above
```

ramperher (approved):
LGTM

PR #784, opened 2025-09-17
mirror_images: allow empty org like for registry.access.redhat.com/ubi10

fredericlepied:
##### ISSUE TYPE

- Bug


manurodriguez (approved):
LGTM, I think it's good to have this option, but just FYI, ubi10 can be pulled from registry.access.redhat.com/ubi10/ubi

PR #912, opened 2026-05-06
use $TMPDIR instead of hardcoding /tmp

fredericlepied:
Build-Depends: https://github.com/distributedci/dci-openshift-agent/pull/66

TestBos2: abi
TestBos2: acm-hub
TestBos2: ocp-4.14-vz-sec
TestBos2: ocp-4.19-aspenmesh-vm
TestBos2: ocp-4.21-vanilla-abi
TestBos2: virt
TestBos2: virt-prega-4.22
TestBos2Sno: abi-sno
TestBos2Sno: acm-hub-sno
Testvcp: ocp-5.0-vanilla


fredericlepied:
retest

thekad (changes requested):
Remember that all lookups are done on the ansible controller where the playbook runs. In this case there are some roles that potentially run on a different host (e.g. the provisiour could potentially be a diffent machine than the jumphost) the fact that everything is a linux host and they all write to /tmp is by pure accident. This feels wrong somehow and doesn't add anything since $TMPDIR will be dictated by a single server's env. Is there an ansible fact that you can use instead?

PR #914, opened 2026-05-11
Move patching of assistedserviceconfig to acm.utils

tonyskapunk:
##### SUMMARY

Patching the agent, is more likely to be a common action that not necessarily needs to be done during ACM configuration. Thus moving to a more generic location.

Also this is expected to be used by spoke deployments and not hub deployments, to allow the installation of non-public builds.

Gitleaks-Sign: OC4zMC4wfDIwMjYtMDUtMTFUMTg6MjI6NTU=
Gitleaks-Hash: 968eca9ed461861d875a7269ce397c417b8f9235

##### ISSUE TYPE

- Enhanced Feature

##### Tests

- [x]   ocp-4.22-spoke-ztp - https://www.distributed-ci.io/jobs/b3d4d2b2-5827-46db-8aaa-faf469e0a5f6

---

Test-Hints: no-check

### DIFF
--- a/roles/acm_setup/README.md
+++ b/roles/acm_setup/README.md
@@ -26,7 +26,6 @@ The configuration of the ACM hub can be customized by using the following variab
 | hub_img_volume_size          | 80Gi                     | No        | This value specifies how much storage is allocated for the images of the clusters. You need to allow 1 GB of image storage for each instance of Red Hat Enterprise Linux CoreOS
 | hub_img_svc_skip_tls_verify  | false                    | No        | Skip TLS verification in the AgentServiceConfig for image service operations. Useful in environments with self-signed certificates or certificate issues.
 | hub_os_images                | <Undefined>              | No        | Locations of OS Images to be used when generating the discovery ISOs for different OpenShift versions. See [OS images](./README.md#os-images). It is mandatory for disconnected environments.
-| hub_use_configmap            | false                    | No        | Allow the AgentServiceConfig to use a ConfigMap with custom env vars to modify its behavior like to disable Image Policy.
 
 ## Requirements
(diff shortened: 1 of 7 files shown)


roles/acm_setup/README.md
```diff
@@ -26,7 +26,6 @@ The configuration of the ACM hub can be customized by using the following variab
 | hub_img_volume_size          | 80Gi                     | No        | This value specifies how much storage is allocated for the images of the clusters. You need to allow 1 GB of image storage for each instance of Red Hat Enterprise Linux CoreOS
 | hub_img_svc_skip_tls_verify  | false                    | No        | Skip TLS verification in the AgentServiceConfig for image service operations. Useful in environments with self-signed certificates or certificate issues.
 | hub_os_images                | <Undefined>              | No        | Locations of OS Images to be used when generating the discovery ISOs for different OpenShift versions. See [OS images](./README.md#os-images). It is mandatory for disconnected environments.
-| hub_use_configmap            | false                    | No        | Allow the AgentServiceConfig to use a ConfigMap with custom env vars to modify its behavior like to disable Image Policy.
```
manurodriguez, line 29: Just a question, now without the boolean, by default ztp deployments will disable Image Policy when matching the condition: `build not in ['ga', 'candidate', 'dev']` right?  That condition contemplates all our uses cases, and we do not see anymore the need to define a boolean to control it? Thanks

manurodriguez (approved):

roles/acm_setup/README.md
tonyskapunk: Correct, now the call of the play in acm.utils is purposeful, meaning that you will call it only when you want to activate the custom configuration that disables the signature validation.

PR #930, opened 2026-06-02
Add wait task to check policy compliance

rajchiluveru:
##### SUMMARY
Add  new task in role, to check for policy compliance before setting facts to vars.


rajchiluveru:
this is not needed any more.

PR #936, opened 2026-06-09
mirror_ocp_release: fixed lookup plugin, removed REGISTRY_AUTH_FILE e…

kononovn:
##### SUMMARY

Fix mirror_ocp_release oc-mirror tasks for remote hosts and auth handling

The debug task that prints the rendered ImageSetConfiguration used lookup('file', ...), which always reads from the Ansible controller. When mirroring runs on a remote host, the rendered file lives under a remote temp directory (_mor_tmp_dir), so the lookup fails or shows the wrong content.

Replace it with ansible.builtin.slurp, which reads the file on the target host, and decode the result with b64decode before passing it to debug. 

When running oc-mirror, unset REGISTRY_AUTH_FILE in the shell command via env -u REGISTRY_AUTH_FILE so the tool does not pick up a host-level auth file that may conflict with the explicit --authfile {{ mor_auths_file }} argument. Credentials should come only from the path the role configures.

Test-Hint: no-check

### DIFF
--- a/roles/mirror_ocp_release/tasks/oc-mirror.yml
+++ b/roles/mirror_ocp_release/tasks/oc-mirror.yml
@@ -65,9 +65,14 @@
         dest: "{{ _mor_tmp_dir.path }}/ocp-release.yaml"
         mode: "0644"
 
+    - name: Read ImageSetConfiguration content
+      ansible.builtin.slurp:
+        src: "{{ _mor_tmp_dir.path }}/ocp-release.yaml"
+      register: _mor_imageset_config
+
     - name: Print ImageSetConfiguration content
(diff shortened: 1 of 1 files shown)


betoredhat:
retest

tonyskapunk:
@kononovn thanks for the fix, please take a look to https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/CONTRIBUTING.md#sign-commits as this repo requires commits to be signed.

kononovn:
@tonyskapunk +1. Fixed

betoredhat (approved):
lgtm

betoredhat:
retest

PR #938, opened 2026-06-10
Add in_cluster_check role

rajchiluveru:
**SUMMARY**
Add redhatci.ocp.in_cluster_check role to run Pen Drive in-cluster health checks against an OpenShift cluster

The role does following :

- Executes the pen-drive-scanner-rhel9 container via containers.podman.podman_container, mounting the cluster kubeconfig
- Discovers report generated as JSON results (*.in_cluster_checks.json)
- Converts JSON check results into JUnit XML using the new json2junit filter, writing *-junit.xml

Successful job : https://www.distributed-ci.io/jobs/62742699-b580-4ffe-b59c-b83c909e005e/jobStates

**ISSUE TYPE**
New

Test-Hint: no-check

Gitleaks-Sign: djguMzAuMXwyMDI2LTA2LTEwVDExOjUxOjMxfDlmMjdmMjY2MjE0YzI0ZGE4ZDllNmQ4M2NkMjA4ZWU3YTBhYjRiZDc=
Gitleaks-Hash: d3cae2abd41d89310d91f0949536db4e943ecdfa36d750f18c6d5f4e513d4fee


roles/in_cluster_check/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+icc_image: "registry.redhat.io/pen-drive/pen-drive-scanner-rhel9:1.0"
+icc_flow: "in-cluster-check"
+icc_keep_must_gather_data: false
+icc_tls_verify: false
+icc_pullsecret_file: "{{ pullsecret_file }}"
+icc_kubeconfig_path: "{{ KUBECONFIG_PATH }}"
```
tonyskapunk: In the collection roles we follow the practice to have [KUBECONFIG provided as an env var](https://github.com/redhatci/ansible-collection-redhatci-ocp/issues/296) to avoid defining it in each task.

roles/in_cluster_check/meta/argument_specs.yml
```diff
@@ -0,0 +1,43 @@
+---
+argument_specs:
+  main:
+    short_description: Run in-cluster checks which generate HTML report file. This report will be uploaded to DCI Control Server.
```
tonyskapunk: Roles should not interact with DCI controller, as that forces consumers to use DCI agents.

roles/in_cluster_check/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+icc_image: "registry.redhat.io/pen-drive/pen-drive-scanner-rhel9:1.0"
+icc_flow: "in-cluster-check"
+icc_keep_must_gather_data: false
+icc_tls_verify: false
+icc_pullsecret_file: "{{ pullsecret_file }}"
```
tonyskapunk: Input vars should be the ones with the prefix, i.e. icc_, `pullsecret_file` should not be an expected var.

roles/in_cluster_check/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
+---
+- name: Validate in_cluster_check required variables
+  ansible.builtin.assert:
+    that:
+      - icc_pullsecret_file is defined
+      - icc_pullsecret_file | length > 0
+      - icc_kubeconfig_path is defined
+      - icc_kubeconfig_path | length > 0
+      - job_logs.path is defined
+      - job_logs.path | length > 0
```
tonyskapunk: If job_logs are expected those should be defined in the argument_spec as well. and should be in the form prefix_var_name

roles/in_cluster_check/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: Execute in-cluster check container tool
+  ansible.builtin.shell: |
+    podman run --rm \
+      --tls-verify={{ icc_tls_verify | bool | ternary('true', 'false') }} \
```
tonyskapunk: icc_tls_verify is defined as a default, no need to set the default value again.
```suggestion
      --tls-verify={{ icc_tls_verify }} \
```

roles/in_cluster_check/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    that:
+      - icc_pullsecret_file is defined
+      - icc_pullsecret_file | length > 0
+      - icc_kubeconfig_path is defined
+      - icc_kubeconfig_path | length > 0
+      - job_logs.path is defined
+      - job_logs.path | length > 0
+
+- name: Execute in-cluster check container tool
+  ansible.builtin.shell: |
```
tonyskapunk: Please use containers.podman collection instead of shell.

roles/in_cluster_check/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    podman run --rm \
+      --tls-verify={{ icc_tls_verify | bool | ternary('true', 'false') }} \
+      --authfile={{ icc_pullsecret_file }} \
+      --userns=keep-id --user=$(id -u):$(id -g) \
+      -v {{ icc_kubeconfig_path }}:/opt/app-root/.kube/config:Z,ro \
+      -v "{{ job_logs.path }}:/mg:Z" \
+      --tz=local \
+      {{ icc_image }} {{ icc_flow }} --keep-must-gather-data={{ icc_keep_must_gather_data | bool | ternary('true', 'false') }}
```
tonyskapunk: No need to redefine the default value:
```suggestion
      {{ icc_image }} {{ icc_flow }} --keep-must-gather-data={{ icc_keep_must_gather_data }}
```

roles/in_cluster_check/meta/argument_specs.yml
```diff
@@ -0,0 +1,43 @@
...
+        default: "in-cluster-check"
+        description: Flow option passed to the container entrypoint
+        choices:
+          - "in-cluster-check"
+          - "gather"
+          - "scan"
+          - "full-run"
+          - "kube-compare"
+      icc_keep_must_gather_data:
+        type: bool
```
tonyskapunk, line 35: This one is used as a string, not as a bool

roles/in_cluster_check/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+icc_image: "registry.redhat.io/pen-drive/pen-drive-scanner-rhel9:1.0"
+icc_flow: "in-cluster-check"
+icc_keep_must_gather_data: false
```
tonyskapunk, line 4: This one is used as a string not as a bool

roles/in_cluster_check/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
...
+icc_tls_verify: false
```
tonyskapunk: ditto

roles/in_cluster_check/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: Identify in-cluster check HTML test report files
+  ansible.builtin.find:
+    paths: "{{ job_logs.path }}"
+    patterns: "*-report.html"
+    file_type: file
+  register: _icc_html_files
+  when:
+    - _icc_tool_result.stdout is defined
+    - _icc_tool_result.rc == 0
+  ignore_errors: true
```
tonyskapunk, line 39: This task seems to be mixed with actions of the DCI agent, if the task needs to copy the files somewhere else that's fine, but shouldn't assume this will run through the DCI agent.

tonyskapunk (changes requested):
I have concerns on what the role is doing, in the end the only task that's executing is a command through a container, please re-evaluate if a role is necessary for that single task.

Please provide more details on why this change is needed in the body of the PR, there's a template for it, here if you need it: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/.github/PULL_REQUEST_TEMPLATE.md

PR #939, opened 2026-06-11
Remove kvm_hosts_redfish servers group

betoredhat:
##### SUMMARY

Remove kvm_hosts_group to avoid collision with existing groups from ocp_on_libvirt

Test-Hint: no-check

### DIFF
--- a/roles/installer/tasks/59_power_off_cluster_servers.yml
+++ b/roles/installer/tasks/59_power_off_cluster_servers.yml
@@ -15,8 +15,6 @@
         ( item not in groups['dell_hosts_redfish'] | default([]) )
           and
         ( item not in groups['hp_hosts_redfish'] | default([]) )
-          and
-        ( item not in groups['kvm_hosts_redfish'] | default([]) )
       loop: "{{ groups.masters + groups.workers | default([]) }}"
 
     - name: Check node power status via ipmi to filter off nodes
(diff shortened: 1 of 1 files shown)


manurodriguez (dismissed):
LGTM, hotfix CILAB-2139

manurodriguez (approved):
LGTM

ghost:
@manurodriguez 

Having an affair with a married woman with newborn babies ?! 
A team colleague?!  You have no shame. You literally make dumps in the plate you eat. 

Being a step-father, while taking advantage of married women ?!  
Learning Russian to flirt with Russian women ?!
Pretending to be a volunteer for red-cross, while scheming to hook up Ukrainian refugee girls ?! 
Squatting an apartment with the woman you cheated with ?! 

Your deeds are even more disgusting than your appearance, a criminal profile indeed. 

Screenshot 2026-06-22 at 01 02 45

You’re a perfect representation of the place you came from, Mexico. 

You should never have existed in this world!  You belong to the toilet that was 44 years old. 

Your very existence is proof that the so-called God you believe in does not exist. Otherwise, he would be better off producing trash, which has more value than you. Seriously, every breath you draw is a waste of resources.  

Honestly, how dare you live and pretend nothing happened? 

An animal like you should never reproduce; you would never have any children of your genetics.

PR #944, opened 2026-06-17
mirror_ocp_release: added new var mor_oc_mirror_binary_url and fixed …

kononovn:
…auth issue

##### SUMMARY

This update improves OCP 4.20+ release mirroring in mirror_ocp_release in two ways.

Configurable oc-mirror download URL

The oc-mirror tarball URL was hardcoded in oc-mirror.yml. It is now exposed as mor_oc_mirror_binary_url (defaulting to the existing OpenShift mirror URL). That lets users point at an  mirror or pin a specific client version(bug free), without changing role tasks. 

Auth workaround for oc-mirror v2

oc-mirror v2 does not always honor --authfile the way the role expects. Before the mirror run, the role copies mor_auths_file to $XDG_RUNTIME_DIR/containers/auth.json, which is where the containers stack looks for credentials. The copy is removed after mirroring. The mirror command still passes --authfile and unsets REGISTRY_AUTH_FILE so env-based auth does not override the intended file. Auth-related tasks use no_log: true to avoid leaking credentials.

Documentation

mor_oc_mirror_binary_url is documented in the role README. No change to the mirroring flow beyond configurability and more reliable registry authentication.

##### ISSUE TYPE

- New or Enhanced Feature

##### Tests

---

### DIFF
--- a/roles/mirror_ocp_release/tasks/oc-mirror.yml
+++ b/roles/mirror_ocp_release/tasks/oc-mirror.yml
@@ -76,7 +76,7 @@
 
     - name: Extract latest oc-mirror
       ansible.builtin.unarchive:
-        src: "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest/oc-mirror.tar.gz"
+        src: "{{ mor_oc_mirror_binary_url }}"
         dest: "{{ _mor_tmp_dir.path }}"
         remote_src: true
         mode: "0755"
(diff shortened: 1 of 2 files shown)


roles/mirror_ocp_release/tasks/oc-mirror.yml
```diff
@@ -89,6 +89,14 @@
       redhatci.ocp.find_available_port:
       register: _mor_ephemeral_port
 
+    - name: Workaround oc-mirror v2 authfile issue
```
betoredhat, line 92: Thanks for the patch @kononovn,  If I’m not mistaken, oc-mirror v2 uses the DOCKER_CONFIG environment variable for authentication. We could copy {{ mor_auths_file }} to {{ _mor_tmp_dir.path }}/docker/config.json and pass it to the task that executes oc-mirror below. Would that work for your use case? Instead of touching `$XDG_RUNTIME_DIR/containers/auth.json"` that could be in use by other tools.

PR #951, opened 2026-06-24
Include tolerations of master, control-plane & storage when odf_setup_ocs_on_control_plane is enabled

rajchiluveru:
##### SUMMARY

Updating templates to include toleration of master/control-plane taints
This is requires to make use of odf_setup role on samsungRAN where ODF is deployed on control-plane nodes. Master nodes also carry the control-plane taint (node-role.kubernetes.io/control-plane:NoSchedule), which the gatherer does not tolerate. Kubernetes therefore treats zero nodes as eligible.

##### ISSUE TYPE

Enhanced Feature

SUCCESS https://www.distributed-ci.io/jobs/921d9afb-9fe1-4505-b0cf-69fb61fc7462/jobStates

Test-Hint: no-check

### DIFF
--- a/roles/odf_setup/templates/local-storage-block.yml.j2
+++ b/roles/odf_setup/templates/local-storage-block.yml.j2
@@ -5,6 +5,14 @@ metadata:
   name: local-block
 spec:
   tolerations:
+{% if (odf_setup_ocs_on_control_plane | default(false) | bool) and (ocp_version is version("4.16", ">=")) %}
+  - key: node-role.kubernetes.io/master
+    operator: Exists
+    effect: NoSchedule
+  - key: node-role.kubernetes.io/control-plane
+    operator: Exists
+    effect: NoSchedule
+{% endif %}
   - key: "node.ocs.openshift.io/storage"
     value: "true"
     effect: NoSchedule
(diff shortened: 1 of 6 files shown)


roles/odf_setup/templates/local-storage-block.yml.j2
```diff
@@ -5,6 +5,12 @@ metadata:
   name: local-block
 spec:
   tolerations:
+  - key: node-role.kubernetes.io/master
```
betoredhat, line 9: Installing ODF in the control plane is not a common or recommended use case. Can we apply this tolerations just in cases were the masters have OCS labels? I think that we need to avoid the tolerations when the workers act as OCS nodes.

rajchiluveru:
retest

roles/odf_setup/templates/openshift-storage-cluster.yml.j2
```diff
@@ -24,6 +24,97 @@ spec:
 {% elif ocs_install_type == 'internal' %}
   manageNodes: false
   monDataDirHostPath: /var/lib/rook
+{% if odf_setup_ocs_on_control_plane | default(false) | bool %}
```
manurodriguez: Great work, just one last thing, I doubt anyone will use this for OCP < 4.16, but as you can see in the doc https://access.redhat.com/articles/6408481#for-odf-416-and-above-4 all these resources in roles/odf_setup/templates/openshift-storage-cluster.yml.j2 template are for OCP >= 4.16, older versions do not have all those resources, so let's just include a condition that these settings are juts for OCP release >= 4.16

rajchiluveru: updated.

rajchiluveru:
retest

manurodriguez (approved):
LGTM

PR #967, opened 2026-07-01
Allow templates rendering for OCP 5x

betoredhat:
##### SUMMARY

Allow the templates rendering on ocp-5x

- [x] TestVcp: ocp-5.0-vanilla - https://www.distributed-ci.io/jobs/5f9561b9-b946-4b80-88e9-ef10084e674a/jobStates

Test-hints: no-check

### DIFF
--- a/roles/installer/templates/etc-chrony.conf.j2
+++ b/roles/installer/templates/etc-chrony.conf.j2
@@ -30,7 +30,7 @@ spec:
         group:
           name: root
         mode: 420
-{% if ((release_version.split('.')[0]|int == 4) and (release_version.split('.')[1]|int > 5)) %}
+{% if (release_version.split('.')[0]|int >= 5) or ((release_version.split('.')[0]|int == 4) and (release_version.split('.')[1]|int > 5)) %}
         overwrite: true
 {% endif %}
         path: /etc/chrony.conf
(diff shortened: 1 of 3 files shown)


tonyskapunk (approved):
LGTM

manurodriguez (approved):
It's interesting to see IPI is still around in OCP 5, I was expecting at this point we'll be using primarily other installers, but as long as it works and it's supported, let's maintain it.

betoredhat:
> It's interesting to see IPI is still around in OCP 5, I was expecting at this point we'll be using primarily other installers, but as long as it works and it's supported, let's maintain it.

Good point, let's ask on the meeting.

betoredhat:
Fixes to be addressed in a separated change

manurodriguez:
> Fixes to be addressed in a separated change

Actually this is still needed, specially because IPI with NMstate is supported since 4.10 and the current conditions do not consider OCP 5, so we won't be able to render the install-config files if we do not include OCP 5 in the conditions, the template in https://github.com/dci-labs/service-configs/pull/165 will do the heavy lifting, but this change #967 will invoke the template.

manurodriguez:
retest

manurodriguez:
retest

manurodriguez:
retest

manurodriguez:
NMstate rendering was done this time, bond1 interfaces were set with names for OCP >= 5.0 and access to storage network is working again

PR #970, opened 2026-07-03
Replace hardcoded worker node names with variables in multibench_run

manurodriguez:
##### SUMMARY

Added multibench_target_nodes variable to support FQDN node names. Templates now use configurable node lists instead of hardcoded worker-0/1/2/3. Includes validation tasks and updated documentation with usage examples.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

Gitleaks-Sign: OC4zMC4wfDIwMjYtMDctMDNUMTg6NDc6MDF8M2E5NDAyMWFlNjcyMTNlYTc3MGRjMTQyNGYyODE1NTYzODJkNjg0OQ==
Gitleaks-Hash: 0438faff2defdd25bdbcd5f08cb8712557ad1a37a2804d643b0fe54e94a1014b

##### ISSUE TYPE

Fix

##### Tests

- [x] multibench - https://www.distributed-ci.io/jobs/40bc2e29-ba22-4375-b57f-9505334155dd/jobStates?sort=date
- [x] multibench-uperf - https://www.distributed-ci.io/jobs/9b457b8c-e435-40b1-88e2-5a4603499acb/jobStates?sort=date
---

build-depends: https://github.com/dci-labs/vcp-config/pull/1663

Test-hints: no-check

### DIFF
--- a/roles/multibench_run/defaults/main.yml
+++ b/roles/multibench_run/defaults/main.yml
@@ -6,6 +6,16 @@ multibench_script_dir: "/root/crucible-examples/multibench/openshift/example-A/d
 multibench_uperf_config_base: "/root/crucible-examples/multibench/openshift/example-A/daily/uperf_config"
 multibench_script: "run.sh"
 multibench_uperf_controller_ip: "{{ multibench_controller_ip | default('192.168.5.30') }}"
+# Target worker nodes for client and server roles
+multibench_target_nodes:
+  client:
+    - worker-0
+    - worker-1
+    - worker-2
+  server:
+    - worker-0
+    - worker-1
+    - worker-2
 # Crucible metric collection (used after run when result-summary is available)
 multibench_crucible_metric_source: "mpstat"
 multibench_crucible_metric_type: "Busy-CPU"
--- a/roles/multibench_run/README.md
+++ b/roles/multibench_run/README.md
@@ -15,11 +15,56 @@ It uses a pre-existing VM/host where all the crucible binaries are ready.
 | multibench_script         | run.sh	                                                 | string    | false    | Name of the script to be executed for launching the crucible cmd. Located in the `multibench_script_dir` |
 | multibench_script_dir     | /root/crucible-examples/multibench/openshift/example-A	 | string    | false    | Path of the multibench script on the multi-bench host                                                    |
 | multibench_tags           | automation:ansible                                      | key:value | false    | Tag to be added to the multi-bench Run in the Opensearch DB                                              |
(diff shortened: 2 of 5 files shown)


manurodriguez:
retest

rajchiluveru (commented):
I would need to validate another PR on top of this change .

rajchiluveru (approved):
lgtm

roles/multibench_run/defaults/main.yml
```diff
@@ -6,6 +6,16 @@ multibench_script_dir: "/root/crucible-examples/multibench/openshift/example-A/d
 multibench_uperf_config_base: "/root/crucible-examples/multibench/openshift/example-A/daily/uperf_config"
 multibench_script: "run.sh"
 multibench_uperf_controller_ip: "{{ multibench_controller_ip | default('192.168.5.30') }}"
+# Target worker nodes for client and server roles
+multibench_target_nodes:
```
thekad, line 10: does this list correlate at all with `multibench_sample`? if so, can it default to build from that number as well? otherwise it may lead to unexpected behavior on non-default setups

manurodriguez: Could you please clarify to me what is `multibench_sample`? 

The hardcoded node names in the templates are the same I'm adding in this defaults file, so it shouldn't affect setups using the default.

Currently I'm only permitting to pass customized node names where to run the tests, but honestly when I look at the content of the templates I'm afraid it mostly matches a very particular Lab setup, so anyone that wants to do differently has very limited options to customize the setup : / 

I'm just getting familiar to this role but I can see is very opinionated to run an uperf scenario with 3 worker nodes https://github.com/redhatci/ansible-collection-redhatci-ocp/tree/main/roles/multibench_run#presentation-of-the-workload

So definitely more improvements are required on this role to allow customization.

thekad: the `multibench_sample` variable is referenced in the same defaults file, but the core of what I am asking is: if I change any of the other parameters in this role, would this new addition break? and if so, can we tie the default of this list to those other parameters?

manurodriguez: I see, the answer is no, changing any of the available parameters in the role should not break this new addition. The `multibench_sample` is the number of times the benchmark will be run, so it's not related to the number or name of worker nodes. Thanks.

thekad (approved):

manurodriguez:
retest

PR #978, opened 2026-07-09
fix(node_info): add native tools fallback for hw collection

semihkisa:
## Summary

- The lshw-based hardware collection in the `node_info` role fails on nodes that cannot reach `cdn-ubi.redhat.com` (e.g. far-edge SNOs on restricted networks). The `microdnf install` inside the UBI 10 debug pod times out, and hardware info is silently skipped.
- Wraps the existing lshw extraction in `block/rescue`. When it fails, a fallback collects equivalent hardware data using tools already present on RHCOS (`dmidecode`, `lscpu`, `lsblk`, `lspci`, `ip`) via a Python script piped through `oc debug -- chroot /host`.
- The fallback outputs JSON that mimics lshw's top-level structure so the downstream `from_json | dot_to_underscore` pipeline works without changes. A `collection_method: native-tools-fallback` field marks fallback-collected data.


ramperher:
Hi @semihkisa, thanks for your contribution. Some notes from my side:

- Could you review the failed github actions' checks to make them work? (especially the ansible-lint and the check version-docs-readme)
- Do you have a way of testing the new script? Because, in our labs, the hw collection task typically works and it's not presenting the issue you're mentioning here.

semihkisa:
> Hi @semihkisa, thanks for your contribution. Some notes from my side:
> 
> * Could you review the failed github actions' checks to make them work? (especially the ansible-lint and the check version-docs-readme)
> * Do you have a way of testing the new script? Because, in our labs, the hw collection task typically works and it's not presenting the issue you're mentioning here.

Sure! I just converted this PR to Draft, I am testing this locally but no luck so far. Once the main functionality is fine, I will fix the other errors you mentioned.

semihkisa:
PR is ready and verified here: https://www.distributed-ci.io/jobs/05695801-009e-4aae-aded-4967f87180ac/jobStates?sort=date

tonyskapunk:
Hi @semihkisa thanks for your PR, I have some questions to understand better this change

> * The lshw-based hardware collection in the `node_info` role fails on nodes that cannot reach `cdn-ubi.redhat.com` (e.g. far-edge SNOs on restricted networks). The `microdnf install` inside the UBI 10 debug pod times out, and hardware info is silently skipped.

When you mention that a cluster is on a restricted network, does that mean the cluster is disconnected? Or that it can't reach specifically the location where the ubi image is located? In the former scenario the role allows to specify an alternative location with https://github.com/redhatci/ansible-collection-redhatci-ocp/tree/main/roles/node_info#disconnected-environment For the latter, it would be best to find a way to allow access to registry.access.redhat.com where UBI is pulled from, since more images are use from there as well.
 
> * Wraps the existing lshw extraction in `block/rescue`. When it fails, a fallback collects equivalent hardware data using tools already present on RHCOS (`dmidecode`, `lscpu`, `lsblk`, `lspci`, `ip`) via a Python script piped through `oc debug -- chroot /host`.

Do you know if the equivalent you're producing is 1:1 to lshw?
We rely on lshw output in DCI to produce the Hardware section in a job. Because of this I really would like to avoid a re-implementation of lshw (there has been contributions back to lshw https://github.com/lyonel/lshw/pull/111).

semihkisa:
> Do you know if the equivalent you're producing is 1:1 to lshw?

Hi @tonyskapunk, sorry for the delay, just back from PTO.

To answer your questions, our environment is disconnected and we don't use a local registry configured, so the ni_disconnected path was not an option for us. Without this fallback, lshw fails silently and there is no hardware info collected. The fallback output is not 1:1 with lshw. It covers some basic stuff like system, BIOS, CPU, memory, storage, PCI, and network at a high level, but lshw provides much deeper detail fore sure. 
I see this fallback as a safe option, partial data marked as such, rather than silent empty data. But I understand if you'd prefer a different approach.

tonyskapunk:
> our environment is disconnected and we don't use a local registry configured

How is the cluster obtaining the OCP images?

I've never seen a disconnected cluster without a local registry, that's probably the reason why we have the role depend on a local registry for disconnected clusters, we're more than happy to consider other use cases, but that's why we would like to understand it better.

> The fallback output is not 1:1 with lshw. It covers some basic stuff like system, BIOS, CPU, memory, storage, PCI, and network at a high level, but lshw provides much deeper detail fore sure.

The concern remains on two fronts, maintainability of an implementation of lshw, and the dependency we have of lswh output in the DCI dashboard.

semihkisa:
> > our environment is disconnected and we don't use a local registry configured
> 
> How is the cluster obtaining the OCP images?
> 
> I've never seen a disconnected cluster without a local registry, that's probably the reason why we have the role depend on a local registry for disconnected clusters, we're more than happy to consider other use cases, but that's why we would like to understand it better.
> 
> > The fallback output is not 1:1 with lshw. It covers some basic stuff like system, BIOS, CPU, memory, storage, PCI, and network at a high level, but lshw provides much deeper detail fore sure.
> 
> The concern remains on two fronts, maintainability of an implementation of lshw, and the dependency we have of lswh output in the DCI dashboard.

Well, okay our environment is a little tricky, we call it disconnected because the debug pods can't reach any external network. However, the SNO cluster itself is using a proxy to pull the images from the RH registries... 

So I think in theory if we can pass the proxy var to the debug pod for the role, it would pull the image without any problem (I just confirmed that below). So maybe the right thing to do is to improve the role to support PROXY vars? 
```
[root@skisaVM-609 ~]# oc get nodes
NAME                       STATUS   ROLES                                              AGE     VERSION
cloudransno-site10-spree   Ready    control-plane,master,worker,worker-cnf,worker-rt   4d23h   v1.33.13
[root@skisaVM-609 ~]# oc debug node/cloudransno-site10-spree
Starting pod/cloudransno-site10-spree-debug-vftnp ...
To use host binaries, run `chroot /host`. Instead, if you need to access host namespaces, run `nsenter -a -t 1`.
Pod IP: fd12:3456:789a:45::73
If you don't see a command prompt, try pressing enter.
sh-5.1# microdnf install -y lshw && lshw -json -sanitize -numeric
Downloading metadata...
^C^C^Z
[1]+  Stopped(SIGTSTP)        microdnf install -y lshw

[root@skisaVM-609 ~]# oc debug node/cloudransno-site10-spree
Starting pod/cloudransno-site10-spree-debug-7jq7g ...
To use host binaries, run `chroot /host`. Instead, if you need to access host namespaces, run `nsenter -a -t 1`.
Pod IP: fd12:3456:789a:45::73
If you don't see a command prompt, try pressing enter.
sh-5.1# export https_proxy=http://[fd12:3456:789a:45::99]:3128 http_proxy=http://[fd12:3456:789a:45::99]:3128; microdnf install -y lshw && lshw -json -sanitize -numeric
Downloading metadata...
Downloading metadata...
Downloading metadata...
Package                                                                                                                                                   Repository                                                    Size
Installing:                                                                                                                                                                                                                 
 lshw-B.02.20-4.el9.x86_64                                                                                                                                ubi-9-baseos                                              336.4 kB
Transaction Summary:
 Installing:        1 packages
 Reinstalling:      0 packages
 Upgrading:         0 packages
 Obsoleting:        0 packages
 Removing:          0 packages
 Downgrading:       0 packages
Downloading packages...
Running transaction test...
Installing: lshw;B.02.20-4.el9;x86_64;ubi-9-baseos
Complete.

```

tonyskapunk:
> the SNO cluster itself is using a proxy to pull the images from the RH registries...

Oh I see, thanks for sharing this.

> So maybe the right thing to do is to improve the role to support PROXY vars?

Yes, that would be quite useful for cases like yours, :+1: 


> sh-5.1# export https_proxy=http://[fd12:3456:789a:45::99]:3128 http_proxy=http://[fd12:3456:789a:45::99]:3128; microdnf install -y lshw && lshw -json -sanitize -numeric
> Downloading metadata...
> Downloading metadata...
> Downloading metadata...
> Package                                                                                                                                                   Repository                                                    Size
> Installing:                                                                                                                                                                                                                 

:tada: 

So yeah, if you could add a proxy option to it, that should do it, similarly to `preinstall`:
https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/a4c05b08696105a0b925542b6b9c8bda03fb6f27/roles/node_info/tasks/main.yml#L132

fredericlepied (changes requested):
Don't do that as it will break the Elastic Database schema. Implement proxy support instead.t

vladir-t-k:
image

### "Mexican Wanted" 
**Rapist, child molester, cheater, squatter.**

To all of you who harbor the mexican, be wary of your female colleagues, your wife, your daughter. The Mexican is hiding among you, staring at your family. This is the animal who "offered" his hands to the young Ukrainian refugee girls at the Red Cross, who hooked up with his colleague while attending conferences on business trips, who laid his hands on newborn babies, and who disguised his whereabouts while hiding in a sanctioned country.

You’re living with a rat in your emails, in your workgroup. Every coffee you sip in the morning is mixed with rat shit. The fact that the company is still harboring this animal says a lot about the company itself. Good luck with that, and enjoy your work.

To the mexican: if you have any guts, you’d have torn out your intestines and hanged yourself in the toilet, a destiny that has been delayed for 44 years. Soon, justice and karma will bring you to your fate. An animal like you would and should never have your own family or children.

PR #980, opened 2026-07-10
include wait-sno play for MCP updates on SNO nodes

manurodriguez:
Gitleaks-Sign: OC4zMC4wfDIwMjYtMDctMTBUMTk6NTQ6MjR8NzQ3OWNkNTU0MTU2NjBhMDYwM2JiMTU5NDgyZTlkOTU0ZDQyMjk2Yg==
Gitleaks-Hash: 648de544d7bc0a4b20a89b0e3b988065dd91dca7d28af45cc9935a40acb25254

##### SUMMARY
Roles such as `deploy_cr` invoke the  `check_resource` role to wait for MCPs to be available after applying a change, but in the case of SNO nodes, the nodes are immediately restated and the MCP checks fail. This PR introduces a play to validate if the cluster is an SNO and wait for API to be available and the node ready before checking the MCP resource.

##### ISSUE TYPE

Enhanced Feature

#### Tests

- [x] TestBos2Sno: sno - https://www.distributed-ci.io/jobs/98c36a07-5e37-457c-b7f4-87ae00d72ed2/jobStates
- [x]  hub: - https://www.distributed-ci.io/jobs/e13748ea-f94d-4498-8865-0bb121f96bf5/jobStates

Test-hints: no-check

### DIFF
--- a/roles/check_resource/tasks/wait-mcp.yml
+++ b/roles/check_resource/tasks/wait-mcp.yml
@@ -4,6 +4,11 @@
     api_version: machineconfiguration.openshift.io/v1
     kind: MachineConfigPool
   register: _cr_mcp_check
+  until:
+    - _cr_mcp_check.resources is defined
+    - _cr_mcp_check.resources | length > 0
+  retries: "{{ check_wait_retries }}"
+  delay: "{{ check_wait_delay }}"
   no_log: true
 
 - name: "Wait for MCP updates"
--- a/roles/deploy_cr/README.md
+++ b/roles/deploy_cr/README.md
@@ -92,3 +92,5 @@ This Ansible role is used to deploy a Kubernetes Custom Resource (CR) to a speci
         dc_name: "{{ kubelet_config_cr.name }}"
         dc_spec: "{{ kubelet_config_cr.spec }}"
         dc_wait_condition: "{{ kubelet_config_cr.wait_condition }}"
+
+NOTE: The role detects whether the cluster is an SNO and defines the timeout values for the check_mcp tasks and consider a possible node restart.
--- a/roles/deploy_cr/tasks/main.yml
+++ b/roles/deploy_cr/tasks/main.yml
@@ -28,9 +28,22 @@
   until: _deploy_cr_result is not failed
   when: dc_wait_for_cr | bool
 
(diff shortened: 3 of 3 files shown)


betoredhat:
@manurodriguez  just wondering if instead of making the role aware of the of the cluster-type, do you think we can use a similar approach to the used during the upgrades of SNOs where the API goes down and we need to wait more time?

https://github.com/distributedci/dci-openshift-agent/blob/fa292896aeaec0e86e0949f8e8c5a2a2d7fa5c05/plays/upgrade-process.yml#L209-L227
https://github.com/distributedci/dci-openshift-agent/blob/fa292896aeaec0e86e0949f8e8c5a2a2d7fa5c05/plays/monitor-upgrade.yml#L8-L9

Then just modify the first required tasks to honor the defined wait/retries?
https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/8f3e024416f08612de1c2e7e25f155e3f98e77ea/roles/check_resource/tasks/wait-mcp.yml#L6-L7

Thanks

manurodriguez:
> @manurodriguez just wondering if instead of making the role aware of the of the cluster-type, do you think we can use a similar approach to the used during the upgrades of SNOs where the API goes down and we need to wait more time?

yes, your idea is simpler, let me give it a try

roles/check_resource/tasks/wait-mcp.yml
```diff
@@ -4,6 +4,9 @@
     api_version: machineconfiguration.openshift.io/v1
     kind: MachineConfigPool
   register: _cr_mcp_check
+  until: "'resources' in _cr_mcp_check and _cr_mcp_check.resources != []"
```
tonyskapunk: nit:
```suggestion
  until:
    - _cr_mcp_check.resources is defined
    - _cr_mcp_check.resources | length > 0
```

tonyskapunk (dismissed):

roles/check_resource/tasks/wait-mcp.yml
manurodriguez: done, thanks

tonyskapunk (approved):

betoredhat (commented):
lgtm, thanks

PR #983, opened 2026-07-15
Allow no-mcps in a cluster

betoredhat:
##### SUMMARY

Do not fail if MCPs are not present in the cluster. HCP clusters do not use MCP but the Kind is present

```
$ oc get machineconfigpool -A
No resources found
```

Test-hint: no-check


rajchiluveru (approved):
lgtm

betoredhat:
retest

PR #985, opened 2026-07-15
Fix defaults and skip log collection if no pods found

rajchiluveru:
##### SUMMARY
get_logs_from_namespace role defines a stale defaults which were not used by the role. Fixed them.
Also added condition before trying to collect logs from pods.

##### ISSUE TYPE

Enhanced Feature

Test-Hints: no-check

### DIFF
--- a/roles/get_logs_from_namespace/defaults/main.yml
+++ b/roles/get_logs_from_namespace/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
-get_logs_ns: default
-get_logs_folder: "{{ job_logs.path }}"
+glfn_ns: default
+glfn_dir: "/var/tmp/"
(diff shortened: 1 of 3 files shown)


roles/get_logs_from_namespace/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
-get_logs_ns: default
-get_logs_folder: "{{ job_logs.path }}"
+glfn_ns: default
+glfn_dir: "{{ job_logs.path }}"
```
betoredhat: This is a variable available in DOA, it will be better to default to something different and allow the user to pass the variable when calling the role.

rajchiluveru: get_logs_ns and get_logs_folder : these 2 vars are no where being used in the role. So, my intention is to  update the correct vars used in the role and mentioned in readme.md

betoredhat (dismissed):
lgtm, thanks

betoredhat (approved):
lgtm

rajchiluveru:
recheck

PR #991, opened 2026-07-24
Enable TLS verification, supress credential logging

tonyskapunk:
##### SUMMARY

Default validate_certs to true via a new jjl_validate_certs variable, allowing callers with self-signed certs to opt out explicitly. Add no_log: true to all three uri tasks to prevent the Jenkins API token from appearing in Ansible output and log collectors.

Assited-by: Claude

Gitleaks-Sign: OC4zMC4wfDIwMjYtMDctMjRUMjE6NDE6MTV8OTFiZjQ3OTA4MDBiY2M4ZjU4NDVhNDMxZTA0ZjNhZTgxMGE3NmM2ZQ==
Gitleaks-Hash: 9121e58a51137ff258f8dd918ed004bb39ce91063d82eabb3465c0ffd7424e95

##### ISSUE TYPE

- Enhanced Feature

##### Tests

Test-Hint: no-check

### DIFF
--- a/roles/jenkins_job_launcher/defaults/main.yml
+++ b/roles/jenkins_job_launcher/defaults/main.yml
@@ -1,3 +1,4 @@
 ---
 jjl_delay_seconds: 60
+jjl_validate_certs: true
 ...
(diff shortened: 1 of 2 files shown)


tonyskapunk:
@natifridman, @shaior, do you mind taking a look to this change?  This one defaults to verify TLS certs, for `jenkins_job_launcher` role, with the option to explicitly not do so.

tonyskapunk:
@natifridman, @shaior wondering if this  role is still in use, couple of thoughts
1. We could apply the changes if still used
2. Remove the roles if it is not used at all.

fredericlepied (approved):

PR #994, opened 2026-07-27
example-cnf | Improve log gathering of TRexApp

ramperher:
##### SUMMARY

Right now, there are cases like [this one](https://www.distributed-ci.io/files/d5be823a-e918-4faa-9326-b701caf648d1) where TRexApp results are not displayed in the example-cnf summary results because the filter we currently use is not enough to reach the line where the results are showed.

Changing the filter to use `grep -oE` instead.

##### ISSUE TYPE

- Bug

##### Tests

- [x] Updated log gathering: https://www.distributed-ci.io/files/fa30373b-a40a-4d25-8ac6-9d1bfb604431

Test-Hints: no-check

### DIFF
--- a/roles/example_cnf_deploy/scripts/get-example-cnf-results.sh
+++ b/roles/example_cnf_deploy/scripts/get-example-cnf-results.sh
@@ -72,12 +72,8 @@ echo
 
 echo "> App"
 echo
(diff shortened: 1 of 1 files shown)


ramperher:
Change is ready for review

manurodriguez (approved):
LGTM

PR #1005, opened 2026-07-31
removing scorecard from preflight call it is now depricated

acornett21:
##### SUMMARY

  
  Remove scorecard integration from the preflight role.

  Upstream preflight has dropped scorecard in favor of direct bundle validation ([openshift-preflight#1501](https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/1501)).

  ## Changes

  - Remove `runtime-assets` step and scorecard image files
  - Remove scorecard-related env vars (`PFLT_NAMESPACE`, `PFLT_SERVICEACCOUNT`, `PFLT_SCORECARD_IMAGE`)
  - Remove namespace create/cleanup used by scorecard
  - Remove scorecard pod/event log collection
  - Clean up defaults and docs (`preflight_sa`, `preflight_namespace`)

##### ISSUE TYPE

##### Tests

- [x] preflight-green - https://www.distributed-ci.io/jobs/eb1d29ed-ef0c-47b7-8967-25ed2318520b/jobStates

---

Test-hints: no-check


ramperher:
@acornett21 , is this applying to the latest stable preflight release? (1.19.2), or only to the code from https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/1501? Because, in that case, I believe we should wait for an official release containing that code (maybe to be done after merging the preflight change) before making changes in the collections role, which is based in latest stable code and not unstable one.

tonyskapunk:
retest

tonyskapunk:
@acornett21 you'll have to [sign](https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/CONTRIBUTING.md#sign-commits) your commits so we can merge this change.

I've added to the main comment a `Build-depends` that should validate this change along your https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/1501 change to validate it.

acornett21:
@ramperher It's a chicken and egg problem, do you want to see passing tests for DCI before this mergers? Or see failing tests? I personally want to see DCI tests passing, and since these tests/configs are in a bunch of different repos IMO it would make sense for DCI tests to pass and be merged, but I'll defer to you all.

tonyskapunk:
It can be done both at the same time if we test them also, at the same time, that's what I've been trying to do :)

ramperher:
> @ramperher It's a chicken and egg problem, do you want to see passing tests for DCI before this mergers? Or see failing tests? I personally want to see DCI tests passing, and since these tests/configs are in a bunch of different repos IMO it would make sense for DCI tests to pass and be merged, but I'll defer to you all.

As @tonyskapunk said, we can test alltogether. Now we can confirm it's passing with the proposed modifications. Then, the question is, as the preflight change will break the current logic and checks, for me it makes sense to create a new release from that change, then we can link the updates in the collections and the pipelines to that new release. What do you think?

acornett21:
@ramperher This change will have a release when it makes it to main, after it has been reviwed/approved by maintainers. After that we will cut a release. Though I don't think there is anything else to do after that, since it seems the DCI bot updates the image.

https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/81d76080272a28cef820f210b4b27561099d4ded/roles/preflight/defaults/main.yml#L2

tonyskapunk:
retest

tonyskapunk (approved):

ramperher:
> @ramperher This change will have a release when it makes it to main, after it has been reviwed/approved by maintainers. After that we will cut a release. Though I don't think there is anything else to do after that, since it seems the DCI bot updates the image.
> 
> https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/81d76080272a28cef820f210b4b27561099d4ded/roles/preflight/defaults/main.yml#L2

Yes, once the release is created, the DCI bot will automatically bump the version of the preflight image in a separate change. No worries.

ramperher (approved):
LGTM

PR #1006, opened 2026-08-03
Fix git components never being created or attached to DCI jobs

fredericlepied:
The dci_component module with state=search always returns changed=True
on a successful API call (status 200), regardless of whether any
components were found. The condition `is not changed` was therefore
always False, causing the entire create+attach block to be skipped.

Fix by checking the actual search results list instead, and move the
attach step outside the create block so that existing components are
also attached to new jobs.

TestBos2Workload: control-plane

### DIFF
--- a/roles/include_components/tasks/track_git_repo.yml
+++ b/roles/include_components/tasks/track_git_repo.yml
@@ -17,37 +17,43 @@
   register: _ic_search_git_component
   when: last_commit_id.rc == 0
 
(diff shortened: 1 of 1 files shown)


fredericlepied:
retest

ramperher (dismissed):
LGTM

tonyskapunk:
retest

tonyskapunk (approved):

PR #1010, opened 2026-08-06
Enable Two-Node with Fencing OCP support in generate_manifests (ABI) and installer (IPI) roles

ramperher:
## Summary

Enable Two-Node with Fencing (TNF) for OpenShift 4.22+ in the generate_manifests and installer roles.

When the inventory has two masters and no workers, and the target release is 4.22 or later, install-config now includes controlPlane.fencing.credentials so Agent-Based Installer and IPI can deploy TNF. BMC Redfish URLs are chosen per vendor (Dell iDRAC, HPE iLO5, or generic Redfish).

generate_manifests gates this on the new gm_ocp_version variable (documented in the role README). Unused defaults (version_filter, release_image, ocp_version_major) were dropped. Collection version is bumped 5.0 → 5.1.

## Tests

> Note: IPI scenario with TNF will not be tested in this change, as ABI support is the specified requirement on the partner that requires this feature. Just leaving the placeholder for the code in the installer role in case this is needed in the future

- [x] IPI scenario without TNF - https://www.distributed-ci.io/jobs/c5f75e4a-f680-4305-a228-4165470b66fb/jobStates?sort=date
- [x] ABI scenario without TNF - https://www.distributed-ci.io/jobs/037b4043-074b-4435-a66f-a38aa6ea6ad9/jobStates?sort=date
- [x] ABI scenario with TNF - https://www.distributed-ci.io/jobs/7999334e-cdc8-475d-a5a0-25b8daa5fd88/jobStates

Test-Hints: no-check

### DIFF
--- a/roles/generate_manifests/defaults/main.yml
+++ b/roles/generate_manifests/defaults/main.yml
@@ -32,3 +32,4 @@ ingress_vips: "{% if extra_ingress_vip is defined %}{{ [ingress_vip] + [extra_in
 
 ocp_version_major: "{{ openshift_version }}"
 gm_image_sources: ""
+gm_ocp_version: ""
--- a/roles/generate_manifests/defaults/main.yml
+++ b/roles/generate_manifests/defaults/main.yml
@@ -6,8 +6,6 @@ mac_interface_default_mapping: "interfaces[?(name != null && mac != null)].{logi
 static_network_config: {}
 root_device_hints: {}
 arch: x86_64
-version_filter: "[?(openshift_version == '{{ openshift_version }}') && (cpu_architecture == '{{ arch }}')]"
-release_image: "{{ (assisted_installer_release_images | json_query(version_filter))[0].url }}"
 
 mirror_registry_fqdn: "{{ hostvars['registry_host']['registry_fqdn'] | default(hostvars['registry_host']['ansible_fqdn']) }}"
 mirror_registry_port: "{{ hostvars['registry_host']['registry_port'] | default(5000) }}"
@@ -30,5 +28,4 @@ fetched_dest: "{{ repo_root_path }}/fetched"
 api_vips: "{% if extra_api_vip is defined %}{{ [api_vip] + [extra_api_vip] }}{% else %}{{ [api_vip] }}{% endif %}"
 ingress_vips: "{% if extra_ingress_vip is defined %}{{ [ingress_vip] + [extra_ingress_vip] }}{% else %}{{ [ingress_vip] }}{% endif %}"
 
-ocp_version_major: "{{ openshift_version }}"
 gm_image_sources: ""
--- a/roles/generate_manifests/templates/install-config.yaml.j2
+++ b/roles/generate_manifests/templates/install-config.yaml.j2
@@ -7,6 +7,27 @@ cpuPartitioningMode: AllNodes
 controlPlane:
   name: master
   replicas: {{ groups['masters'] | length }}
+{% if groups['masters'] | length == 2 and (groups['workers'] | default([])) | length == 0 and gm_ocp_version is version('4.20', '>=') %}
+  fencing:
+    credentials:
+    {% for hostname in groups['masters'] | sort %}
+      {% set vendor = hostvars[hostname]['vendor'] | lower %}
+      {% set bmc = hostvars[hostname]['bmc_address'] %}
+      - hostname: {{ hostname }}
+      {% if hostvars[hostname]['fencing_address'] is defined %}
+        address: {{ hostvars[hostname]['fencing_address'] }}
+      {% elif vendor == 'dell' %}
+        address: idrac-redfish+https://{{ bmc }}/redfish/v1/Systems/System.Embedded.1
+      {% elif vendor in ['hpe', 'hp'] %}
+        address: ilo5-redfish+https://{{ bmc }}/redfish/v1/Systems/1
+      {% else %}
+        address: redfish+https://{{ bmc }}/redfish/v1/Systems/{{ hostvars[hostname]['fencing_system_id'] | default('1') }}
+      {% endif %}
+        username: {{ hostvars[hostname]['bmc_user'] }}
+        password: {{ hostvars[hostname]['bmc_password'] }}
+        certificateVerification: {{ hostvars[hostname]['fencing_certificate_verification'] | default('Disabled') }}
+    {% endfor %}
+{% endif %}
 compute:
   - name: worker
     replicas: {{ (groups['workers'] | default([]))| length }}
--- a/roles/generate_manifests/templates/install-config.yaml.j2
+++ b/roles/generate_manifests/templates/install-config.yaml.j2
@@ -7,6 +7,25 @@ cpuPartitioningMode: AllNodes
 controlPlane:
   name: master
   replicas: {{ groups['masters'] | length }}
+{% if groups['masters'] | length == 2 and (groups['workers'] | default([])) | length == 0 and gm_ocp_version is defined and gm_ocp_version is version('4.22', '>=') %}
+  fencing:
+    credentials:
+    {% for hostname in groups['masters'] | sort %}
+      {% set vendor = hostvars[hostname]['vendor'] | lower %}
+      {% set bmc = hostvars[hostname]['bmc_address'] %}
+      - hostname: {{ hostname }}
+      {% if vendor == 'dell' %}
+        address: idrac-redfish+https://{{ bmc }}/redfish/v1/Systems/System.Embedded.1
+      {% elif vendor in ['hpe', 'hp'] %}
+        address: ilo5-redfish+https://{{ bmc }}/redfish/v1/Systems/1
+      {% else %}
+        address: redfish+https://{{ bmc }}/redfish/v1/Systems/{{ hostvars[hostname]['redfish_system_id'] | default('1') }}
+      {% endif %}
+        username: {{ hostvars[hostname]['bmc_user'] }}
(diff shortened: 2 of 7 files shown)


tonyskapunk (commented):
The fencing should be enabled in any installer that supports it, currently IPI, UPI and ABI, we can ignore UPI because the admin should take care of it.

The installer should use fencing when two control nodes are detected and when the version is 4.22 and above, despite being available since 4.20. There is no need for a flag to enable it.

ramperher:
> The fencing should be enabled in any installer that supports it, currently IPI, UPI and ABI, we can ignore UPI because the admin should take care of it.
> 
> The installer should use fencing when two control nodes are detected and when the version is 4.22 and above, despite being available since 4.20. There is no need for a flag to enable it.

I've followed your advice and add the support for IPI and ABI. I'm checking OCP to be higher or equal than 4.20 as it's tech-preview in that release, but we can directly check for 4.22 if preferred.

roles/generate_manifests/defaults/main.yml
```diff
@@ -32,3 +32,4 @@ ingress_vips: "{% if extra_ingress_vip is defined %}{{ [ingress_vip] + [extra_in
 
 ocp_version_major: "{{ openshift_version }}"
 gm_image_sources: ""
+gm_ocp_version: ""
```
tonyskapunk: No need to add another var when we have one already:
https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/96f540fb723bb986f942f7e59a6a5c0241a750d7/roles/generate_manifests/defaults/main.yml#L33

roles/generate_manifests/templates/install-config.yaml.j2
```diff
@@ -7,6 +7,27 @@ cpuPartitioningMode: AllNodes
 controlPlane:
   name: master
   replicas: {{ groups['masters'] | length }}
+{% if groups['masters'] | length == 2 and (groups['workers'] | default([])) | length == 0 and gm_ocp_version is version('4.20', '>=') %}
```
tonyskapunk: I don't think there is a requirement on the workers being 0, the only requirement is 2 control-planes, right?
Do we plan to support tech-preview (4.20)? or only GA (4.22)?

roles/generate_manifests/templates/install-config.yaml.j2
```diff
@@ -7,6 +7,27 @@ cpuPartitioningMode: AllNodes
...
+  fencing:
+    credentials:
+    {% for hostname in groups['masters'] | sort %}
+      {% set vendor = hostvars[hostname]['vendor'] | lower %}
+      {% set bmc = hostvars[hostname]['bmc_address'] %}
+      - hostname: {{ hostname }}
+      {% if hostvars[hostname]['fencing_address'] is defined %}
+        address: {{ hostvars[hostname]['fencing_address'] }}
```
tonyskapunk: I don't think this is needed, it will require someone to add a variable named `fencing_address` with information we already have (the bmc address)

It should be a simplified version of: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/96f540fb723bb986f942f7e59a6a5c0241a750d7/roles/installer/templates/install-config.j2#L95-L139

Please include all the supported BMCs, not only dell and hp

roles/installer/templates/install-config.j2
```diff
@@ -44,6 +44,25 @@ controlPlane:
   replicas: {{ nummasters }}
   platform:
     baremetal: {}
+{% if groups['masters'] | length == 2 and (groups['workers'] | default([])) | length == 0 and release_version is version('4.20', '>=') %}
+  fencing:
+    credentials:
+    {% for hostname in groups['masters'] | sort %}
+      - hostname: {{ hostname }}
+      {% if hostvars[hostname]['fencing_address'] is defined %}
+        address: {{ hostvars[hostname]['fencing_address'] }}
```
tonyskapunk: Likewise, I don't think we need yet another var for fencing address, that should be the bmc/ipmi_address.

Please add all the BMCs supported and also add its support in the virtualmedia template.

roles/generate_manifests/defaults/main.yml
ramperher: I had my concerns with the usage of that legacy variable but I'll give a try.

roles/generate_manifests/templates/install-config.yaml.j2
ramperher: In the install-config, there must be 2 control plane nodes and 0 workers, that's why I'm making that check.
I'm finally moving to just support from GA (4.22), because, in previous versions under tech-preview, more fields are needed, and since this is not going to be tested in our test cases, I'd prefer to keep GA support only.

ramperher:
> from change [dci-labs/samsung-ran-lab-config#342](https://github.com/dci-labs/samsung-ran-lab-config/pull/342):
> 
> * FAILURE https://www.distributed-ci.io/jobs/df6b4f3c-2d1c-42dc-a603-7ed1e19168da/jobStates

@tonyskapunk , this is what I was suspecting: `openshift_full_version` variable is undefined because `openshift_version` is undefined in `generate_manifests`. I'll keep testing with the extravar `gm_ocp_version`, but I'd like to know what you recommend here to overcome this situation (a quick solution would be to change `openshfit_version` by `version`, as it's used in the dci-openshift-agent, but I don't want to couple this role to the agent

roles/generate_manifests/defaults/main.yml
ramperher: As said in the PR history, `openshift_version` is undefined, so `ocp_version_major` is undefined. I'm removing all variables depending on `openshift_version` (which are, in fact, unused) and keep using `gm_ocp_version` instead

roles/generate_manifests/templates/install-config.yaml.j2
ramperher: I will remove the fencing_address then.
Regarding the supported BMCs, I just added Dell and HP because they're the only mentioned vendor-specific BMCs in the [docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/installing_a_two_node_openshift_cluster/two-node-with-fencing#sample-install-config-two-node-fencing-abi_install-tnf). In other case, you can use the `redfish+https://` scheme, which is the default case:

_fencing.credentials[].address: The Redfish BMC URL. Must use the redfish+https:// scheme (for example, redfish+[https://192.168.1.10:443/redfish/v1/Systems/1](https://192.168.1.10:443/redfish/v1/Systems/1)). IPMI addresses are not supported. Vendor-specific Redfish schemes such as idrac-redfish+https:// and ilo5-redfish+https:// are also accepted._

So, I'd prefer to keep following this, rather than adding more BMC cases that are not appearing in the docs.

roles/installer/templates/install-config.j2
ramperher: I'll do the same than in generate_manifests; remove fencing_address, and just keep support for Dell and HP as vendor-specific BMCs. I'll apply the same in the virtualmedia template.

Also, bear in mind that it's not under the scope of this change to test TNF on IPI. The requirements from the partner we're testing this feature is just to use ABI. I'll just verify that the new code is not affecting current deployments (already validated).

ramperher:
The change is ready for review. @tonyskapunk , I've addressed all your feedback, including the justification for each decision made. Anyway, feedback to improve the PR is welcome. Thanks!

roles/generate_manifests/templates/install-config.yaml.j2
```diff
@@ -7,6 +7,25 @@ cpuPartitioningMode: AllNodes
 controlPlane:
   name: master
   replicas: {{ groups['masters'] | length }}
+{% if groups['masters'] | length == 2 and (groups['workers'] | default([])) | length == 0 and gm_ocp_version is version('4.22', '>=') %}
```
tonyskapunk: As long as you're sure the requirement is only 2 control-plane nodes and no workers, then that's OK and yes GA is better :+1: 

`gm_ocp_version is version('4.22', '>=')` breaks when `gm_ocp_version` is empty, the default.
I think `gm_ocp_version` should be optional, i.e. not define a default value and evaluate only when defined.
```suggestion
{% if groups['masters'] | length == 2 and (groups['workers'] | default([])) | length == 0 and gm_ocp_version is defined and gm_ocp_version is version('4.22', '>=') %}
```

roles/generate_manifests/defaults/main.yml
tonyskapunk: Yes, is better to get rid of the variables that are not used in the role, not sure if they are expected in other roles, but if that's the case then that can be addressed later, roles should be as independent as possible.

>  openshift_version is undefined, so ocp_version_major is undefined

Technically `ocp_version_major` is always defined, but empty, regardless, they are not used and using a variable with prefix of the role is better, but in this case I don't think we need one at all, not at least in the default, see my comment in the template.

> what you recommend here to overcome this situation

What you did is the way to go, passing the version to the role, but we have to be careful when that variable is not passed and we should be good.

tonyskapunk (commented):
Everything is looking good, we can address the details on the BMCs supported later.
The only concern is the definition of `gm_ocp_version` on the defaults.

roles/generate_manifests/templates/install-config.yaml.j2
ramperher: Ack, will do that change

ramperher:
Made the required modifications to avoid setting a default value for gm_ocp_version + rebased the PR. Ready for review again.

tonyskapunk (approved):
LGTM

PR #1017, opened 2026-08-18
Using yaml instead of text operations to manipulate the image source files

nsilla:
##### SUMMARY

The preflight and operator_sdk roles are affected by an error due to the use of blockinfile operations to extend the contents in the image source files.

The blockinfile module ignores the yaml structure of these files and may corrupt it when appending new content.

This PR, creates a "run on request" tasks file into the mirror_catalog role and replaces the use of the blockinfile module with calls to this tasks file from the prefligh and operator_sdk roles. The append_image_source_mirrors.yml tasks file is able to edit the IS files in a yaml aware mode.

Note the tasks file was added to the mirror_catalog role and not to one of the affected roles, because the mirror_catalog role is the one that initially creates the IS file and thus the tasks are resusable and shared among the affected roles.

Fixes CILAB-2760

##### ISSUE TYPE

- Bug

##### Tests

- [ ] Testvcp: ocp-4.18-vanilla example-cnf - <JobURL>
- [ ] TestvcpWorkload: operator-sdk - <JobURL>

---

Testvcp: ocp-4.18-vanilla example-cnf
TestvcpWorkload: operator-sdk

### DIFF
--- a/roles/operator_sdk/tasks/mirroring.yml
+++ b/roles/operator_sdk/tasks/mirroring.yml
@@ -61,17 +61,14 @@
     oo_index: "{{ mc_catalog_digest }}"
 
 - name: Append operator-sdk images to Image Sources
+  ansible.builtin.include_role:
+    name: redhatci.ocp.mirror_catalog
+    tasks_from: append_image_source_mirrors
   vars:
     scorecard_basic_images: "{{ [scorecard_test_img, scorecard_storage_img, scorecard_untar_img] }}"
-  ansible.builtin.blockinfile:
-    block: |
-      # Scorecard container
-        - mirrors:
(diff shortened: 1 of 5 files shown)


roles/operator_sdk/tasks/mirroring.yml
tonyskapunk, line 1: There is no need to update operator-sdk, it is not used and it has been removed from preflight, we should deprecate its use as well.

tonyskapunk (commented):
I tracked the issue to #1005 IMO the proposed fix is a bit overkill, I'm proposing in #1018  to keep the old behavior instead.

roles/operator_sdk/tasks/mirroring.yml
nsilla: good catch, Tony. Already sent #1018 to the merge queue. I'll close this one.

nsilla:
As indicated, there's a simpler approach available in PR #1018.

PR #1028, opened 2026-08-25
fix: enforce no-changed-when lint rule across all command/shell tasks

fredericlepied:
## Summary

- Removed `no-changed-when` from `.ansible-lint` `warn_list` — the rule is now enforced, not silently skipped
- Added `changed_when: false` to ~100 read-only `command`/`shell` tasks and `changed_when: true` to ~116 state-mutating tasks across 140 role files
- Applied special-case idempotency fixes: `loginctl enable-linger` (3 files) converted to `args.creates`; `restorecon -irv` (2 files) use stdout-length-based `changed_when`

## Changes

- `.ansible-lint`: removed `no-changed-when` from `warn_list` so violations now fail the lint run
- 143 role task files across the collection updated to add explicit `changed_when`:
  - Read-only commands (inspect, cat, grep, wc, list, validate, etc.) → `changed_when: false`
  - State-mutating commands (install, create, configure, etc.) → `changed_when: true`
  - `roles/installer/` (tasks 15, 24) and `roles/sno_installer/` (task 22): `loginctl enable-linger` replaced with `cmd:` form + `args: creates:` for proper idempotency
  - `roles/ocp_on_libvirt/tasks/setup.yml` and `roles/vbmc/tasks/install.yml`: `restorecon -irv` now registers output and uses `changed_when: _restorecon_result.stdout | length > 0`
- Zero `no-changed-when` violations remain after this change

---
Generated with the help of [AI Assist](https://github.com/fredericlepied/ai-assist)


fredericlepied:
Closing: ticket returned to backlog due to inactivity. PRs can be reopened if the ticket is resurrected.

PR #1037, opened 2026-08-27
Additional options to add extra trust bundles to the install-config manifest

nsilla:
##### SUMMARY

In the generate_manifests role, additional trust bundles may be added to the install-config manifest when a local mirror exists in the environment. Some environments, though, contain local registries which are not local mirrors. Therefore we need to extend the additionalTrustBundle options so we can provide the installer with additional certificates without a need to define the mirroring content sources.

##### ISSUE TYPE

- Enhanced Feature

##### Tests

- [x] Testvcp: ocp-4.21-vanilla-abi - https://www.distributed-ci.io/jobs/10ca9b8e-7a83-4d76-93fd-5176cdf585bf/jobStates?sort=date

---

Test-hints: no-check


roles/generate_manifests/tasks/main.yml
```diff
@@ -28,6 +28,33 @@
   when: use_local_mirror_registry  | bool
   no_log: true
 
+- name: Initialize additional trust bundle file contents
+  ansible.builtin.set_fact:
+    _gm_file_certs: []
+
+- name: Load additional trust bundle files
```
betoredhat: If you are using agent, this could be covered by changing some conditions in https://github.com/distributedci/dci-openshift-agent/blob/bed9dee7577108ba6b80174c808900b41fd9ad3b/plays/pre-install.yml#L157-L186

Adding a trust bundle should be possible for all scenarios if the variable is defined.

nsilla: Hi Beto,

Thanks for the suggestion.

If I'm understanding it correctly both code blocks have the same effect and both are initially designed to setup the trust bundle and the content sources for disconnected environments, right?

In your opinion, what are the pros of moving this change to the agent?

betoredhat: Hello Nacho, 

IMO, I think we need to start decoupling the dci_disconnected variable from the configuration options that enable the use of local registries, web servers, additional CAs, ICSP, etc.

None of these infrastructure services necessarily mean that a cluster is disconnected. Using these services is perfectly valid for clusters with or without an internet connection.

So, my proposal is to start making all of this more flexible directly in DOA.

Thanks

nsilla: I see your point now, Beto.

The way I see it, the first step would be aligning the IPI deployment workflow to reuse as much code as possible from the ABI workflow. That way se also expose a single workflow to handle the additional trust bundle and mirror registries.

Right now I'm in a rush to get my lab back running, so I update this PR in a way that I can support my use case with a minimum change to the generate_manifests role. Just by moving the path to the install-config.yml template to a variable I can replace the file with a custom version stored in the lab's config repo.

I'd really appreciate having this change approved so I can unblock the lab status.

vladir-t-k:
image

### "Mexican Wanted" 
**Rapist, child molester, cheater, squatter.**

To all of you who harbor the mexican, be wary of your female colleagues, your wife, your daughter. The Mexican is hiding among you, staring at your family. This is the animal who "offered" his hands to the young Ukrainian refugee girls at the Red Cross, who hooked up with his colleague while attending conferences on business trips, who laid his hands on newborn babies, and who disguised his whereabouts while hiding in a sanctioned country.

You’re living with a rat in your emails, in your workgroup. Every coffee you sip in the morning is mixed with rat shit. The fact that the company is still harboring this animal says a lot about the company itself. Good luck with that, and enjoy your work.

To the mexican: if you have any guts, you’d have torn out your intestines and hanged yourself in the toilet, a destiny that has been delayed for 44 years. Soon, justice and karma will bring you to your fate. An animal like you would and should never have your own family or children.

roles/generate_manifests/README.md
```diff
@@ -55,3 +61,5 @@ All of the variables have sane defaults which you can override to force things,
 [^1]: As per the [OCP 4.17 docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html-single/installing_an_on-premise_cluster_with_the_agent-based_installer/index#root-device-hints_preparing-to-install-with-agent-based-installer)
 
 [^2]: When `gm_image_sources` is defined, it completely replaces the default `imageContentSources` block in the registry configuration. The value should be a YAML string containing the full `ICSP/IDMS` block with proper indentation. If not defined, the default block with standard OpenShift release sources and mirrors is used.
+
+[^3]: Template paths are resolved by Ansible's `template` lookup and module. Use the role-relative filename (for example `agent-config.yaml.j2`) to keep the bundled templates, or pass an absolute path to a custom template on the control node.
```
manurodriguez: Nacho, Could you please include some text here that explains briefly the idea behind the  `gm_*_config_template` variables? My understanding is that they could allow to pass extra settings like the extra trust bundles, but I guess it could be something else right?. Thanks

nsilla: Manny, I just updated the README.md file following your comment. Please check if the new version is satisfactory.

manurodriguez (approved):
LGTM

ramperher (approved):
LGTM

PR #1038, opened 2026-08-27
Use correct status fields and extra conditions for OCS health status check

rajchiluveru:
##### SUMMARY
This PR is to fix incorrect "status" fields used for OCS health status check task

- Wait on CephCluster status.phase and status.state instead of a quoted substring check on status.state.
- Internal ODF reports phase: Ready and state: Created. External ODF reports Connected on both. 
- The old condition looked for Ready/Connected in state, which internal clusters never set.
- Require status.ceph.health to be defined and not HEALTH_ERR.

##### ISSUE TYPE
- Bug

Test-Hints: no-check

### DIFF
--- a/roles/odf_setup/tasks/openshift-storage-operator.yml
+++ b/roles/odf_setup/tasks/openshift-storage-operator.yml
@@ -63,9 +63,13 @@
   until:
     - ocs_health_status.resources is defined
     - ocs_health_status.resources[0].status is defined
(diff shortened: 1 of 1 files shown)


betoredhat:
retest

betoredhat:
lgtm, launching a test

betoredhat (approved):

rajchiluveru:
recheck

PR #1040, opened 2026-08-29
copy_and_render: expanduser source dir to fix ~ path corruption

fredericlepied:
The role derived src_dir with regex_replace('^./', ''), whose unescaped
"." wildcard also matched the "~" in tilde paths. Since the find module
expands "~" in its paths but src_dir kept the tilde stripped, replace()
only matched the suffix after the home dir and produced a corrupted
destination like /home/dci//tmp/.../sites/sno, failing with
"Destination directory does not exist".

Use expanduser so src_dir matches the paths returned by find for all
forms (~/, ./, absolute).

Add a ~-based regression run to the integration test.

Test-Hints: no-check

### DIFF
--- a/roles/copy_and_render/tasks/main.yml
+++ b/roles/copy_and_render/tasks/main.yml
@@ -15,8 +15,10 @@
 
 - name: Render Jinja2 templates to target directory
   vars:
-    # Relative paths starting with "./path/" are handled as "path/"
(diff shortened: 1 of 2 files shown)


fredericlepied:
recheck

fredericlepied:
recheck

fredericlepied:
recheck

pierreblanc (approved):
LGTM

fredericlepied:
recheck

PR #1051, opened 2026-09-09
Resolve must-gather images from CSV annotations; omit unresolved hints

betoredhat:
##### SUMMARY

- Resolve must-gather hints from installed CSVs, preferring operators.openshift.io/must-gather-image, then other must-gather / mustgather annotations, then relatedImages.
- Drop rmgi_fallback_registry
- Role input is hints only; full image refs (containing /) are ignored.
- Return a unique list when several hints map to the same pullspec (mce and multicluster-engine)
- Stop inventing results

Assisted-by: Cursor

Test-hints: no-check

### DIFF
--- a/roles/resolve_must_gather_images/README.md
+++ b/roles/resolve_must_gather_images/README.md
@@ -1,45 +1,53 @@
 # Resolve Must-Gather Images
 
-Resolves must-gather image short names to full image references using
-the `relatedImages` from installed operator ClusterServiceVersion (CSV)
-resources.
+Resolves must-gather hints to full image references using installed
+operator ClusterServiceVersion (CSV) resources.
 
-This allows pipeline configurations to specify short image names
-(e.g. `ptp-must-gather`) instead of hardcoding full image references
-with version tags or digests. The role queries the cluster to find the
-exact image reference matching the installed operator version.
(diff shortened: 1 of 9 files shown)


roles/resolve_must_gather_images/README.md
```diff
@@ -1,107 +1,96 @@
...
-
-```yaml
-rmgi_resolved_images:
-  - "registry.redhat.io/openshift4/ose-must-gather@sha256:abc123..."
-  - "registry.redhat.io/openshift4/ptp-must-gather-rhel9@sha256:def456..."
-```
-
-### Mix short names with full references
+Use a short name that identifies the operator. Prefer the CSV/package
+name when a short token is ambiguous (for example `oadp-operator`
```
tonyskapunk: I'd avoid the term `token` in here, that is an implementation detail. I think it would be nice to simplify the terminology a bit I see:
- Hint, Short name that are almost the same
- Image name that could be used in place of a short name in `rmgi_images`

I'd say use either hint or short name not both, my 2 🪙

tonyskapunk (dismissed):
I have some comments on the doc, but LGTM!

betoredhat:
retest

tonyskapunk (approved):

PR #1055, opened 2026-09-14
Set a default redinessProbe for VMs in kvirt_vm

tonyskapunk:
##### SUMMARY

Set the guestAgentPing readinessProbe by default.
Sets the evictionStrategy when one is provided, otherwise no strategy is
added to the resource.

Gitleaks-Sign: OC4zMC4wfDIwMjYtMDktMThUMjM6MDM6MzN8YjliYWMwMWRiZDk1YTJjNmM3MTZiN2IxMjgwMzY0Y2MxMjI4MWEzOQ==
Gitleaks-Hash: 667d2733b72263bbd5fc2da6fea0a311195a48e0eb53e2b40168bbd6f52dbe67

##### ISSUE TYPE

- Enhanced Feature

##### Tests

- [x] https://www.distributed-ci.io/jobs/48e98832-a643-4fc1-8321-fbf17ff996c0

Test-hints: no-check


roles/kvirt_vm/templates/vm-template.yml.j2
```diff
@@ -62,9 +62,20 @@ spec:
...
+        initialDelaySeconds: 120
+        periodSeconds: 20
+        timeoutSeconds: 10
+        failureThreshold: 3
+        successThreshold: 3
 {% endif %}
 {% if vm.node_selector is defined %}
       nodeSelector: {{ vm.node_selector }}
+{% endif %}
+{% if vm.eviction_strategy is defined %}
```
manurodriguez: Can you please include this var `eviction_strategy` in the role README.md? Or should we default it to `None` as a recommended practice? wdyt?

tonyskapunk: Good idea, I documented the option, I opted for not setting the default to `None`, because the default is not defining one. Could set it to `None` if we want to be more specific for these VMs.

betoredhat (dismissed):
lgtm

roles/kvirt_vm/templates/vm-template.yml.j2
```diff
@@ -62,9 +62,20 @@ spec:
...
 {% if vm.readiness_probe is defined and vm.readiness_probe %}
 {{ vm.readiness_probe | to_nice_yaml(indent=2) | indent(6, first=True) }}
+{% else %}
+      readinessProbe:
+        guestAgentPing: {}
+        initialDelaySeconds: 120
+        periodSeconds: 20
+        timeoutSeconds: 10
+        failureThreshold: 3
+        successThreshold: 3
```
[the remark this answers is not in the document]
tonyskapunk: GuestAgentPing contacts the qemu-guest-agent for availability checks.
    Probe failures are automatically suppressed when the guest agent is
    unreachable for a non-fault reason: during live migration (guest paused
    on one pod while memory is transferred) and whenever the VM is paused
    for an intentional or transient reason such as a user pause, snapshot,
    save, or dump. Failures are not suppressed when the VM is paused due to
    a fault (IO error, crash, or postcopy failure).

manurodriguez (approved):
LGTM

tonyskapunk:
retest

